PR #10, opened 2023-05-07
Add `pipx run` tip to Docs

aaronsteers:
First of all, the tool looks awesome. Nice work!

This PR proposes guidance for users to create an ephemeral environment using `pipx run`, avoiding the need to pre-install and letting `harlequin` be run as a one-liner anywhere that `pipx` exists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,8 @@ Use `pip` or `pipx`:
 pipx install harlequin
 ```
 
-> **Tip**
+> **Tip:**
+>
 > You can run invoke directly with [`pipx run`](https://pypa.github.io/pipx/examples/#pipx-run-examples) anywhere that `pipx` is installed. For example:
 > - `pipx run harlequin --help`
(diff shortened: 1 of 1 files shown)


tconbeer:
Sure! I love pipx. Why pipx run and not just pipx install?

aaronsteers:
> Sure! I love pipx. Why pipx run and not just pipx install?

@tconbeer - Thanks for merging! 

To answer your question above, just from my own experience, I find `pipx run` very convenient personally as almost a stateless invocation pattern for users and developers, which gets through the "do I want to install this?" hesitancy and just let's folks run directly without thinking about whether it's necessary to uninstall later if they aren't actively using it on a regular basis. And if there isn't already a local cache, then the user gets the latest version as of the time they invoke `pipx run`.

Because pipx should be available most places where DuckDB is installed, this lets anyone use (or try out) your TUI with a single command - which I think is really handy for people working across many different environments.

PR #38, opened 2023-05-09
fix: allow starting harlequin via `python -m harlequin`

Mause:
This is especially useful for use with different python versions/installations, such as virtualenvs


tconbeer:
Love it. Thank you!

PR #150, opened 2023-07-27
Fix crash from incorrect selector in error modal.

dljsjr:


tconbeer:
Oof. Thanks so much for this patch!

tconbeer:
Tests are failing since your fork can't use my motherduck token. Merging anyway.

PR #243, opened 2023-09-22
chore: update textual to 0.38.1

tconbeer:


tconbeer:
failure is due to a MotherDuck outage.

PR #251, opened 2023-10-02
feat: use fastdatatable for 1000x speedup

### LINKED ISSUE #181: Implement a new data table widget
The built-in widget is ok, but its data structures are super inefficient, which limits us to about 10k rows. It shouldn't be too bad(?) to fork the current widget and represent each row as a tuple of its data.

tconbeer:
closes #181


tconbeer:
failure is motherduck-related. annoyingly can't use the same duckdb version on all platforms at the moment.

PR #367, opened 2023-12-21
Add BigQuery adapter

joshtemple:


tconbeer:
@joshtemple I can't push to this  branch, but the checks are failing because you need to run `poetry lock` to generate a new `poetry.lock` file, and then commit that file alongside the updated `pyproject.toml` file

tconbeer:
going to close this, since your commit is included in #368

PR #386, opened 2024-01-07
list items in the data_catalog pane ordered by label

Fuzzwah:
A quick and easy change to order the items in each subtree alphabetically based on their label.

I realise that this isn't the optimal or well fleshed solution, and perhaps something more configurable would be better. However, just as a default starting point I think it's a decent change.


tconbeer:
Thanks for your PR. Maybe should be a discussion instead, though.

I think sorting catalog items should probably be the responsibility of the adapter; some items (columns in particular) have lexical ordering that the adapter is more likely to understand, and that Harlequin core should probably respect.

tconbeer:
Postgres catalog objects are now sorted; the patch was released as harlequin-postgres v0.2.2

PR #448, opened 2024-02-05
Add number separators to query history results

### LINKED ISSUE #437: Add number separators to query history results
image

just a tweak to the f-string here:
https://github.com/tconbeer/harlequin/blob/a78944fafca3c52498b053c473f83f6a1a663d1b/src/harlequin/history.py#L26

code-master-ajay:
This fixes #437 

The previous code was :

```
f"{self.result_row_count:n} " 
```
which is replaced by :

```
f"{self.result_row_count:,} " 
```

this will make sure there are separators in the numbers, example given below: 

harlequin

### DIFF
--- a/src/harlequin/history.py
+++ b/src/harlequin/history.py
@@ -23,7 +23,7 @@ class QueryExecution:
             result = Text("ERROR", style="bold italic red", justify="right")
         else:
             res = (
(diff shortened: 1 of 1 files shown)


tconbeer:
Thank you!!

PR #459, opened 2024-02-09
refactor: don't use adapter for export; feat: add orc, feather

### LINKED ISSUE #413: Refactor export; make adapter-agnostic
**Describe the bug**
When connected to a database (in my case, SQL Server) with the odbc driver, you cannot export data with ctrl+e, even though the data is clearly exposed in the interface.

**To Reproduce**
1. connect to harlequin with odbc connector
2. run any query
3. press "ctrl+e" to export the data
4. you will get an error

**Expected behavior**
data exports as csv.

**Actual behavior**
you get this error.
image

**Additional context**

What database adapter are you using with Harlequin? ODBC

Can you tell us more about your system?
- Shell: Powershell
- Terminal: Windows Terminal
- OS of the shell: Windows 11 (native)
- OS of the terminal (if different from the shell):

tconbeer:
closes #413


tconbeer:
ORC isn't supported on windows, apparently: https://issues.apache.org/jira/browse/ARROW-7811

need to skip the test but should also remove the ORC option on windows, so it's not visible.

Issue #383, opened 2024-01-03
FR:  add Snowflake adapter


kylelundstedt:
Hi @tconbeer!  Just started playing with Harlquin; seems very cool.
In addition to DuckDB, we also use Snowflake.  Any chance we'll see an adapter for it?

Thanks!

rupurt:
It would be awesome to implement this as an ADBC adapter. Snowflake has support and there Arrow Snowflake driver https://arrow.apache.org/adbc/0.5.1/driver/snowflake.html

tconbeer:
Agreed that Arrow is the way to go. Harlequin's data table is backed by arrow, so results returned as arrow tables will load faster. Snowflake connector 2.2+ also lets you fetch results as arrow with fetch_arrow_all.

rymurr:
Ive started working on one here: https://github.com/rymurr/harlequin-snowflake

Basics are workign but the catalog is very slow.

TylerHillery:
@rupurt I have implemented an ADBC adapter [here](https://github.com/TylerHillery/harlequin-adbc) and I got it to work with my Snowflake account but it appears others are having issues. I can't seem to reproduce on my end so would love some more people to test it out.

I've tired adding to Harlequin pyproject.toml as an extra but the ADBC adapter requires python 3.9 or greater so it wouldn't let me. I also just haven't gotten around to contributing to the docs yet.

tconbeer:
> I've tired adding to Harlequin pyproject.toml as an extra but the ADBC adapter requires python 3.9 or greater so it wouldn't let me. I also just haven't gotten around to contributing to the docs yet.

@TylerHillery you can add a python version constraint to the extra dependency, like this:
```toml
harlequin-adbc = { version = "^0.1", python = ">=3.9.0", optional = true }
```
*However*, I tried that just now, and there is a dependency issue with PyArrow versions; `harlequin-databricks` basically requires pyarrow 14 (via `databricks-sql-connector`), and you require pyarrow 15:

```
    Because harlequin-databricks (0.1.1) depends on databricks-sql-connector (>=3.0.3,<4.0.0)
 and no versions of harlequin-databricks match >0.1,<0.1.1 || >0.1.1,<0.2, harlequin-databricks (>0.1,<0.2) requires databricks-sql-connector (>=3.0.3,<4.0.0).
[...]

PR #516, opened 2024-04-13
Add ADBC Adapter

TylerHillery:
No open issue by does relate to #383 (I didn't want to mark it as "resolves" as it's not technically a Snowflake specific adapter so I will leave that up to you to decide)

Adds adbc adapter as an extra

Followed typical PR format for adding an adapter as an extra

- [ ] No.
- [x] Yes; I haven't opened a PR, but the gist of the change is: Need to add ADBC adapter docs

- [ ] Yes.
- [x] No, I believe tests aren't necessary.


TylerHillery:
I am an idiot 🤦‍♂️  I added `harlequin-adbc = { version = "^0.1", python = ">=3.9.0", optional = true }` then ran `poetry lock` forgetting to add `adbc = ["harlequin-adbc"]` so I went back to add it forgetting to run `poetry lock` again afterwards

PR #518, opened 2024-04-15
feat: close the adapter's connection on quit

mpetazzoni:
This change introduces a new optional method of HarlequinConnection implementations in adapters, giving them an opportunity to cleanly close the connection to the underlying database when the app exists.

The `close()` method is called in the `action_quit()` handler.

- [ ] No.
- [x] Yes, and I have opened a PR at https://github.com/tconbeer/harlequin-web/pull/67
- [ ] Yes; I haven't opened a PR, but the gist of the change is that the new `HarlequinConnection.close()` can be documented in the "Creating an Adapter" guide.

- [ ] Yes.
- [x] No, I believe tests aren't necessary – this is an optional functionality of adapters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Features
+
+-   A new `HarlequinConnection.close()` method can be implemented by adapters to gracefully close database connections when the application exits.
+
 ### Bug Fixes
 
 -   Wrong link on text_area_clipboard_error error message ([#509](https://github.com/tconbeer/harlequin/issues/509))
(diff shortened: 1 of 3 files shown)


tconbeer:
Thanks. Can I ask what prompted this? I thought most python DB libraries close the connection when the handle leaves scope/is deleted.

mpetazzoni:
@tconbeer Thanks for taking a look! Python does not guarantee that the `__del__` finalizer gets called on interpreter exit, and since the Harlequin app keeps a reference to it (`self.connection`), nor does it use a context manager for the connection object around the lifecycle of the Harlequin app, there is no _current_ possible mechanism for an adapter to perform a graceful shutdown/close of its connection to the database.

Of course, connections are severed when Python exists, but there might situations where adapters want to perform on-shutdown actions (flush, commit, emit telemetry, or gracefully release resources). This new `close()` method gives them an opportunity to do so, if they want to. See for example the [DB-API spec](https://legacy.python.org/dev/peps/pep-0249/#connection-objects).

tconbeer:
ok, I'm on-board. thanks for the extra info, makes sense.

PR #565, opened 2024-06-08
feat(extra): Add Cassandra Adapter

vkhitrin:
Adding an experimental `harlequin-cassandra` adapter.

Closes #XYZ (an existing open issue)

**What** are the key elements of this solution?

Adding an experimental Cassandra driver.

**Why** did you design your solution this way? Did you assess any alternatives? Are there tradeoffs?

Why? `harlequin` is fun :)

Does this PR require a change to Harlequin's docs?
- [ ] No.
- [x] Yes, and I have opened a PR at [tconbeer/harlequin-web](https://github.com/tconbeer/harlequin-web/pull/88).
- [ ] Yes; I haven't opened a PR, but the gist of the change is: ...

Did you add or update tests for this change?
- [ ] Yes.
- [x] No, I believe tests aren't necessary.
- [ ] No, I need help with testing this change.

Please complete the following checklist:
- [x] I have added an entry to `CHANGELOG.md`, under the `[Unreleased]` section heading. That entry references the issue closed by this PR.
- [x] I acknowledge Harlequin's MIT license. I do not own my contribution.

PR #566, opened 2024-06-11
fix/#565/cassandra lockfile

tconbeer:
Supersedes #565 

Updates the poetry lockfile with the new extra


tconbeer:
@vkhitrin FYI, this will auto-merge when the checks complete; GHA are currently down but hopefully that resolves soon. Thanks again for contributing the Cassandra adapter!!

PR #807, opened 2025-04-25
feature(tui): Add Adapter Details Screen

vkhitrin:
Adds a new window that displays details returned by the adapter.

- [x] No.

- [ ] Yes.
- [x] No, I need help with testing this change.

- [ ] I have added an entry to `CHANGELOG.md`, under the `[Unreleased]` section heading. That entry references the issue closed by this PR.

### DIFF
--- a/src/harlequin/adapter.py
+++ b/src/harlequin/adapter.py
@@ -213,6 +213,7 @@ class HarlequinAdapter(ABC):
     COPY_FORMATS: list[HarlequinCopyFormat] | None = None
     """DEPRECATED. Adapter Copy formats are now ignored by Harlequin."""
     IMPLEMENTS_CANCEL = False
+    ADAPTER_DETAILS: str | None = None
 
     @abstractmethod
     def __init__(self, conn_str: Sequence[str], **options: Any) -> None:
@@ -267,3 +268,10 @@ class HarlequinAdapter(ABC):
         also provide options for customizing the Export dialog GUI.
         """
         return self.COPY_FORMATS is not None
+
+    @property
+    def provides_details(self) -> None:
+        """
+        Provides adapter details.
+        """
+        return self.ADAPTER_DETAILS is not None
--- a/src/harlequin/components/adapter_details.py
+++ b/src/harlequin/components/adapter_details.py
@@ -1,0 +1,65 @@
+from pathlib import Path
+
+from textual import events
+from textual.app import ComposeResult
+from textual.containers import Vertical, VerticalScroll
+from textual.screen import ModalScreen
+from textual.widgets import Markdown, Static
+
+
+class VerticalSuppressClicks(Vertical):
+    def on_click(self, message: events.Click) -> None:
+        message.stop()
+
+
+class AdapterDetailsScreen(ModalScreen):
+    def __init__(
+        self,
+        details: str,
+        name: str | None = None,
+        id: str | None = None,  # noqa: A002
+        classes: str | None = None,
+    ) -> None:
+        super().__init__(name, id, classes)
+        self.details = details
+
+    header_text = """
+        Harlequin can display details about the adapter collected from client + server.
+    """.split()
+
+    def compose(self) -> ComposeResult:
+        markdown = f"""{self.details}"""
+
+        with VerticalSuppressClicks(id="adapter_outer"):
+            yield Static(" ".join(self.header_text), id="adapter_header")
+            with VerticalScroll(id="adapter_inner"):
+                yield Markdown(markdown=markdown)
+            yield Static(
+                "Scroll with arrows. Press any other key to continue.",
+                id="adapter_footer",
+            )
+
+    def on_mount(self) -> None:
+        container = self.query_one("#adapter_outer")
+        container.border_title = "Harlequin Adapter Details"
+        self.body = self.query_one("#adapter_inner")
+
+    def on_key(self, event: events.Key) -> None:
+        event.stop()
+        if event.key == "up":
+            self.body.scroll_up()
+        elif event.key == "down":
+            self.body.scroll_down()
+        elif event.key == "left":
+            self.body.scroll_left()
+        elif event.key == "right":
+            self.body.scroll_right()
+        elif event.key == "pageup":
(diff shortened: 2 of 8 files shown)


vkhitrin:
Hey @tconbeer, I initially proposed this a year ago in a discussion https://github.com/tconbeer/harlequin/discussions/564
(I have added an example there, I verified that this also works in Harlequin 2.x, by adding placeholder text to SQLite adapter).

Please let me know if you would like this kind of feature in Harlequin. If you are interested, I will refine this PR with proper code, examples, and tests (if required).

src/harlequin/components/adapter_details.py
```diff
@@ -0,0 +1,65 @@
+from pathlib import Path
```
tconbeer, line 1: Needed for 3.9 compatibility:

```suggestion
from __future__ import annotations

from pathlib import Path
```

src/harlequin/components/adapter_details.py
```diff
@@ -0,0 +1,65 @@
...
+from textual.screen import ModalScreen
+from textual.widgets import Markdown, Static
+
+
+class VerticalSuppressClicks(Vertical):
+    def on_click(self, message: events.Click) -> None:
+        message.stop()
+
+
+class AdapterDetailsScreen(ModalScreen):
```
tconbeer, line 15: Let's name this screen something like `DebugDetailsScreen` or `InformationScreen`

tconbeer: Actually, since you're re-using the HelpScreen, best long-term to refactor the common bits out into a Screen that can be used by both Help and this screen. That way you don't have to repeat the CSS, key handling etc.

src/harlequin/adapter.py
```diff
@@ -213,6 +213,7 @@ class variable.
     COPY_FORMATS: list[HarlequinCopyFormat] | None = None
     """DEPRECATED. Adapter Copy formats are now ignored by Harlequin."""
     IMPLEMENTS_CANCEL = False
+    ADAPTER_DETAILS: str | None = None
```
tconbeer, line 216: let's make it clear this is markdown

src/harlequin_sqlite/adapter.py
```diff
@@ -279,6 +280,10 @@ def __init__(
             self.can_load_extensions = hasattr(
                 sqlite3.Connection, "enable_load_extension"
             )
+            self.ADAPTER_DETAILS = (
```
tconbeer, line 283: This makes me think a class var isn't the right way to do this -- I think Harlequin should just present the configured values for all of the `ADAPTER_OPTIONS`; that way adapter authors don't have to implement anything.

tconbeer (changes requested):
This is a great start. Thanks so much for this PR!

I think you're actually onto something quite a bit bigger. Instead of just showing the configured Adapter options, let's create a "Debugging Information" screen that shows ALL Harlequin config, including Adapter options. So that would include things like the active profile, the config file location(s), the active keymaps, the theme, etc., as well as all of the adapter options. Essentially anything that gets fed into the `Harlequin()` constructor in `cli.py`, plus either that module's `config` dict, or the underlying properties of the adapter object (plus some other CLI options that don't make it into the constructor, like the locale settings).

To be clear, you don't need to implement this whole thing -- I think it can grow over time -- but I'd like this to fit in a framework that can scale up to include more info.

vkhitrin:
> This is a great start. Thanks so much for this PR!
> 
> I think you're actually onto something quite a bit bigger. Instead of just showing the configured Adapter options, let's create a "Debugging Information" screen that shows ALL Harlequin config, including Adapter options. So that would include things like the active profile, the config file location(s), the active keymaps, the theme, etc., as well as all of the adapter options. Essentially anything that gets fed into the `Harlequin()` constructor in `cli.py`, plus either that module's `config` dict, or the underlying properties of the adapter object (plus some other CLI options that don't make it into the constructor, like the locale settings).

This is a fantastic idea :fire:!

I would be happy to re purpose this PR for this.
Based on your idea, I am thinking of drafting 3 "categories" (even if most of them will be unimplemented in the scope of this PR):
* Harlequin Settings - Config, keymaps, theme, and so on.
* Adapter (client side) - Information which is fetched from the client (for example, masked connection string).
* Adapter (server side) - Information which is fetched after establishing connection with the server (for example, in Cassandra, the negotiated protocol version, and database version).



> 
> To be clear, you don't need to implement this whole thing -- I think it can grow over time -- but I'd like this to fit in a framework that can scale up to include more info.

Ack, I will start refactoring this draft.

Thank you for the initial comments on the current draft, I will take them into account for my next draft.

vkhitrin:
@tconbeer, sorry for the long delay.

I have resumed working on this PR, and implemented a "generic" debug information screen.
Before proceeding with tests, documentation, CSS and other requirements, I would like to discuss this proposal.

The screen is implemented via `DebugInfoScreen` which is accessed using `F12` by default:
image
It provides two sections:
1. `Harlequin Details`
    * This is implemented using `HarlequinDebugInfo` and provides details about Harlequin configuration used (list is subject to change):
       image
    * Contains a collapsible element `Full Config` which renders the config file:
       image
2. `Adapter Details`
    * This is implemented using `AdapterDebugInfo` and provides info about the adapter:
       image
    * "client-side" (adapter) information is populated using `ADAPTER_DETAILS` variable in each adapter. Information is stored in the collapsible element `Details`:
        image
    * Contains a collapsible element `Adapter Options` which displays the arguments passed to the adapter:
       image

tconbeer:
3.12 and 3.13 will have different snapshots, due to changes in the sqlite library; I don't think you want to update any test snapshots except the ones you added, so you can revert those changes, and generate any new snapshots using Python 3.10 or 3.11 (the snapshot tests where the 3.12 and 3.13 differences are visible should be skipped for those higher python versions)

tconbeer:
Ok, yes, that would be true if the key binding for the new screen appears in the footer. I'm slammed this week but might have time to review and get the tests passing next week

tconbeer:
I can't push to this branch, so I'm finalizing things over at #853

PR #977, opened 2026-07-14
Vim support in the code editor and vim keybinds in results

Dking08:
Closes discussions like #371 and other issues users closed becuase they shifted to other tools

Harlequin currently lacks a Vim-style code editor, so I implemented one while keeping changes to the existing codebase to a minimum.
I modified `components/code_editor` to support **both** the existing code editor and a new Vim-based code editor. I also added a small `vim_status_bar` component that displays the current editing mode (e.g., NORMAL, INSERT). This status bar is only visible when Vim support is enabled.
Other changes in cli.py, app.py and config.py are just to support the config-file based enable of this vim feature.

---

Why? - because I have seen multiple users including me getting frustrated as Harlequin lacks vim support, so I built one.
The implementation relies on an additional dependency, `textual-vim-textarea`, a package I developed specifically for this feature. The package is stable and self-contained, requiring no further changes for this integration.
I tried other other packages but nothing worked, because Harlequin uses textual-textarea, which is a custom widget with different requirements.
There are effectively no tradeoffs. When Vim support is disabled, Harlequin behaves exactly as it did before. When enabled users can enjoy the vim keybinds!

---

- [x] No.

- [ ] Yes.
- [x] No, I need help with testing this change.

- [x] I have added an entry to `CHANGELOG.md`, under the `[Unreleased]` section heading. That entry references the issue closed by this PR. - it don't close any issue.


Dking08:
The Vim enabled interface

Dking08:
User can enable the vim support using the `--config` command too!

image

Dking08:
@tconbeer should I also raise a PR to change docs? I am not sure

Dking08:
Any changes required @tconbeer ?

clang-engineer:
I strongly agree that Harlequin needs a proper way to use Vim in the Query Editor. I already use a custom Vim-style keymap for the Data Catalog and Results Viewer, so this would complete the workflow nicely.

One thought about the configuration: would `code_editor = "vim"` be preferable to `vim_code_editor = true`?

Even without introducing a full editor plug-in system, representing this as an editor choice would avoid an implementation-specific boolean and scale more naturally if Harlequin supports another editor implementation in the future.

Thanks for working on this!

Dking08:
Thanks for the support @clang-engineer
Your `vim_code_editor` flag suggestions would be a better approach for this and can scale too in future. 
I am updating the code to use this flag instead of hardcoding 'vim' in code_editor

clang-engineer:
> I am updating the code to use this flag instead of hardcoding 'vim' in code_editor

Thanks! Just to clarify, I was suggesting the opposite: code_editor = "vim" instead of a vim_code_editor = true flag.

My thought was that code_editor models the editor choice directly and would scale more naturally if Harlequin supports additional editor implementations in the future.

Dking08:
Thanks for clarifying, @clang-engineer! I had understood your suggestion, but my poor phrasing made it sound confusing. I’ve started working on it and will push the changes soon.

Dking08:
I have updated the code,
1. Merge conflicts are resolved
2. Made changes to config

clang-engineer:
Would it make sense to constrain code_editor in the config schema as well, e.g. with a string type and an enum of ["default", "vim"]?

The CLI already models this as an explicit editor choice via click.Choice, so keeping the schema equally strict would make the configuration contract consistent and avoid unknown values silently falling back to the default editor.

Dking08:
Thank you @clang-engineer for another valuable suggestion - adding a stricter schema to `code_editor`

Dking08:
Updated schema

clang-engineer:
Sounds good, thanks for incorporating it!

waldnzwrld:
Came here specifically looking for this, since trying to set keybindings is recording left_shift when I try to record ? 

Can't wait for this to merge

PR #984, opened 2026-07-25
feat: open a table's data on double-click / submit in the Data Catalog

scantinibudden:
Double-clicking (or pressing <kbd>enter</kbd> on) a relation node in the Data Catalog currently just inserts the relation name into the editor. This changes it to open the table's data directly: it puts `select * from <relation> limit 100` in a new buffer and runs it. Non-relation nodes (columns, schemas) keep the existing insert-name behavior.

Relations are recognized by checking the catalog item's class name against `RelationCatalogItem` (by name, to avoid importing any adapter), so it works for every adapter's tables/views/materialized views.

### Note on the behavior change
This changes the default meaning of a double-click, so it's the most opinionated of the set — I'm very happy to gate it behind a config option (e.g. `catalog_double_click = "insert" | "open"`) if you'd prefer to keep the current default. Flagging that explicitly for discussion.

### Tests
Adds functional tests covering both paths (relation submit runs `select *`; non-relation still inserts the name).

https://claude.ai/code/session_01FZYuqGN3VBZYL84jz4rWYz

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Features
+
+- Double-clicking (or pressing <kbd>enter</kbd> on) a table or view in the Data Catalog now opens its data directly, running `select * from <relation> limit 100` in a new buffer, instead of only inserting its name into the editor.
+
 ### Performance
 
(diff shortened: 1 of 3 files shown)


tconbeer:
There is an interaction for this ("Preview Data"), which I think is sufficient. I don't like having the default double-click behavior different for tables vs. other catalog objects.

PR #985, opened 2026-07-25
feat: clickable ✕ close button on each Query Editor tab

scantinibudden:
Adds a clickable `✕` close button to each Query Editor tab (using Textual `@click` markup), in addition to the existing <kbd>ctrl+w</kbd> binding — a small discoverability win for closing buffers with the mouse.

The close is deferred on a short timer so the click's own tab-activation message is processed first. This also fixes a crash when closing an **inactive** tab: without the defer, the activation lands on the just-removed tab and Textual raises `No Tab with id ...`. The `✕`'s link underline is removed via `app.tcss`.

### Tests
Adds a regression test that reproduces the inactive-tab close crash (via the Textual pilot) and asserts the tab closes cleanly; refreshes the affected editor snapshots (the `✕` now appears in tab labels).

https://claude.ai/code/session_01FZYuqGN3VBZYL84jz4rWYz


tconbeer:
couple of issues:

The  styling isn't great:
- I'd prefer a smaller X, maybe https://www.compart.com/en/unicode/U+02E3, probably with one more space in between the tab name and the x, to separate the click targets a little.
- This seems like a textual/measurement bug, but the wide X seems to overflow off the end of the tab (this might get fixed by changing the symbol), and the hover state is very broken
- I'd prefer a grey/dim x that turns green (secondary) on hover (of the x only)

Also:
- the logic for changing focus is wrong... clicking the X seems to focus the tab, so closing a tab with clicking always resets focus to the first tab. For example, with tabs 1, 2, 3 and focus on 2, if I click X on 3, it refocuses to 1.

scantinibudden:
Thanks — reworked on top of the single-`TextArea` refactor (#1072), so the branch is rebuilt rather than rebased.

- Glyph is now U+02E3 `ˣ` with two spaces before it. It's the single `BufferTab.CLOSE_BUTTON` constant if you'd rather have `×`.
- Styling: dim (`$foreground 50%`) by default, `$secondary` on hover of the ˣ only — no hover background, no bold, so nothing overflows the tab.
- Focus: the click on the ˣ is handled by the tab itself and no longer activates it. Closing an inactive tab leaves the active buffer untouched; closing the active tab moves to its neighbour, same as ctrl+w. Covered by `test_tab_close_button`.

Snapshots regenerated on 3.10; full `pytest -m "not online"` green locally apart from `test_load_extension`, whose committed `hello0.dylib` is x86_64-only.

PR #992, opened 2026-08-03
perf: debounce the autocomplete merge during lazy catalog loads

scantinibudden:
No linked issue — this came from using Harlequin daily against a high-latency (remote postgres) database; happy to open one if you'd prefer.

Every lazily-fetched catalog node calls `extend_catalog` on both completers, and each call ended with `_merge_completions`, which re-sorts the **entire** completion corpus (keywords + functions + catalog) — two full O(n log n) sorts on the main thread per fetched node. Against a remote database, the Data Catalog's background loader streams `NewCatalogItems` for many seconds, and the repeated merges make the whole UI feel sticky (even pure client-side actions like collapsing a tree node queue behind them).

This PR:
- adds a `defer_merge` flag to `extend_catalog` (default `False`, so the public behavior is unchanged for other callers) and extracts the merge into a public `merge()` method;
- the app now extends with `defer_merge=True` and schedules a single merge at most once per second while items are streaming in.

Fetching behavior is unchanged — the catalog stays exactly as fresh as before; new completions just become visible up to 1s later, instead of costing a full re-sort each.

- A trailing (resetting) debounce would do fewer merges but could starve completions for the entire duration of a long lazy-load stream; the non-resetting 1s timer bounds staleness at ~1s.
- An incremental merge (insort per batch) would avoid the full sort entirely, but it's a bigger change to the completer internals for little additional gain once the merge runs ≤1/sec.
- Tradeoff: completions from a just-loaded node may take up to 1s to appear in the menu.

- [x] No.

- [x] Yes.

https://claude.ai/code/session_01EecCPQjLNeHoYHCRQ8QJkZ


tconbeer (approved):
thanks for this PR!

Issue #524, opened 2024-04-16
Add `-c`/`-cmd`/`--command` option to run sql command from the command-line


tconbeer:
Inspired by #472

Should just execute the query in the normal way.

PR #1017, opened 2026-08-06
perf: keep Textual out of the adapter API's import graph (#524)

tconbeer:
Toward #524. This is **PR 1 of the M1 `hsql` plan** — "Import hygiene, and the guard that keeps it" (`docs/plans/m1-hsql-technical-plan.md` §4). The value is that the adapter API becomes importable without the TUI, which is the precondition for everything else in Release A.

Each bullet is one of the five causes in §1.2 of the plan:

1. **`harlequin/__init__.py` resolves its public names through a PEP 562 `__getattr__`.** Importing *any* `harlequin.*` submodule executes this `__init__`, which imported `harlequin.app` at module scope — a ~660ms floor under every consumer, including adapters that never touch the TUI. All eleven exported names still resolve, from the same modules, and are cached in module globals on first access.
2. **`AutoBackendType` moves under `TYPE_CHECKING`** in `adapter.py`, `harlequin_duckdb`, and `harlequin_sqlite`. It is an alias for `Any`, and importing it cost every adapter `textual_fastdatatable.backend`.
3. **`NewCatalog`/`NewCatalogItems` move to `harlequin/messages.py`.** They are Textual messages that lived in `catalog.py`, which every adapter imports. Nothing imports them but the app, so they move with no shim and no deprecation.
4. **`options.py` defers its renderer imports** into `to_widgets()` and `to_questionary()`. Declaring an option no longer costs `questionary` (130ms) or Textual (150ms, plus 264ms for the themes `harlequin.colors` pulls in). `_CustomValidator` becomes `copy_widgets.CustomValidator`, because subclassing `Validator` binds Textual at class-definition time and can't be deferred in place.
5. **The two stdout leaks from §1.5 are fixed** — a plug-in that fails to import, and `pretty_print_warning` — both now write to stderr, as `pretty_print_error` already did.

Cause 3 in §1.2 is upstream, so this also **bumps `textual-fastdatatable` to 0.16.1**, the release that makes the `DataTable` widget a lazy import and defers `pyarrow.parquet`. Without it, moving `AutoBackendType` under `TYPE_CHECKING` would still leave the backend unreachable without Textual.

Plus the guards the plan asks for: two import-linter contracts, `tests/unit_tests/test_import_hygiene.py`, and `scripts/cold_start.py` (with a `make cold-start` target and a CI step that writes the table to the job summary).

Measured best-of-five against `origin/main` at the same commit, on a quiet machine:

| import | before | after | Textual? |
| --- | --- | --- | --- |
| `harlequin_duckdb` | 768ms | **121ms** | no |
| `harlequin_sqlite` | 699ms | **65ms** | no |
| `harlequin` | — | **21ms** | no |
| `from harlequin import Harlequin` (the TUI) | 685ms | 685ms | yes, as it should |

`import duckdb` alone is 85ms of that 121ms, so Harlequin's own share is ~36ms. Re-measured against the released 0.16.1 wheel as a paired before/after run, the ratios hold: 6.1x for duckdb and 11.9x for sqlite, with identical module counts.

**The two guards check different things, and both are needed.** import-linter reads the static graph, so it cannot distinguish a module-scope import from one deferred into the function that needs it — every deferral introduced here shows up as a contract violation. Rather than weaken the contract, the four deliberate deferrals are listed in `ignore_imports`, one commented line each, so a reviewer sees exactly what is exempt and any *new* route to Textual still fails. The subprocess tests are what prove the deferrals actually defer: they import each module for real and read `sys.modules`. A new deferral needs an entry in both places, which is the point.

Timing is deliberately not asserted. §5 of the plan is explicit that a wall-clock assertion on a shared runner is a flaky test that gets deleted within a month, so `cold_start.py` is informational and reports into the job summary; the blocking guards are the deterministic ones.

**Deferred imports have a real readability cost**, and they are confined to the three functions where the money is: two renderer methods and one warning printer. `click` stays at module scope in `options.py` — it is 21ms net, and both commands need it.

Note for review
- **The first commit is `ruff format` on the M1 plan doc, and is unrelated to this work.** ruff 0.16 formats Python code blocks inside Markdown and the lockfile pins 0.16.1, so `ruff format . --diff` — which the static workflow runs — already fails on `main` as merged. I verified that against a clean `origin/main` worktree. It's a whitespace-only change to code blocks, no prose touched; it's in its own commit so it can be dropped or landed separately.
- `AGENTS.md` gains an "Import hygiene" section stating the invariant and both guards, since this is exactly the kind of rule a future contributor would undo by accident.

- [x] No.

- [x] Yes.

- `tests/unit_tests/test_import_hygiene.py` — the thirteen headless import paths, and the lazy package root: all eleven public names resolving to the same objects as a direct import, and `AttributeError` still raised for anything else.
- `tests/unit_tests/test_plugins.py`, `tests/unit_tests/test_exception.py` — the two stdout leaks, each with the module it covers. The subprocess runner they share is a fixture in the new `tests/unit_tests/conftest.py`.

Full suite: 205 passed, 6 skipped, 140 snapshots passed on 3.10; 3 passed on the `py12` selection. `ruff`, `mypy --no-incremental` and `lint-imports` all clean.

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -36,3 +36,17 @@ jobs:
           uv run --no-sync ruff format . --diff
           uv run --no-sync ruff check .
           uv run --no-sync mypy --no-incremental
+          uv run --no-sync lint-imports
+
+      # Informational: the deterministic guards are the import-linter contracts
+      # above and tests/unit_tests/test_import_hygiene.py. A wall-clock
+      # assertion on a shared runner would just be a flaky test.
+      - name: Report cold start
+        run: |
+          {
+            echo '### Cold start'
+            echo
+            echo '```'
+            uv run --no-sync python scripts/cold_start.py
+            echo '```'
+          } >> "$GITHUB_STEP_SUMMARY"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -49,6 +49,25 @@ uv run --python 3.12 --group test pytest -m 'py12 and not online' --snapshot-upd
 - Async tests need `@pytest.mark.asyncio`. Await `wait_for_workers(app)` (fixture) rather than sleeping — it skips the catalog background loader, which never finishes on its own.
 - Shared fixtures: `tests/conftest.py` builds throwaway DuckDB/SQLite databases (`tiny_*`, `small_*`) and app instances (`app`, `app_all_adapters`, `app_small_duck`, …); `app_all_adapters` is parametrized so one test body covers both bundled adapters.
 
+## Import hygiene
+
+The headless CLI (`hsql`, planned in `docs/plans/`) has to reach a database without paying for the TUI, so **the adapter-facing modules must stay free of Textual, questionary, prompt_toolkit and sqlfmt at import time.** Concretely:
+
+- `harlequin/__init__.py` resolves its public names through a PEP 562 `__getattr__`. Never add a module-scope import there — importing *any* `harlequin.*` submodule executes it, so one eager import puts the whole app behind every consumer.
+- Type-only imports of Textual (e.g. `AutoBackendType`) go under `if TYPE_CHECKING:`; every module involved already has `from __future__ import annotations`.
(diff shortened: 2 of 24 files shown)


tconbeer:
Can't merge until the upstream changes are released (blocked by GitHub outage) and the pin on textual-fastdatatable is moved to a version.

Issue #1019, opened 2026-08-07
Query splitting breaks inside a dollar-quoted body (for some statements)


tconbeer:
A semicolon inside a `$$ … $$` body is treated as a statement separator:

```sql
create function f() as $$ select 1; $$; select 2
```

splits into three statements:

```
"create function f() as $$ select 1;"
"$$;"
"select 2"
```

rather than the two there actually are.

**Cause.** `tree-sitter-sql` does not model dollar quoting, so the semicolon inside the body is captured by the `(";" @semicolon)` query like any other. It is the grammar's behavior, not something the call site can work around — which is why it affects the Query Editor and any other consumer of the splitter identically. Unrelated to #1015, which was a units bug in how the editor read the grammar's output; this one is the grammar itself.

**Status.** Recorded rather than fixed: `tests/unit_tests/test_statements.py::test_dollar_quoted_body_is_not_split` is an `xfail(strict=True)` for exactly this script, so the day the grammar learns dollar quoting, the test fails and tells us.

Fixing it means either an upstream grammar change or a pre-pass over dollar-quoted regions before the tree-sitter query runs.

tconbeer:
The issue is much narrower than the title suggests - the grammar only gets it wrong in some cases, like `create function f() as $$...;$$`. A regular `select $$foo$$, $label$bar$label$` works fine.

tconbeer:
Reading through some issues in tree-sitter-sql, I think this might be more of an issue with the create-function parser than with $$

Issue #1018, opened 2026-08-07
Query splitting cuts in the wrong place when a line has non-ASCII text before a semicolon


tconbeer:
Running a selection in the Query Editor splits at the wrong character when anything non-ASCII appears earlier on the same line as a semicolon.

```sql
select '日本語';select 2
```

Ctrl-A, Ctrl-J submits two statements:

```
"select '日本語';select"
"2"
```

Both are syntax errors. The correct split is `select '日本語';` and `select 2`.

**Cause.** `CodeEditor._query_separators()` collects `node.end_point.column` from tree-sitter, whose `Point.column` is a **byte** offset (Textual's own docstring says so, and it hands the raw nodes back). That column is passed straight to `TextArea.get_text_range()`, which expects a **character** column. `日本語` is 9 bytes and 3 characters, so the cut lands 6 positions late.

**Scope.** Narrow — it needs non-ASCII before the semicolon *and* a non-space character right after it, since `.strip()` absorbs the off-by-one in the usual `select 1; select 2` shape. But it is reproducible on `main` today.

Found while extracting the shared statement splitter for `hsql` (#524); the fix is to own the byte→character conversion in one place so both front ends get it right.

PR #1020, opened 2026-08-07
refactor: extract the statement splitter and execution core (#524)

### LINKED ISSUE #1015: Query splitting cuts in the wrong place when a statement contains non-ASCII characters
**Before Proceeding, please acknowledge**:
- [x] I have consulted the [Troubleshooting guide](https://harlequin.sh/docs/troubleshooting/index).
- [x] I have searched Issues and Discussions in this repo.
- [x] Feature requests should be initiated as Discussions. This is a bug report.

*Found by code inspection while writing the M1 technical plan (#1014), not by hitting it in the wild — so I have no user report to attach, but the reproduction below is exact.*

**Describe the bug**

`CodeEditor.selected_queries()` splits the buffer at the wrong character when a statement contains non-ASCII text before its semicolon. Both resulting fragments are then invalid SQL, so the user gets a syntax error for a query they typed correctly.

The cause is a units mismatch. Tree-sitter reports positions with **byte** columns — Textual's own `query_syntax_tree` docstring says "(row, column byte)", and it hands the raw nodes back. But `get_text_range()` takes **character** columns. `src/harlequin/components/code_editor.py:159` passes one straight into the other:

```python
return sorted(
    (n.end_point.row, n.end_point.column) for n in query_result.get("semicolon", [])
)
```

Each non-ASCII character before the semicolon shifts the cut by `len(utf8) - 1` bytes — one position for `é`, two for each character of `日本語`.

**To Reproduce**

In the Query Editor, enter:

```sql
select 'café';select 2
```

and run the buffer (`ctrl+enter`).

No TUI needed to see it — this exercises the same two calls:

```python
from textual.document._document import Document
import tree_sitter_sql
from tree_sitter import Language, Parser, Query, QueryCursor

sql = "select 'café';select 2"
doc = Document(sql)
lang = Language(tree_sitter_sql.language())
caps = QueryCursor(Query(lang, '(";" @semicolon)')).captures(
    Parser(lang).parse(sql.encode()).root_node
)
seps = sorted((n.end_point.row, n.end_point.column) for n in caps["semicolon"])

start, out = (0, 0), []
for end in [*seps, doc.end]:
    q = doc.get_text_range(start, end).strip()
    if q:
        out.append(q)
    start = end
print(out)
```

**Expected behavior**

```
["select 'café';", 'select 2']
```

**Actual behavior**

```
["select 'café';s", 'elect 2']
```

Both are sent to the database; both are syntax errors.

**Scope**

Narrower than it first looks. It needs **both** of these on the same line:

1. at least one non-ASCII character before the semicolon, and
2. non-whitespace immediately after the semicolon.

If whitespace or a newline follows the semicolon, `.strip()` on line 64 absorbs the drift and the result is correct. And because tree-sitter columns are per-row, non-ASCII on an *earlier* line is harmless. Measured across variants:

| buffer | result |
| --- | --- |
| `select 1;select 2` | ✅ correct (ASCII control) |
| `select 'café';select 2` | ❌ `["select 'café';s", 'elect 2']` |
| `select 'café'; select 2` | ✅ correct — space absorbed by `.strip()` |
| `select 'café';\nselect 2` | ✅ correct |
| `select 'café';\nselect 'x';select 2` | ✅ correct — non-ASCII on a different row |

So it bites people who write compact one-liners and have accented text, non-Latin scripts, emoji, or typographic quotes in a string literal or an inline comment. Rare, but silent and confusing when it happens: the editor shows valid SQL and the database rejects it.

**Fix**

Convert byte columns to character columns before handing them to `get_text_range()`. The conversion needs the line's text, since it's per-row:

```python
line = self.text_input.document.get_line(row)
char_col = len(line.encode("utf-8")[:byte_col].decode("utf-8"))
```

**Contributing**

Are you interested in contributing a fix?
- [x] Yes
- [ ] Maybe
- [ ] No

It's already scheduled: PR 2 of the M1 sequence in #1014 moves statement splitting into a shared `harlequin.statements` library that both the editor and the planned `hsql` use, and owning the byte→character conversion in one place is part of that. Worth fixing independently first if M1 slips — it's a self-contained change to `_query_separators()` plus a test.

A regression test belongs in the tricky-SQL corpus that PR lands: non-ASCII before a semicolon with no space after, which fails on `main` today.

**Additional context**

What is the output of `harlequin --version`? — `harlequin, version 2.8.0`; also present on `main` at e5a69ec.

What database adapter are you using with Harlequin? — Adapter-independent. The split happens in the editor before any adapter sees the text.

What other options are you using when invoking Harlequin? — None; default profile.

tconbeer:
Toward #524. Closes #1015.

This is **PR 2 of the M1 `hsql` plan** — "`harlequin.statements` and `harlequin.query`" (`docs/plans/m1-hsql-technical-plan.md` §4). Rebased onto `main` now that #1017 has landed; this is the whole diff.

§1.1 of the plan puts the problem plainly: query execution lives entirely inside the Textual app, as two threaded workers that communicate by posting messages. There is no importable function, so `hsql` cannot call any of it, and a copy is exactly the "two products" failure the product plan warns about. Same for splitting: the live splitter is a method on a Textual widget.

1. **`harlequin.statements` — one splitter, for both front ends.** `split(text)` returns `Statement`s; `find_separators(text)` returns the `Point`s the editor slices at. Both drive `tree-sitter-sql` directly, through the same `(";" @semicolon)` query `CodeEditor` has always used — so `-f script.sql` and the editor cannot disagree about where a statement ends. `tree-sitter` and `tree-sitter-sql` become direct dependencies (§3.2, Ted's call); both are installed for everyone already, transitively via `textual[syntax]`.

2. **`CodeEditor` is refactored onto it, and its fallback branch is deleted.** With one splitter there is no second implementation to keep in sync, so `SEMICOLON_QUERY`, `_semicolon_query`, `_query_separators`, `has_shown_tree_sitter_error` and the whole `is_syntax_aware` regex branch are gone — along with the warning pointing at `harlequin.sh/docs/troubleshooting/tree-sitter`, which described a degraded mode that no longer exists. (PR 7 retires the page itself.)

3. **That fixes #1015.** Tree-sitter's `Point.column` is a **byte** offset — Textual's own docstring says so — but `get_text_range()` takes *character* columns, and `code_editor.py` passed one straight into the other. `select 'café';select 2` split into `select 'café';s` and `elect 2`, both syntax errors. Owning the conversion once, in the shared library, is what stops `hsql` from inheriting it.

4. **`harlequin.query` — the execution core.** `execute()` runs statements and `fetch()` drains one cursor into a `ResultSet`. It keeps the app's two-phase *execute all, then fetch all* shape rather than flattening it: that split is what lets the TUI show "query executed" before results materialize.

5. **`_execute_query` and `_fetch_data` become thin wrappers**, still posting the same messages. `QueriesExecuted.cursors` now maps to `ExecutedStatement` (which carries its own statement) instead of a `tuple[HarlequinCursor, str]`, and `ResultsFetched.data` becomes `.results: dict[str, ResultSet]`. The dead `Harlequin._split_query_text` — a naive `text.split(";")` nothing called — is deleted.

6. **`ResultsTable` gets the backend `fetch()` built**, rather than raw data plus a `max_rows` it applies itself. Same object either way; the row cap just moves one call earlier.

Plus the guards PR 1 established: a third import-linter contract for the two new modules, and both added to `test_import_hygiene.py` and `scripts/cold_start.py`.

| import | ms | modules | textual |
| --- | --- | --- | --- |
| `harlequin.statements` | 35 | 106 | no |
| `harlequin.query` | 177 | 292 | no |

`harlequin.query`'s cost is `textual_fastdatatable.backend` (177ms, mostly pyarrow), which it shares — the core itself adds ~19 modules on top.

**`ResultSet` wraps a `DataTableBackend` rather than reimplementing one.** `HarlequinCursor.fetchall()` returns `AutoBackendType`, which is `Any`; `create_backend()` already normalizes every shape it can be, it is the normalization the TUI uses, and per PR 1 it needs no Textual. A second normalizer would put the most drift-prone question in the codebase — what counts as a row, what counts as null, how a nested struct comes out — in two places, and the disagreement would surface as `hsql` and the TUI showing different data for the same query.

**`RowLimit` is a value object, not an `int`, because Harlequin has two different limits.** The app's `--limit` is a *soft display cap over a full fetch*, which is why it can report `100,000 of 3,412,887`. `hsql -l` will be the *hard* one — fetching a million rows to print five hundred defeats the point — and a hard limit makes truncation unknowable: `set_limit(n)` then `fetchall()` returns at most n rows and says nothing about whether an n+1th existed. So `detect_overflow` asks for `limit + 1` and the backend's existing `source_row_count > row_count` does the rest. Principle 5 of the product plan ("truncation must always be announced") is unimplementable without it, and it costs exactly one row. The app passes `detect_overflow=False` and is unaffected.

**Two deviations from the plan's signatures, both to keep the TUI's behavior:**

- **`ExecutedStatement` gained an `error` field.** The plan has `execute()` return an iterator and doesn't say how a failure is reported. Raising would lose the association between the error and the statement it belongs to, which is exactly what the app needs to post `QueryError(query_text=…)`. Carrying it keeps `execute()` a pure iterator and gives `on_error="continue"` somewhere to put failures. `KeyboardInterrupt` deliberately propagates instead — Ctrl-C is not a query error, and `"continue"` must not swallow it.
- **`ResultSet.backend` is `DataTableBackend | None`.** A cursor is allowed to return `None` for an empty result, which is not a shape `create_backend()` can normalize — and there is nothing to normalize. This matches what `ResultsTable` already did with `data=None`.

**`ResultSet.text_columns()` is not here.** The plan puts it in PR 3, with the layouts that need it.

**One deliberate behavior change, in an unreachable-in-practice path.** `create_backend()` now runs in the worker rather than in `ResultsTable.__init__`, so an adapter returning a type it cannot normalize surfaces through `_fetch_data`'s existing `except BaseException` as a query error naming the query, instead of as `DataTable.DataLoadError`. Both open a "Query Error" modal with the same header logic; the new one attributes the failure to a statement. `handle_data_load_error` is kept — `DataTable` is third-party and may post it from other paths.

**The trade the plan names, restated:** sharing the grammar means `hsql` inherits the editor's behavior *including its bugs*. The dollar-quoting bug (`$$ … ; … $$` splits inside the body) is the editor's today, identically, because it is the grammar's. Recorded as a strict `xfail` against #1019 rather than fixed, so the day it's fixed upstream the test fails and tells us. The other cost is textual-textarea's incremental tree: the shared library reparses. Measured, that's 0.02ms for a realistic buffer and 10ms for a pathological 2000-statement one, and `selected_queries()` runs on submit, not on keystroke.

Note for review
- **`tree-sitter` is pinned `>=0.25,<0.27`** to match Textual's own floor for the `QueryCursor` API, and `tree-sitter-sql >=0.3.11,<0.4` to match `textual[syntax]`. The accepted cost the plan names: Harlequin stops installing on a platform with no tree-sitter wheel.
- `test_no_tree_sitter` is rewritten rather than deleted. It asserted the degraded mode; it now asserts the invariant that replaced it — with Textual's tree-sitter monkeypatched off, `select 'a;b'; select 2` still splits correctly (the old regex fallback split it wrong), and nothing warns.
- I filed #1018 for the split bug before finding #1015, which was already open with the same diagnosis and a fuller writeup. #1018 is closed as a duplicate; everything here references #1015.

- [ ] No.
- [x] Yes; I haven't opened a PR, but the gist of the change is: `docs/troubleshooting/tree-sitter` describes a degraded mode that no longer exists and should be retired. PR 7 of the plan owns that, along with the "Headless & Agents" topic.

- [x] Yes.

- **`tests/unit_tests/test_statements.py`** — the tricky-SQL corpus: 23 scripts through `split()`, and the same 23 through `find_separators()` asserting the two agree on every one. String literals, quoted identifiers, line and block comments, escaped quotes, consecutive and bare separators, CRLF, and five flavors of non-ASCII — including #1015's exact `select 'café';select 2` and an astral-plane character. Plus the strict `xfail` for #1019.
- **`tests/functional_tests/test_query_editor.py`** — `test_selected_queries_split_on_character_columns` is the #1015 regression through the real editor, asserting it agrees with `split()`; `test_selected_queries` now goes through `find_separators`.
- **`tests/unit_tests/test_query.py`** — 25 tests: laziness, DDL, error stop/continue, `KeyboardInterrupt` propagating, a raw driver exception not escaping, zero rows vs. a `None` cursor, the row cap being distinct from the fetch limit, and truncation at under/exactly/one-over the limit **on both duckdb and sqlite**, which implement `set_limit` differently enough to matter.

Verified on the rebased branch: 283 passed, 6 skipped, 1 xfailed, 140 snapshots passed on 3.10; 3 passed on the `py12` selection. `ruff`, `mypy` and `lint-imports` all clean, and `uv.lock` is consistent with `main`'s released `textual-fastdatatable==0.16.1`. The snapshots are the safety net for the TUI refactor and none of them moved.

### DIFF
--- a/tests/functional_tests/test_query_editor.py
+++ b/tests/functional_tests/test_query_editor.py
@@ -412,7 +412,8 @@ async def test_selected_queries_split_on_character_columns(
     app: Harlequin,
     wait_for_workers: Callable[[Harlequin], Awaitable[None]],
 ) -> None:
-    """The editor and `harlequin.statements` split at the same place.
+    """Regression test for #1015: the editor and `harlequin.statements` split
+    at the same place.
 
     tree-sitter reports columns in bytes, and the editor used to feed one
     straight to `get_text_range()`, which wants characters. Any non-ASCII
--- a/tests/functional_tests/test_query_editor.py
+++ b/tests/functional_tests/test_query_editor.py
@@ -10,6 +10,7 @@ from textual.widgets.text_area import Selection
 
 from harlequin import Harlequin
 from harlequin.app import QuerySubmitted
+from harlequin.statements import find_separators, split
 
 
 @pytest.mark.asyncio
@@ -244,11 +245,20 @@ async def test_member_autocomplete(
 
 
 @pytest.mark.asyncio
-async def test_no_tree_sitter(
+async def test_splitting_survives_a_text_area_without_syntax_awareness(
     app: Harlequin,
     monkeypatch: pytest.MonkeyPatch,
     wait_for_workers: Callable[[Harlequin], Awaitable[None]],
 ) -> None:
+    """Splitting no longer depends on the editor widget's syntax tree.
+
+    It used to: with Textual's tree-sitter unavailable the editor fell back to
+    a regex over semicolons and warned the user that query splitting might
+    misbehave. `harlequin.statements` drives the grammar itself, so a text area
+    that cannot highlight still splits exactly, and there is nothing to warn
+    about -- which is the degraded mode the tree-sitter troubleshooting page
+    described.
+    """
     import textual.document._syntax_aware_document
     import textual.widgets._text_area
 
@@ -264,7 +274,8 @@ async def test_no_tree_sitter(
 
         assert app.editor is not None
         assert app.editor.text_input is not None
-        app.editor.text = "select 1; select 2"
+        # a semicolon in a string literal, which the old regex fallback split on
+        app.editor.text = "select 'a;b'; select 2"
 
         assert not app.editor.text_input.is_syntax_aware
 
@@ -276,21 +287,14 @@ async def test_no_tree_sitter(
         submitted_msg = next(
             iter(filter(lambda m: isinstance(m, QuerySubmitted), messages))
         )
-        assert submitted_msg
         assert isinstance(submitted_msg, QuerySubmitted)
-        assert len(submitted_msg.queries) == 2
+        assert submitted_msg.queries == ["select 'a;b';", "select 2"]
 
-        text_area_warning = next(
-            iter(
-                filter(
-                    lambda m: (
-                        isinstance(m, Notify) and m.notification.severity == "warning"
-                    ),
-                    messages,
-                )
-            )
-        )
-        assert text_area_warning
+        assert not [
+            m
+            for m in messages
+            if isinstance(m, Notify) and "tree-sitter" in m.notification.message.lower()
+        ]
 
 
 @pytest.mark.asyncio
@@ -359,7 +363,12 @@ async def test_selected_queries(
         await pilot.pause()
 
         # tree-sitter does not capture the semicolons in buffer order
-        assert app.editor._query_separators() == [(0, 25), (1, 27), (2, 18), (3, 15)]
+        assert find_separators(app.editor.text) == [
+            (0, 25),
+            (1, 27),
+            (2, 18),
+            (3, 15),
+        ]
 
         # the whole buffer
         app.editor.selection = Selection((0, 0), (4, 0))
@@ -396,3 +405,28 @@ async def test_selected_queries(
         await pilot.pause()
         app.editor.selection = Selection((0, 0), (0, 0))
         assert app.editor.selected_queries() == ["select 'a;b'"]
+
+
+@pytest.mark.asyncio
+async def test_selected_queries_split_on_character_columns(
+    app: Harlequin,
+    wait_for_workers: Callable[[Harlequin], Awaitable[None]],
+) -> None:
+    """The editor and `harlequin.statements` split at the same place.
+
+    tree-sitter reports columns in bytes, and the editor used to feed one
+    straight to `get_text_range()`, which wants characters. Any non-ASCII
+    before a semicolon shifted the cut, and both halves came out as syntax
+    errors -- this split into "select '日本語';select" and "2".
+    """
+    async with app.run_test() as pilot:
+        await wait_for_workers(app)
+
+        while app.editor is None:
+            await pilot.pause()
+
+        app.editor.text = "select '日本語';select 2"
+        await pilot.pause()
+        app.editor.selection = Selection((0, 0), (0, 22))
+        assert app.editor.selected_queries() == ["select '日本語';", "select 2"]
+        assert app.editor.selected_queries() == [s.sql for s in split(app.editor.text)]
--- a/tests/functional_tests/test_query_editor.py
+++ b/tests/functional_tests/test_query_editor.py
@@ -4,12 +4,9 @@ import sys
 from typing import Awaitable, Callable, List
 
 import pytest
-from textual.message import Message
-from textual.notifications import Notify
 from textual.widgets.text_area import Selection
 
 from harlequin import Harlequin
-from harlequin.app import QuerySubmitted
 from harlequin.statements import find_separators, split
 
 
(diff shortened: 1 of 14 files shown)


tests/functional_tests/test_query_editor.py
```diff
@@ -244,11 +245,20 @@ async def test_member_autocomplete(
 
 
 @pytest.mark.asyncio
-async def test_no_tree_sitter(
+async def test_splitting_survives_a_text_area_without_syntax_awareness(
```
tconbeer: Remove this test.

tconbeer: Removed in 2cd14f5, along with the three imports it was the last user of (`Message`, `Notify`, `QuerySubmitted`).

285 passed, 6 skipped, 1 xfailed; `ruff`, `mypy` clean.

---

PR #1021, opened 2026-08-07
feat: sort the Results Viewer by clicking a column header

scantinibudden:
No linked issue — built for my own use, then reworked to the design you described in review.

Clicking a column header sorts the rows the table holds by that column, in memory: ascending, then descending, then back to the order the database returned them. Only the sorted column's header carries a mark (`▲`/`▼`). Nothing is read from or written to the query.

- `ResultsTable` keeps `sort_column` / `sort_descending` and the pristine Arrow table; each state is produced from that table (`pa.Table.sort_by` by column *position*, so unnamed and duplicate columns need no naming), and "unsorted" restores it.
- The mark is appended to the header label with no style of its own, so header hover treats the name, the type and the mark alike. The sorted column's `content_width` grows to fit the mark and is restored on unsort; other headers are untouched, so existing header snapshots are unchanged.
- If the result was cut short — `fetch_truncated`, or the viewer holds fewer rows than were fetched — the first sort warns once: *"Sorted the 500 rows fetched, not the whole result."*
- Gone from the previous version: `order_by.py` and its parser, the editor write-back, the app-level `SortRequested` round trip, the `⇅` hint on every header, and the indicator sniffed from `order by` in the SQL.

This is the design from your review comment, point for point. The trade-off is the one you named: under a limit, the sort orders the rows that were fetched, not the true top-N by that column, and the warning is what tells the user so. If database-side sorting ever becomes desirable, it can be done without a parser — wrap the statement in a subquery and `order by` position — but that is a separate conversation and not what this PR does.

- [ ] No.
- [x] Yes; I haven't opened a PR, but the gist of the change is: the Results Viewer docs mention that clicking a header sorts the fetched rows (asc → desc → unsorted) and that a limited result only sorts what was fetched. Glad to open it if you want the feature.

- [x] Yes.

`test_sort_on_header_click.py`: the asc → desc → unsorted cycle (whole rows move; the third click restores the database's order; one snapshot of the sorted header), switching to another column starts ascending, and a limited 600-row result sorts its 500 fetched rows and warns exactly once.

- [x] I have added an entry to `CHANGELOG.md`, under the `[Unreleased]` section heading. (No issue to reference; see above.)


tconbeer:
Thanks for this PR. I think it needs more work.

1. I don't like the idea of rewriting queries, especially not with a sql parser on board. As of 0.17.1, the results back-end detects truncation (from limits). You should just sort the in-memory arrow table, and maybe if it's truncated show a warning to the user. Rewriting has quite a few pitfalls - one is that different adapters handle unnamed columns differently. For example, in duckdb:
```sql
select a, a+1
from range(1000) as t(a)
```
Clicking on the second column (displayed as `(a + 1)`) rewrites to:
```sql
select a, a+1
from range(1000) as t(a)
order by "(a + 1)" asc
```

But that creates an immediate error:
```
Binder Error: Referenced column "(a + 1)" not found in FROM clause!
Candidate bindings: "a"
```

2. I find the UI awkward. The `⇅` on every cell really clutters the header, so I think that has to go. The hover state is also wrong... it highlights the column type only but not the sort indicator.

3. There is no unsort

4. I don't think sniffing the query to populate the sort indicator is a good idea. Among other problems, the simple/common `order by 1` does not work.

scantinibudden:
Thanks — reworked to your four points, as a fresh branch on `main`:

1. No query rewriting. The header click sorts the in-memory Arrow table by column position, so `select a, a+1 from range(1000) as t(a)` sorts by the second column without ever naming it. A result cut by a limit warns once that only the fetched rows were sorted. `order_by.py` and the editor write-back are gone.
2. The `⇅` is gone; only the sorted header carries `▲`/`▼`. The mark has no style of its own, so hover highlights name, type and mark together.
3. Third click unsorts, back to the order the database returned.
4. Nothing is read from the SQL; sort state lives on the widget.

Diff is now +433, almost all of it tests and one snapshot; the widget change is ~80 lines in `results_viewer.py`. Existing snapshots are unchanged since unsorted headers look as they do today.

PR #1022, opened 2026-08-07
refactor: decouple export from the results table, add text layouts (#524)

tconbeer:
PR 3 of the M1 hsql plan. Nothing arranges a result set as text today, and the
one thing that writes one to a file takes a Textual widget -- so a headless
front end has no way to emit either.

- export.copy(ResultsTable, ...) becomes write_file(pa.Table, ...), plus
  write_stream() for stdout. Every writer underneath produces a file, so the
  stream path writes a temp file and copies it out in binary: /dev/stdout would
  save the copy on Linux and doesn't exist on Windows, and duckdb's \n must not
  become \r\n on the way out
- tsv, jsonl/ndjson and arrow are the csv, json and feather writers under
  different default options, not new writers. A caller's explicit option beats
  the format's default, which is what keeps the export dialog's array flag
  working now that `json` means an array of row objects
- an option left empty is an option not set (the dialog renders every option of
  a format whether or not it was touched), but False is a real value -- so only
  None and "" are dropped, and header=False reaches the writer
- column de-duplication moves out of the dialog and into write_file(), where it
  belongs: it is duckdb that cannot take `select 1 as a, 2 as a`, not the widget
- harlequin.layout arranges a result as table, markdown or vertical text.
  LayoutOptions are independent switches, so psql's -t/-A/-tA algebra falls out
  rather than being special-cased
- ResultSet.arrow_table() is the rows kept, under the names the cursor
  described; text_columns() is that cast to VARCHAR in duckdb. Not str(), and
  not the data table's cell_formatter, which is display formatting and would put
  a locale's thousands separators in a numeric column

Golden files cover every format against one result set holding decimals, blobs,
intervals, structs, maps, nulls and unicode, and a test asserts -F table and
-F csv agree cell for cell. No user-visible change: neither module has a
consumer until the CLI lands.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01MWfadUkv68QV8DaCKKNxAP


src/harlequin/export.py
```diff
@@ -1,94 +1,165 @@
+"""Writing a result set to a file, serialized by duckdb or pyarrow.
+
+Nothing here serializes a value itself. duckdb's writers own the rendering of
```
tconbeer: This docstring relitigates the design. Simplify. Be descriptive not defensive.

src/harlequin/query.py
```diff
@@ -106,6 +108,109 @@ def row_count(self) -> int:
     # dict per row (1.8s vs 0.34s over 100k rows) and materializes the whole
     # result as Python objects, which is what `--limit` exists to avoid.
 
+    def arrow_table(self) -> pa.Table:
```
tconbeer, line 115: This is really messy. Can we upstream a change to textual-fastdatatable instead?

src/harlequin/query.py
```diff
@@ -106,6 +108,109 @@ def row_count(self) -> int:
...
+
+    names = data.column_names
+    # duckdb refuses an Arrow table with duplicate field names, and a column
+    # name is otherwise an identifier we would have to quote correctly. Neither
+    # is worth a rendezvous with quoting rules, so the projection is written
+    # against positional names and the originals are put back afterwards.
+    positional = [f"c{i}" for i in range(len(names))]
+    relation = duckdb.from_arrow(data.rename_columns(positional))
+    projection = ", ".join(f"CAST({name} AS VARCHAR) AS {name}" for name in positional)
+    projected = relation.project(projection).arrow()
```
tconbeer: I think later versions of duckdb got more strict in the relation API and expect Statements here, not a string. Can you check? If Statememts are supported by older duckdb I would prefer them over string interpolation here.

src/harlequin/query.py
```diff
@@ -106,6 +108,109 @@ def row_count(self) -> int:
...
+                f"Could not serialize this result set with duckdb ({e}); "
+                "falling back to Python's str(), which may render some values "
+                "differently than an exported file would.",
+                RuntimeWarning,
+                stacklevel=2,
+            )
+            return _cast_to_text_with_str(data)
+
+
+def _empty_table(names: Sequence[str]) -> pa.Table:
```
tconbeer: This is definitely the right way to handle an empty result set. Elsewhere I think we raise if someone tries to export an empty table? Can you fix that so we export an empty table instead of raising?

src/harlequin/query.py
tconbeer: Also please note that Harlequin only uses the ArrowBackend so you don't have to handle the PolarsBackend

tests/unit_tests/test_export.py
```diff
@@ -0,0 +1,137 @@
...
+
+import pyarrow as pa
+import pytest
+
+from harlequin.exception import HarlequinCopyError
+from harlequin.export import file_format_names, write_file, write_stream
+
+TEXT_FORMATS = ["csv", "tsv", "json", "jsonl", "ndjson"]
+BINARY_FORMATS = ["parquet", "orc", "feather", "arrow"]
+
```
tconbeer, line 29: I would like to also add tests for each option on each format, like the timestamp format option for Jason, etc.

PR #1023, opened 2026-08-09
refactor: share the profile/CLI precedence rule, load adapters one at a time (#524)

tconbeer:
Part of #524 — PR 4 of the M1 `hsql` plan (`docs/plans/m1-hsql-technical-plan.md`, §3.4–3.5). Not `Closes`: #524 stays open until PRs 5–8 land the command itself.

Two extractions, both behavior-neutral for the `harlequin` command.

- `config.merge_profile_with_cli()` is the precedence rule that used to sit inside `inner_cli`: **a CLI value beats the profile only if the user actually typed it**, since an option resting on its default carries no intent. It takes the *names* of the options that were set rather than a click `Context`. The empty-`conn_str` carve-out moves with it and is now stated in a docstring rather than implied by an `elif`.
- `plugins.adapter_names()` lists every installed adapter without importing any of them; `plugins.load_adapter(name)` imports exactly one. `harlequin`'s CLI is refactored onto the first and keeps `load_adapter_plugins()`, since its `--help` describes every installed adapter.

Neither new plugins function has a consumer yet — they're what makes the two-phase parse in PR 5 possible.

*Names, not a `Context`.* Taking `explicitly_set: Container[str]` means the rule can be applied — and tested — without building a command first, which is what `hsql` needs: it resolves its adapter before it has a full command to get a `Context` from. The tradeoff is that the caller now derives the set, which is one comprehension over `ctx.get_parameter_source()`.

*Why `load_adapter` raises where `load_adapter_plugins` warns.* Scanning can afford to skip an adapter that won't import — the app still runs with the ones that did. A caller that named an adapter has nothing to fall back to, so a `HarlequinConfigError` (exit 2, with the underlying `ImportError` in the message) is the honest answer. Alternative considered: returning `None` and letting callers decide, which just moves the same decision to every call site.

*Why the split at all.* `ep.load()` on every installed adapter is the only unbounded cost in start-up, and it's paid to build a command that will use exactly one. Two subprocess tests hold that line: `adapter_names()` imports no `harlequin_*` package, and `load_adapter("duckdb")` imports exactly `harlequin_duckdb`.

*One incidental improvement:* the merge no longer mutates the profile dict in place. It was a live reference into the merged config files, and the code downstream `pop()`s from it.

- [x] No. Nothing user-facing changes; the `hsql` docs topic is PR 7 of this plan. `AGENTS.md` is updated with the precedence rule and the three loading grains.

- [x] Yes. 8 new in `test_plugins.py`, 6 in `test_config.py`, and `test_conn_str_overrides_the_profile` in `test_cli.py` — a precedence case the existing tests never covered (a CLI `conn_str` beating a profile's, with the rest of the profile intact). The rest of the refactor is covered by the existing `test_cli.py`. Full suite 451 passed / 6 skipped, py12 3 passed, mypy clean, import contracts kept.

- [ ] ~I have added an entry to `CHANGELOG.md`~ — deliberately none, [per review](https://github.com/tconbeer/harlequin/pull/1023#discussion_r3744662817): nothing in this PR is visible to a user of the `harlequin` command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Refactoring
 
+- The rule that decides whether a CLI option or a config profile wins is now `harlequin.config.merge_profile_with_cli()`, rather than living inside the `harlequin` command ([#524](https://github.com/tconbeer/harlequin/issues/524)). It takes the names of the options the user actually typed instead of a click `Context`, so the precedence rule can be shared by a command that builds itself differently, and tested without the CLI framework. `harlequin.plugins` gains `adapter_names()`, which lists every installed adapter without importing any of them, and `load_adapter()`, which imports exactly one and fails loudly if it can't. Both are for the planned headless CLI, which uses one adapter per invocation and shouldn't pay to import the rest; the `harlequin` command still loads them all, since its `--help` describes every one.
 - Writing a result set to a file no longer requires a Textual widget ([#524](https://github.com/tconbeer/harlequin/issues/524)). `harlequin.export.write_file()` and `write_stream()` take an Arrow table and write it to a path or an open binary stream, and `harlequin.layout` arranges a result set as `table`, `markdown` or `vertical` text. Both are for the planned headless CLI and have no user-facing surface yet; the Data Exporter behaves as before. `export.py` also gains `tsv`, `jsonl`/`ndjson` and `arrow` as option variants of formats it already wrote.
 
 ### Performance
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@ All notable changes to this project will be documented in this file.
 
 ### Refactoring
 
-- The rule that decides whether a CLI option or a config profile wins is now `harlequin.config.merge_profile_with_cli()`, rather than living inside the `harlequin` command ([#524](https://github.com/tconbeer/harlequin/issues/524)). It takes the names of the options the user actually typed instead of a click `Context`, so the precedence rule can be shared by a command that builds itself differently, and tested without the CLI framework. `harlequin.plugins` gains `adapter_names()`, which lists every installed adapter without importing any of them, and `load_adapter()`, which imports exactly one and fails loudly if it can't. Both are for the planned headless CLI, which uses one adapter per invocation and shouldn't pay to import the rest; the `harlequin` command still loads them all, since its `--help` describes every one.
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Refactoring
 
+- The rule that decides whether a CLI option or a config profile wins is now `harlequin.config.merge_profile_with_cli()`, rather than living inside the `harlequin` command ([#524](https://github.com/tconbeer/harlequin/issues/524)). It takes the names of the options the user actually typed instead of a click `Context`, so the precedence rule can be shared by a command that builds itself differently, and tested without the CLI framework. `harlequin.plugins` gains `adapter_names()`, which lists every installed adapter without importing any of them, and `load_adapter()`, which imports exactly one and fails loudly if it can't. Both are for the planned headless CLI, which uses one adapter per invocation and shouldn't pay to import the rest; the `harlequin` command still loads them all, since its `--help` describes every one.
```
tconbeer: Remove this changelog entry. It's irrelevant to users.

src/harlequin/config.py
```diff
@@ -120,6 +120,41 @@ def get_config_for_profile(
     return profile, keymaps
 
 
+def merge_profile_with_cli(
+    profile: Profile,
+    cli_values: Mapping[str, Any],
+    explicitly_set: Container[str],
+) -> Profile:
+    """
+    Layer the values a user passed on the command line over the profile loaded
```
tconbeer: Make this docstring much shorter.

pyproject.toml
```diff
@@ -302,6 +303,20 @@ ignore_imports = ["harlequin.exception -> harlequin.colors"]
 unmatched_ignore_imports_alerting = "error"
 
 
+[[tool.importlinter.contracts]]
```
tconbeer: I don't think we want or need this rule. The hsql CLI will use click also.

src/harlequin/plugins.py
```diff
@@ -2,13 +2,54 @@
...
 from harlequin.adapter import HarlequinAdapter
 from harlequin.exception import HarlequinConfigError
 from harlequin.keymap import HarlequinKeyMap
 
 
+def adapter_names() -> list[str]:
+    """
+    The name of every installed adapter, without importing any of them.
+
+    Reading entry point names costs nothing; `ep.load()` imports the adapter and
```
tconbeer: You don't need to discuss perf in every doc string.

tconbeer (commented):
Fix in this PR and add a note to AGENTS that all docstrings and comments must be as concise as possible. They need not relay the design or instructions given when the function was written. Docstrings should describe the function in one sentence, and may add another sentence or two about why the function operates the way it does.

PR #1025, opened 2026-08-09
feat: add hsql, the headless CLI (#524)

### LINKED ISSUE #524: Add `-c`/`-cmd`/`--command` option to run sql command from the command-line

tconbeer:
Part of #524. This is **PR 5** of [the M1 technical plan](https://github.com/tconbeer/harlequin/blob/main/docs/plans/m1-hsql-technical-plan.md) — the command itself, on top of the plumbing that landed in 2.9 (PRs 1–4). It does **not** close #524; PRs 6 (discoverability hints), 7 (docs) and 8 (the eval suite) remain.

A new `hsql` console script, `hsql = "harlequin.hsql:main"`, in four modules:

| module | owns |
| --- | --- |
| `hsql/cli.py` | the click command, two-phase adapter resolution, the run |
| `hsql/output.py` | choosing between `harlequin.layout` and `harlequin.export`, and writing |
| `hsql/diagnostics.py` | everything on stderr, and `ExitCode` |
| `hsql/__init__.py` | `main()` |

It owns no query logic: `harlequin.query` runs the statements, `harlequin.layout` and `harlequin.export` write the results. `harlequin` itself gains nothing — no new flags, no behavior change.

The full M1 surface: `-c`/`-f`/stdin, `-o`, `-F` and the `--csv`/`--json`/`--jsonl`/`--markdown`/`--vertical` shorthands, `-t`/`-A`/`--no-header`/`--null-string`, `-P`, `-a`, `--config-path`, `-l` defaulting to 500, `--result`, `--on-error`, `--color`/`NO_COLOR`, `--stats`, exit codes, truncation notices, plain errors.

The parts that are a contract rather than an implementation:

- **stdout is data; stderr is narration.** Row counts, timings, truncation notices and errors all go to stderr, and an error is one plain line — `hsql: error: ...`, no panel.
- **Exit codes**: 0 ok, 1 query, 2 usage/config, 3 connection, 130 interrupt. (4 is reserved for `--timeout` in M2.)
- **`--limit` defaults to 500 and is the _hard_ limit**, so truncation is knowable only by fetching one row more than we keep — and it is always announced, `-t` included. `-l 0` counts exactly.
- **Deterministic output**: identical bytes whether stdout is a pipe, a file or a terminal, `\n` on every platform, unaffected by `LC_ALL`. Color is off unless asked for.
- **A format that cannot hold two result sets says so and exits 2**, rather than writing a second header into the same csv.

Mostly as §3.6 of the technical plan specifies. The decisions worth a reviewer's attention:

- **`--help` imports no adapter at all.** It lists the adapter-agnostic surface, the formats, the exit codes and the _names_ of what is installed, and points at `hsql --help -a <adapter>`. A first pass parses `-a`/`-P`/`--config-path` with a throwaway resilient parser to name the one adapter in play; a second builds the real command with only that adapter's options. Measured on the CI-ish container: `hsql --help` 147ms, `hsql -c "select 1"` against DuckDB 403ms, and `harlequin.hsql.cli` imports in 113ms with no Textual and no pyarrow (`harlequin.query` is deferred into the callback).
- **hsql's own flags win a collision** with an adapter's short alias — `_adapter_params` strips the colliding spelling rather than letting a plug-in shadow a documented flag. It stays visible in `hsql --help -a <adapter>`. Nothing installed collides today; a future adapter declaring `-c` or `-f` would.
- **A shared profile's `limit` applies.** Config discovery, profiles and precedence are identical to the IDE's, so a profile with `limit = 100000` gives `hsql` a hard 100k fetch. `-l` overrides it. The alternative — ignoring the key because the two commands mean different things by it — breaks "profiles supply defaults", so I kept it and called it out in the changelog instead.
- **IDE-only profile keys are dropped** (`theme`, `keymap_name`, `show_files`, `show_s3`, `locale`, `no_download_tzdata`) rather than handed to the adapter as options it never declared. `locale` especially: `hsql` must not call `set_locale()`, or output would vary with `LC_ALL`.
- **Bare `hsql` exits 2 with the help on stderr.** That is click 8.3's `no_args_is_help`, and I kept it deliberately: help on stdout would let `hsql | …` carry help text where data belongs.
- **`-c` and `-f` run in the order the options appear.** click hands a repeatable option all of its values at once, so ordering is by each option's first appearance — `-f setup.sql -c "select …"` works, and alternating between the two (`-c A -f X -c B`) is the one lossy case.
- **`-o PATH` goes through the same binary stream stdout does**, rather than handing duckdb the path. It costs a temp-file copy on a large export; it buys the `-o PATH` == `> PATH` property by construction, and it is what lets `jsonl` hold several result sets in one file.

One change outside PR 5's brief: **`harlequin.exception` no longer imports `rich` at module scope.** It is on every headless path via `harlequin.config`, and `hsql --help` was paying ~20ms for a panel it never draws.

Two guards, because the boundary is the point of shipping this as a second command: a new import-linter contract (`hsql does not reach the TUI`) that fails if `harlequin.hsql` ever reaches Textual, questionary, sqlfmt or `harlequin.cli`, and `test_import_hygiene.py` proving it at run time in a subprocess.

**Not in this PR**, per §7 of the plan: `--read-only`, `--timeout`, `--dry-run`, `--single-transaction`, every subcommand, and streaming. Also worth noting for release time: `packaging/hsql/README.md` still says the command isn't here yet — true until the release that ships it, at which point the metapackage wants a dependency bump and a republish.

- [ ] No.
- [x] Yes; I haven't opened a PR. The "Headless & Agents" topic is PR 7 of the M1 plan and lands separately. It needs: the flag reference, the format table, the exit-code table, a "differences from psql" note, and one sentence on `--limit` meaning a soft display cap in the IDE and a hard fetch limit here.

- [x] Yes. `tests/unit_tests/test_hsql.py`, 109 tests: the psql flag algebra, every format, `-tAc` returning a bare number, stdout purity (identical bytes with and without `--stats`; a query error leaves stdout completely empty), truncation under/at/over the limit on both bundled adapters, `--result` and the multi-result-set error, `--on-error`, `-o PATH` vs `> PATH` byte-for-byte, profiles, the exit-code mapping, and — in real subprocesses — LF on every platform, no variation under `LC_ALL=de_DE.UTF-8`, and no variation between a pipe and a file.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AdZHNXduKLEvFjNAxo3Swd)_

### DIFF
--- a/src/harlequin/hsql/cli.py
+++ b/src/harlequin/hsql/cli.py
@@ -1,0 +1,741 @@
+"""The `hsql` command: run SQL against any Harlequin adapter, and exit.
+
+Two-phase parsing is what keeps start-up cheap. The first pass reads `-a`, `-P`
+and `--config-path` well enough to name the one adapter an invocation will use,
+without importing any of them; the second builds the real command with that
+adapter's connection options on it. An invocation only ever uses one adapter,
+so for execution this is not a compromise.
+
+`--help` is the exception, and works the other way round: with no adapter named
+it renders the adapter-agnostic surface plus the *names* of what is installed,
+importing nothing at all. `hsql --help -a postgres` imports postgres alone.
+That keeps the first thing a caller reads small and stable, and keeps it true
+for every adapter rather than for whichever one is the default.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import os
+import sys
+import time
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, BinaryIO, Iterator, Mapping, Sequence
+
+import click
+
+from harlequin.config import get_config_for_profile, merge_profile_with_cli
+from harlequin.exception import (
+    HarlequinConfigError,
+    HarlequinConnectionError,
+    HarlequinError,
+)
+from harlequin.hsql import diagnostics, output
+from harlequin.hsql.diagnostics import ExitCode
+from harlequin.plugins import adapter_names, load_adapter
+
+if TYPE_CHECKING:
+    from harlequin.adapter import HarlequinAdapter, HarlequinConnection
+    from harlequin.layout import LayoutOptions
+    from harlequin.query import ExecutedStatement, OnError, RowLimit
+    from harlequin.statements import Statement
+
+PROGRAM = "hsql"
+
+# Repeated from harlequin.cli rather than imported: that module builds the IDE's
+# command and reaches the app, the config wizard and the keymap editor with it.
+DEFAULT_ADAPTER = "duckdb"
+
+DEFAULT_FORMAT = "table"
+
+DEFAULT_LIMIT = 500
+"""Small on purpose. The IDE's 100,000 is right for a viewport and wrong here.
+
+`hsql -l` is also the *hard* limit -- `cursor.set_limit()`, so fewer rows leave
+the database -- where the IDE's is a soft cap over a full fetch. Same spelling,
+different promise, and the docs say so.
+"""
+
+SHORTHANDS = {
+    "csv": "csv",
+    "json": "json",
+    "jsonl": "jsonl",
+    "markdown": "markdown",
+    "vertical": "vertical",
+}
+"""`--csv` and friends, as flags, spelling the `-F` they stand for."""
+
+TUI_ONLY_KEYS = (
+    "theme",
+    "keymap_name",
+    "show_files",
+    "show_s3",
+    "locale",
+    "no_download_tzdata",
+)
+"""Profile keys that belong to the IDE.
+
+A profile is shared between the two commands, so a profile written for the IDE
+has to work here -- these are dropped rather than handed to the adapter as
+options it never declared. `locale` in particular is one hsql must ignore: the
+IDE sets it to group digits for a human, and output that varied with `LC_ALL`
+would be output a caller could not predict.
+"""
+
+SOURCES = f"{__name__}.sources"
+"""Context key under which `-c` and `-f` record themselves, in order."""
+
+
+def build_cli(args: Sequence[str] | None = None) -> click.Command:
+    """Build the hsql command, importing at most one adapter to do it."""
+    argv = list(args) if args is not None else sys.argv[1:]
+    adapter_name, wants_help = _preflight(argv)
+    installed = adapter_names()
+
+    adapter_cls: type[HarlequinAdapter] | None = None
+    adapter_error: HarlequinConfigError | None = None
+    if adapter_name is not None:
+        try:
+            adapter_cls = load_adapter(adapter_name)
+        except HarlequinConfigError as e:
+            adapter_error = e
+
+    @click.command(
+        name=PROGRAM,
+        no_args_is_help=True,
+        epilog=_epilog(installed, adapter_name),
+    )
+    @click.version_option(package_name="harlequin", prog_name=PROGRAM)
+    @click.argument("conn_str", nargs=-1)
+    @click.option(
+        "-c",
+        "--command",
+        multiple=True,
+        callback=_record_source,
+        help="Execute SQL. Repeatable.",
+    )
+    @click.option(
+        "-f",
+        "--file",
+        multiple=True,
+        callback=_record_source,
+        metavar="PATH",
+        help="Execute SQL from a file, or from stdin for `-`. Repeatable.",
+    )
+    @click.option(
+        "-o",
+        "--output",
+        type=click.Path(dir_okay=False, path_type=Path),
+        metavar="PATH",
+        help="Write results to PATH instead of stdout.",
+    )
+    @click.option(
+        "-F",
+        "--format",
+        default=DEFAULT_FORMAT,
+        show_default=True,
+        metavar="NAME",
+        type=click.Choice(output.format_names(), case_sensitive=False),
+        help="Output format. See below for the list.",
+    )
+    @click.option("--csv", is_flag=True, help="Shorthand for -F csv.")
+    @click.option("--json", is_flag=True, help="Shorthand for -F json.")
+    @click.option("--jsonl", is_flag=True, help="Shorthand for -F jsonl.")
+    @click.option("--markdown", is_flag=True, help="Shorthand for -F markdown.")
+    @click.option("--vertical", is_flag=True, help="Shorthand for -F vertical.")
+    @click.option(
+        "-t",
+        "--tuples-only",
+        is_flag=True,
+        help="Rows only: no header, no footer. As in psql.",
+    )
+    @click.option(
+        "-A", "--no-align", is_flag=True, help="Unaligned output. As in psql."
+    )
+    @click.option(
+        "--no-header", is_flag=True, help="Omit the header row, keeping other chrome."
+    )
+    @click.option(
+        "--null-string",
+        metavar="TEXT",
+        help="Render NULL as TEXT. Defaults to NULL for text formats, empty for csv.",
+    )
+    @click.option(
+        "-P",
+        "--profile",
+        help=(
+            "Load a profile from an available config file. Options passed here take "
+            "precedence over the profile's. Use the profile named None for "
+            "Harlequin's defaults instead of the config file's default profile."
+        ),
+    )
+    @click.option(
+        "-a",
+        "--adapter",
+        default=DEFAULT_ADAPTER,
+        show_default=True,
+        metavar="NAME",
+        type=click.Choice(installed, case_sensitive=False),
+        help="The installed adapter plug-in to connect with.",
+    )
+    @click.option(
+        "--config-path",
+        type=click.Path(exists=True, dir_okay=False, resolve_path=True, path_type=Path),
+        envvar="HARLEQUIN_CONFIG_PATH",
+        show_envvar=True,
+        metavar="PATH",
+        help="Read this config file instead of the ones hsql discovers.",
+    )
+    @click.option(
+        "-l",
+        "--limit",
+        default=DEFAULT_LIMIT,
+        show_default=True,
+        metavar="N",
+        type=click.IntRange(min=0),
+        help="Maximum rows per result set. 0 for no limit.",
+    )
+    @click.option(
+        "--result",
+        default="all",
+        show_default=True,
+        metavar="all|last|N",
+        help="Which result set(s) to emit.",
+    )
+    @click.option(
+        "--on-error",
+        default="stop",
+        show_default=True,
+        type=click.Choice(["stop", "continue"]),
+        help="What to do when a statement fails.",
+    )
+    @click.option(
+        "--stats", is_flag=True, help="Write a one-line JSON summary to stderr."
+    )
+    @click.option(
+        "--color",
+        default="never",
+        show_default=True,
+        type=click.Choice(["auto", "always", "never"]),
+        help="Color text output. `auto` follows the terminal and NO_COLOR.",
+    )
+    @click.pass_context
+    def inner_cli(
+        ctx: click.Context,
+        profile: str | None,
+        config_path: Path | None,
+        **kwargs: Any,
+    ) -> None:
+        """Execute SQL and exit.
+
+        CONN_STR: one or more connection strings, or paths to local db files.
+        """
+        if adapter_error is not None:
+            diagnostics.report_error(adapter_error)
+            ctx.exit(ExitCode.USAGE)
+
+        try:
+            config, _ = get_config_for_profile(
+                config_path=config_path, profile_name=profile
+            )
+        except HarlequinConfigError as e:
+            diagnostics.report_error(e)
+            ctx.exit(ExitCode.USAGE)
+
+        explicitly_set = {
+            k
+            for k in kwargs
+            if ctx.get_parameter_source(k) != click.core.ParameterSource.DEFAULT
+        }
+        values: dict[str, Any] = dict(
+            merge_profile_with_cli(
+                profile=config, cli_values=kwargs, explicitly_set=explicitly_set
+            )
+        )
+        for key in TUI_ONLY_KEYS:
+            values.pop(key, None)
+
+        # every key hsql owns comes off here; whatever is left is the adapter's
+        conn_str: Sequence[str] | str = values.pop("conn_str", tuple())
+        if isinstance(conn_str, str):
+            conn_str = (conn_str,)
+        adapter: str = values.pop("adapter", DEFAULT_ADAPTER)
+        destination: Path | None = values.pop("output", None)
+        result_spec: str = str(values.pop("result", "all"))
+        raw_on_error = str(values.pop("on_error", "stop"))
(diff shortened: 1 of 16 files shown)


src/harlequin/hsql/cli.py
```diff
@@ -0,0 +1,741 @@
...
+    from harlequin.adapter import HarlequinAdapter, HarlequinConnection
+    from harlequin.layout import LayoutOptions
+    from harlequin.query import ExecutedStatement, OnError, RowLimit
+    from harlequin.statements import Statement
+
+PROGRAM = "hsql"
+
+# Repeated from harlequin.cli rather than imported: that module builds the IDE's
+# command and reaches the app, the config wizard and the keymap editor with it.
+DEFAULT_ADAPTER = "duckdb"
```
tconbeer: Maybe this should be in harlequin.config instead, and also imported by the TUI?

src/harlequin/hsql/cli.py
```diff
@@ -0,0 +1,741 @@
...
+SHORTHANDS = {
+    "csv": "csv",
+    "json": "json",
+    "jsonl": "jsonl",
+    "markdown": "markdown",
+    "vertical": "vertical",
+}
+"""`--csv` and friends, as flags, spelling the `-F` they stand for."""
+
+TUI_ONLY_KEYS = (
```
tconbeer: Could this also fit in the config module?

src/harlequin/hsql/cli.py
```diff
@@ -0,0 +1,741 @@
...
+would be output a caller could not predict.
+"""
+
+SOURCES = f"{__name__}.sources"
+"""Context key under which `-c` and `-f` record themselves, in order."""
+
+
+def build_cli(args: Sequence[str] | None = None) -> click.Command:
+    """Build the hsql command, importing at most one adapter to do it."""
+    argv = list(args) if args is not None else sys.argv[1:]
```
tconbeer: When would sys.argv[1:] not be empty if args is None? Best to just default to [] for args is None?

src/harlequin/hsql/cli.py
```diff
@@ -0,0 +1,741 @@
...
+            ctx.exit(ExitCode.USAGE)
+
+        connection: HarlequinConnection
+        try:
+            connection = adapter_instance.connect()
+        except HarlequinConnectionError as e:
+            diagnostics.report_error(e)
+            ctx.exit(ExitCode.CONNECTION)
+
+        run_started = time.monotonic()
```
tconbeer: Should this be automated in _Run?

src/harlequin/hsql/cli.py
```diff
@@ -0,0 +1,741 @@
...
+
+
+def _preflight(args: Sequence[str]) -> tuple[str | None, bool]:
+    """Name the adapter an invocation will use, importing none of them.
+
+    Best-effort by design: a config file it cannot read is not reported here,
+    because the real parse is about to read the same file and has somewhere to
+    report it. All this decides is whose connection options get attached.
+    """
+    wants_help = not args or "--help" in args
```
tconbeer: Can't we use click's parser for this also? Shouldn't -h etc want help?

src/harlequin/hsql/cli.py
```diff
@@ -0,0 +1,741 @@
...
+    name: str | None = ctx.params.get("adapter")
+    if name is None:
+        with contextlib.suppress(HarlequinConfigError, OSError):
+            found, _ = get_config_for_profile(
+                config_path=ctx.params.get("config_path"),
+                profile_name=ctx.params.get("profile"),
+            )
+            name = found.get("adapter")
+
+    installed = {n.lower(): n for n in adapter_names()}
```
tconbeer: Better to pass this in as an arg, so the cli command only builds this list once. Easier to test too

src/harlequin/hsql/cli.py
```diff
@@ -0,0 +1,741 @@
...
+        allow_extra_args=True,
+        allow_interspersed_args=True,
+    )
+    with contextlib.suppress(click.ClickException, ValueError):
+        probe.parse_args(ctx, list(args))
+
+    name: str | None = ctx.params.get("adapter")
+    if name is None:
+        with contextlib.suppress(HarlequinConfigError, OSError):
+            found, _ = get_config_for_profile(
```
tconbeer: Can we only do this once instead of loading in preflight and loading again later?

src/harlequin/hsql/cli.py
```diff
@@ -0,0 +1,741 @@
...
+        return installed.get(str(name).lower()), wants_help
+    if wants_help:
+        return None, wants_help
+    return installed.get(DEFAULT_ADAPTER), wants_help
+
+
+def _adapter_params(
+    adapter_cls: type[HarlequinAdapter], reserved: set[str], taken: set[str]
+) -> list[click.Parameter]:
+    """One adapter's connection options, as click parameters.
```
tconbeer: I like this idea but I do not like the implementation. Maybe each option should grow a method or property to make this simpler?

src/harlequin/hsql/cli.py
```diff
@@ -0,0 +1,741 @@
...
+
+    The default is `never`, so the bytes do not depend on what stdout happens
+    to be. `auto` opts into that dependence, and honors NO_COLOR; `always` is
+    the caller saying they meant it. A file never gets escapes either way.
+    """
+    if destination is not None or when == "never":
+        return False
+    if when == "always":
+        return True
+    return sys.stdout.isatty() and not os.environ.get("NO_COLOR")
```
tconbeer: os.getenv() is nicer.

tests/unit_tests/test_hsql.py
```diff
@@ -0,0 +1,602 @@
...
+
+@pytest.fixture(autouse=True)
+def no_discovered_config(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Keep the machine running the tests out of them.
+
+    Config discovery walks the home directory and the cwd, so without this a
+    developer's own `.harlequin.toml` decides what these assert.
+    """
+    for search in ("_search_home", "_search_config", "_search_cwd"):
+        monkeypatch.setattr(f"harlequin.config.{search}", list)
```
tconbeer: I think there is a better way to do this in the harlequin test_cli module and functional tests. Let's unify to one approach.

tests/unit_tests/test_hsql.py
```diff
@@ -0,0 +1,602 @@
...
+)
+"""Ten rows, in a spelling both bundled adapters accept."""
+
+LAYOUTS = ["table", "markdown", "md", "vertical"]
+TEXT_FILES = ["csv", "tsv", "json", "jsonl", "ndjson"]
+BINARY_FILES = ["parquet", "orc", "arrow"]
+FILE_FORMATS = TEXT_FILES + BINARY_FILES
+
+# the console script's entry point, for the assertions that need a real process
+HSQL_MAIN = (
```
tconbeer: This is gross. Only use the Click CLI runner. Tests that need to inspect something else aren't testing the CLI they are testing output.py etc.

src/harlequin/hsql/cli.py
tconbeer: `sys.argv[1:]` is never empty here in practice — `args=None` *is* the production path. `main()` calls `build_cli()` with nothing, and the preflight needs the real argv to find `-a`/`-P` and decide whose connection options go on the command. Defaulting to `[]` would make `hsql -a postgres --host db.internal` fail with "No such option: --host", because no adapter would ever be loaded.

But you're right that the line reads backwards from that — it looks defensive when it's load-bearing. Fixed by removing the default entirely: `build_cli(argv)` is now required, and `main()` hands the same list to it and to click.

```python
argv = sys.argv[1:]
code = build_cli(argv).main(args=argv, prog_name=PROGRAM, standalone_mode=False)
```

One behavior, no branch, and it closes a bug class I'd left open: before this, a caller could pass one set of args to `build_cli()` and a different set to `.main()`, and the command would be built for one invocation and parse another.

---

src/harlequin/hsql/cli.py
tconbeer: Split these two, because I did the first and deliberately didn't do the second.

**click's parser: yes.** The probe now declares `--help` as a plain (non-eager) flag, so it's parsed rather than string-matched:

```python
click.Option(["--help"], is_flag=True),
...
wants_help = not argv or bool(ctx.params.get("help"))
```

**`-h`: I'd rather not, and it's not really about help.** `-h` is `--host` in psql, and both adapters that declare it here agree — `harlequin-postgres` and `harlequin-mysql` each ship `-h` for their host option. Since hsql's own flags win a collision, claiming `-h` would silently strip it from both, and `hsql --help -a postgres` would stop showing the spelling every postgres user has in their fingers. Given hsql's whole positioning is "be boring and psql-shaped," taking psql's `--host` flag for something else seems like the wrong trade.

Verified on the branch — postgres keeps it:

```
$ hsql --help -a postgres
  -h, --host TEXT                 Specifies the host name of the machine on ...
```

psql's own answer for a short help flag is `-?`, which we could add cheaply if you want one; it's awkward to type in zsh without quoting, which is why I left it out rather than picked it. Pinned either way now, so the decision doesn't drift:

```python
def test_hsql_does_not_claim_dash_h() -> None:
    ...
    assert "--help" in opts
    assert "-h" not in opts
```

---

src/harlequin/hsql/cli.py
tconbeer: Agreed on the implementation, and it turned out not to need anything new on `AbstractOption`.

The ugly part was building a throwaway function just to harvest `__click_params__` off it. But `to_click()` already appends straight to a `Command` — that's what `harlequin.cli` does — so the options can go on the real command and the collision fixup becomes a slice of `cmd.params`:

```python
def _attach_adapter_options(cmd: click.Command, adapter_cls) -> None:
    reserved = {opt for param in cmd.params for opt in param.opts}
    taken = {param.name for param in cmd.params}
    first = len(cmd.params)

    for option in adapter_cls.ADAPTER_OPTIONS or []:
        option.to_click()(cmd)

    for param in cmd.params[first:]:
        param.opts = [opt for opt in param.opts if opt not in reserved]
        param.secondary_opts = [...]
    cmd.params[first:] = [p for p in cmd.params[first:] if p.opts and p.name not in taken]
```

No stub, no `reversed()`, no `__click_params__`, and `_adapter_params` is gone.

I did try the version you suggested first — a `to_click_parameter()` on `AbstractOption` returning a `click.Option` directly. The problem is that doing it *properly* means making the decls/attrs the abstract member instead of `to_click()`, and `to_click()` is currently the abstract method every out-of-tree option class implements. Any adapter that subclasses `AbstractOption` rather than using the five built-ins would stop being instantiable. Doing it *improperly* means the same stub-harvesting trick, just relocated. Since the call site gets clean without either, I'd rather not spend public API on it — but say the word if you want the method for other reasons and I'll add it the non-breaking way.

Now covered by a test that pins all three behaviors — colliding short dropped, whole option dropped when hsql owns its only spelling, non-colliding untouched.

---

tests/unit_tests/test_hsql.py
tconbeer: Agreed — there were three approaches, and now there's one: `no_discovered_config` in `tests/conftest.py`. `mock_config_loader` (functional conftest) and `mock_empty_config` (test_cli) keep their names and become thin wrappers over it, so the 12 existing call sites and `test_keys_app`'s override of the first are untouched.

I unified on *this* module's seam rather than test_cli's, and it's worth saying why, since it's the opposite direction from what you might have expected:

- **`harlequin.cli.get_config_for_profile`** (what test_cli and the functional conftest patched) can't unify by construction — it patches one module's imported name, so it never reaches `harlequin.hsql.cli`, which imports the same function into its own namespace. Any second command needs a second patch.
- **`harlequin.config.load_config`** does reach both, but it breaks `--config-path`: stub it out and an explicitly passed config file is ignored too, which kills `test_the_default_profile_applies` and the other profile tests here.
- **The three `_search_*` functions** are the only seam that both commands go through *and* that leaves an explicit `--config-path` working. It's also the most faithful statement of what the tests actually want — "pretend this machine has no config files" — rather than "pretend config loading returns nothing."

Side benefit: `test_config_path` and `test_config_path_fron_env` in test_cli were relying on real discovery being harmless, so they'd have picked up a developer's `~/.harlequin.toml`. They're hermetic now.

`mock_load_config` stays as-is — supplying a specific config is a different job from removing discovery, and it's the right tool for that one.

---

tests/unit_tests/test_hsql.py
tconbeer: Fair, and the diagnosis is the useful part: those three weren't testing the CLI. Gone, along with `HSQL_MAIN` and `run_hsql` — test_hsql.py is now nothing but the click runner. Each property went where it belongs:

- **`test_output_is_lf_on_every_platform`** → already covered. `test_golden_formats.py` snapshots the bytes of every format, and `.gitattributes` pins `tests/**/__snapshots__/**` to `eol=lf` so a Windows checkout can't quietly rewrite what they assert. The subprocess test was a weaker restatement of that.
- **`test_output_does_not_vary_with_the_locale`** → an import-linter contract. What actually has to be true is "hsql never calls `set_locale()`", so `harlequin.locale_manager` joins the forbidden list for `harlequin.hsql`. That fails identically on every machine, where a subprocess with `LC_ALL=de_DE.UTF-8` set was circumstantial and silently vacuous on any runner without that locale generated — which is most CI images.
- **`test_output_does_not_vary_with_a_pipe`** → `test_output_file_and_redirect_agree` covers it in-process, and both go through the same `click.get_binary_stream("stdout")` anyway.

The one surviving subprocess test is genuinely about imports — that building `hsql --help` loads no adapter — so it moved to `test_import_hygiene.py` and uses that module's existing `run_python` fixture rather than a second way of spawning an interpreter.

---

tconbeer:
All eleven addressed, in `6b8145e` (source) and `c2d5461` (tests). Replied inline on the five with a decision in them; the other six were done as asked:

- `DEFAULT_ADAPTER` and `TUI_ONLY_KEYS` → `harlequin.config`, next to the `Profile` schema they describe. `harlequin.cli` imports the first rather than declaring its own copy.
- The profile is read **once**: `_preflight` returns it alongside the adapter name and any error it hit, and the callback reports that error where there's an exit code to report it with. Pinned by a test that counts `load_config` calls.
- `_preflight` takes the installed adapter names instead of asking for them a second time.
- `_Run` owns its clock and derives `elapsed_ms`, `status` and `exit_code`.
- `os.getenv`.

`make check` is green: ruff, mypy, `lint-imports` 4/4, 559 passed / 6 skipped / 1 xfailed on 3.10, plus the py12 run.

Two things worth your eye, both explained inline: I did **not** add `-h` for help (postgres and mysql both declare it for `--host`, and hsql's flags win collisions, so claiming it would strip theirs), and I did **not** add a method to `AbstractOption` — attaching the adapter's options to the real command instead of a stub function made the call site clean without spending public API.

---

PR #1031, opened 2026-08-11
feat: give each limit one meaning, and one name (#1026)

### LINKED ISSUE #1026: Give the Results Viewer's row cap its own config key, and deprecate `limit` for the TUI
`limit` now names two different things, and inside the TUI it quietly drives three.

**Across the two commands.** In the IDE, `limit` is a *soft display cap over a full fetch* — every row leaves the database and the Results Viewer holds the first N, which is why it can say `(Showing 100,000 of 3,412,887 Records)` ([`results_viewer.py:231`](https://github.com/tconbeer/harlequin/blob/main/src/harlequin/components/results_viewer.py#L231)). In `hsql` it is the *hard fetch limit* — `cursor.set_limit()`, so fewer rows leave the database at all — because a viewport doesn't exist headless and fetching a million rows to print five hundred defeats the point. Same word, different promise, and the two commands share one config file.

**Inside the TUI.** The single key drives all of:

1. the Results Viewer's row cap ([`cli.py:346`](https://github.com/tconbeer/harlequin/blob/main/src/harlequin/cli.py#L346) → `max_results`),
2. the Run Query Bar's limit input *initial value*, as `min(500, max_results)` ([`run_query_bar.py:41`](https://github.com/tconbeer/harlequin/blob/main/src/harlequin/components/run_query_bar.py#L41)),
3. the *maximum* that input will accept ([`run_query_bar.py:45`](https://github.com/tconbeer/harlequin/blob/main/src/harlequin/components/run_query_bar.py#L45)).

(2) and (3) are the hard fetch limit. So the TUI already has both concepts, and one key sets a display cap and bounds a fetch cap at the same time.

### The bug it causes today

A profile written for the IDE is a profile `hsql` reads:

```toml
[profiles.dev]
adapter = "duckdb"
limit = 100_000   # meant: "the viewer holds 100k rows"
```

`hsql -P dev -c "select * from big_table"` reads that as a hard limit and pulls 100,000 rows over the wire into an agent's context, where the command's own default is 500. Nothing is wrong, nothing warns, and the user never chose it. Raised while reviewing #1025; the changelog there documents the behavior rather than fixing it, because the fix is this issue.

### Proposal

Add a key that means only the display cap, and let `limit` converge on the hard fetch limit that both commands honor identically.

```toml
[profiles.dev]
limit = 500                 # rows fetched from the database
viewer_max_rows = 100_000   # rows loaded into the Results Viewer
```

**On the name.** `viewer_max_rows` (with `harlequin --viewer-max-rows`), because it names the widget the user can actually see — the Results Viewer is already a named, documented part of the UI — and a *viewer* cannot be mistaken for something that changes what the database sends.

Runners-up and why not:

- **`max_results`** is tempting, since it is already the internal parameter name from `cli.py` all the way to both widgets, so config and code would finally agree. But "results" is precisely the overloaded word: `hsql --result all|last|N` selects *result sets*, and `max_results` reads as a cap on those. Disambiguating vocabulary is the whole point here.
- **`display_limit`** pairs nicely against `limit` (display vs fetch) and is short enough to type as a flag. It just doesn't say how many *what*, and could be read as truncating wide cells.
- **`max_rows`** is too close to `limit` to teach anyone the difference, which is the job.

### Migration

1. **Add `viewer_max_rows`**, same semantics and same 100,000 default as `limit` has in the TUI today. When both are set, it wins. `harlequin --config` writes the new key ([`config_wizard.py:88-92,159`](https://github.com/tconbeer/harlequin/blob/main/src/harlequin/config_wizard.py#L88) still writes `limit`).
2. **Deprecate `limit` for the TUI.** It keeps working, and a profile that sets it without `viewer_max_rows` gets a deprecation notice on stderr — the same channel `pretty_print_warning` already uses, so it can't contaminate piped output.
3. **In 3.0, `limit` becomes the hard fetch limit in both commands.** In the IDE, `--limit` and the `limit` key will **set the Run Query Bar's limit input and check the box**, so the query is actually limited. That is what makes the two commands comparable: one key, one meaning, `cursor.set_limit()` at the end of it in both.

Today the box starts *unchecked* with an inert `500` in the input — pinned by [`test_run_query_bar.py:28-30`](https://github.com/tconbeer/harlequin/blob/main/tests/functional_tests/test_run_query_bar.py#L28) (`limit_checkbox.value is False`, `limit_input.value == "500"`, `limit_value is None`) — so the IDE fetches everything by default. Step 3 is therefore a real behavior change for anyone who sets `limit`: they go from a full fetch with a display cap to a hard fetch. Nearly the same thing on screen, but it changes what leaves the database, which is why it waits for a major version.

(Correcting something in the first draft of this issue: config *does* already reach that input, as `min(500, max_results)` — so `limit = 100` shows `100` there today. What it can't do is set it above 500, or turn it on.)

### This makes overflow detection in the TUI a prerequisite

Once the box is checked by default, the Results Viewer stops being able to report an exact total — and it will not say so.

That is already true when a user checks the box by hand today. `_execute_query` applies the bar's limit as `RowLimit(max_rows=message.limit, detect_overflow=False)` ([`app.py:1083`](https://github.com/tconbeer/harlequin/blob/main/src/harlequin/app.py#L1083)), so the database returns exactly 500 rows; `_fetch_data` then caps at `max_results` ([`app.py:1154`](https://github.com/tconbeer/harlequin/blob/main/src/harlequin/app.py#L1154)), which 500 never reaches; and `_human_row_count` falls through to `(500 Records)` for what may have been three million. A silent truncation, in the one place the plan claimed the TUI never needed to guard against one.

The M1 technical plan settled `detect_overflow=False` for the TUI on the grounds that "`--limit` there is a soft display cap over a full fetch, so it already reports the exact total." Step 3 retires that premise. So this migration needs the TUI on `detect_overflow=True` whenever the bar's limit is on, and `_human_row_count` needs a `(500 of 500+ Records)` form — the same thing the headless layouts already print, for the same reason. Worth doing ahead of step 3 on its own merits, since it's a live bug for anyone using the checkbox.

### Open questions

- **When `limit` is unset, does the box stay unchecked?** Leaving it unchecked keeps today's IDE feel (a human watching a viewport can afford a full fetch), at the cost of the two commands still having different *defaults* — 500 hard in `hsql`, unlimited in the IDE — even once the key means one thing. Defaulting to 500 checked would be full parity but would quietly change what every existing user's queries return. I'd keep it unchecked and let the defaults differ; the promise worth making is that the key means the same thing, not that the numbers match.
- Does `harlequin --limit` get deprecated alongside the key, or only the key? They share a name and a merge path, so splitting them is its own confusion — but `--viewer-max-rows` is a lot to type interactively.
- Once decoupled, what bounds the Run Query Bar's input? Today its maximum is `max_results`, which only makes sense while the two are the same number. Probably nothing should bound it.
- Is `viewer_max_rows = 0` still "no limit", as `limit = 0` is today? Worth keeping, for symmetry.

tconbeer:
Closes #1026

There are two kinds of limit, and after this every option is exactly one of them:

- **Hard, fetch** — `limit` / `--limit`, in *both* commands. `cursor.set_limit()`, so fewer rows leave the database. In the IDE it fills in the Run Query Bar's limit and checks the box, so the query is actually limited and the number in the bar is the one `set_limit()` receives.
- **Soft, display** — applied over rows already fetched, so the total stays known exactly. `viewer_max_rows` / `--viewer-max-rows` for the Results Viewer (the 100,000 default `limit` used to carry), and `--display-rows` for `hsql`'s text layouts.

Everything else follows from that:

- **`-1` is "no limit" everywhere.** `0` now means zero rows, so `limit 0` can ask a database what a query's columns are — the idiom the old `0`-means-unlimited spelling took away. The one exception is `viewer_max_rows`, where `0` has always meant "hold everything" and a viewer holding no rows would serve nobody.
- **`--display-rows` caps what `hsql` prints, of what `--limit` fetched**, defaulting per layout: 40 rows for `table` and `markdown`, 10 records for `vertical`, where a record is a line per column. The footer holds all three numbers that can now differ — `(40 of 100 rows)`, `(500 of >500 rows)`, `(40 of >50 rows)`. File formats are untouched: dropping rows out of a csv is a different promise from not filling a screen with them, and a `--display-rows` a file format can't honor says so on stderr rather than looking like a cap that was applied.
- **`-l` is gone, in both commands.** It's psql's "list databases", and a flag that lists databases in one client and truncates results in another is a mistake waiting for a script.
- **The TUI detects overflow**, which #1026 called a prerequisite and which was a live bug for anyone who checked the box by hand: a query over three million rows with the limit checked returned 500 and the viewer said `(500 Records)`. It now fetches one row more than the limit — the only way to tell 500 rows from 500-and-more — and says `(Showing 500 of >500 Records)`.
- **A SQLite bug that had no caller until now**: the cursor fetches the first row eagerly and asked for the rest with `fetchmany(limit - 1)`, but `sqlite3` reads `fetchmany(0)` as "all of them", so `set_limit(1)` returned the whole table.

The commits are meant to be read in order; each one is a state that could ship.

**Step 3 now, rather than in 3.0.** The issue deferred the semantic flip of `limit` on the grounds that it changes what leaves the database. Worst case it turns a soft cap into a hard limit — nearly the same thing on screen — so the two-release migration bought less than the confusion of a key that means one thing this month and another thing next. That decision is the one to push back on if you disagree; everything else stands without it.

**No deprecation notice for `limit`.** With step 3 in the same change there is nothing to deprecate, and the migration is one word (`limit` → `viewer_max_rows`, same number). A start-up warning on stderr would also be nearly invisible under the alt-screen, where the checked box in the Run Query Bar showing the number is the notice that lands. Dropping `-l` in the same release is the other half of that: a script that passes it now fails loudly rather than quietly fetching fewer rows than it used to.

**`fetched_row_count` on `ResultSet`, rather than inferring from the backend.** `backend.source_row_count` counts the overflow probe row, so "500 of 501" would be an artifact rather than a fact. Carrying the probe-excluded count means the viewer, the layout footers, query history and the Data Exporter all read the same three numbers, and the exporter writes 500 rows under a limit of 500 rather than 501.

**The soft cap on layouts is applied in `_rows()`, not per layout**, so column widths are measured over what is printed — a value in a row nobody sees doesn't pad the rows they do.

**Two open questions from the issue, answered the way the issue preferred:** an unset `limit` leaves the box unchecked, so the IDE still fetches everything by default and the two commands still have different *defaults* (500 hard in `hsql`, unlimited in the IDE) even though the key now has one meaning; and the Run Query Bar's input is no longer bounded by anything, since the viewer's cap only ever bounded it because one option set both.

- [x] Yes; I haven't opened a PR, but the gist of the change is: the config/CLI reference needs `viewer_max_rows` / `--viewer-max-rows` as the Results Viewer's cap; `limit` / `--limit` documented as the hard fetch limit in both commands, with `-1` for no limit and `0` for a header and no rows; `--display-rows` documented on the headless page, with its per-layout defaults and the note that it doesn't reach file formats; and every `-l` occurrence respelled `--limit`, since the short flag no longer exists. The "Headless & Agents" page's `-l 0` examples need to become `--limit -1`.

- [x] Yes.


src/harlequin/hsql/cli.py
```diff
@@ -44,19 +46,23 @@
...
-`hsql -l` is also the *hard* limit -- `cursor.set_limit()`, so fewer rows leave
-the database -- where the IDE's is a soft cap over a full fetch. Same spelling,
-different promise, and the docs say so.
+`--limit` is the *hard* limit -- `cursor.set_limit()`, so fewer rows leave the
+database -- and it is the same promise the `limit` key makes in the IDE. `-1`
+is unlimited, and `0` fetches a header and no rows, which is how a caller asks
+what a query's columns are.
+
+No `-l`: that is psql's `--list`, and a flag that lists databases in one
+command and truncates results in the other is a mistake waiting for a script.
```
tconbeer: remove this part of the comment.

src/harlequin/components/run_query_bar.py
```diff
@@ -9,6 +9,15 @@
 from textual.widget import Widget
 from textual.widgets import Button, Checkbox, Input
 
+DEFAULT_LIMIT = 500
+"""What the limit input offers, unchecked, when nothing configured a limit.
+
+A number to start from rather than a limit in force: unchecked, it does
+nothing until someone checks the box. Nothing bounds it -- this is the hard
+fetch limit, and the Results Viewer's cap is a different number about a
+different thing, which once bounded this input back when one option set both.
```
tconbeer: shorten by removing these lines.

src/harlequin/components/run_query_bar.py
```diff
@@ -18,10 +27,16 @@ def __init__(
...
-        max_results: int = 10_000,
+        query_limit: int | None = None,
         show_cancel_button: bool = False,
     ) -> None:
-        self.max_results = max_results
+        self.query_limit = query_limit
+        """The limit `--limit` configured, in force from the first query.
+
+        None leaves the box unchecked, which is a full fetch: a human watching
+        a viewport can afford one, and it is what the IDE has always done.
```
tconbeer:
```suggestion
        None leaves the box unchecked, which is a full fetch.
```

src/harlequin/hsql/cli.py
```diff
@@ -182,13 +188,21 @@ def build_cli(argv: Sequence[str]) -> click.Command:
...
-        help="Maximum rows per result set. 0 for no limit.",
+        type=click.IntRange(min=UNLIMITED),
+        help="Maximum rows fetched per result set. -1 for no limit.",
+    )
+    @click.option(
+        "--display-rows",
+        metavar="N",
+        type=click.IntRange(min=UNLIMITED),
+        help=(
+            "Rows printed per result set by the text layouts. -1 for all of them. "
```
tconbeer:
```suggestion
            "Rows printed per result set by the text layouts. -1 for all rows. "
```

src/harlequin/cli.py
```diff
@@ -276,13 +304,22 @@ def build_cli() -> click.Command:
...
-        "-l",
-        default=DEFAULT_LIMIT,
-        type=click.IntRange(min=0),
+        "--viewer-max-rows",
+        default=DEFAULT_VIEWER_MAX_ROWS,
+        type=click.IntRange(min=-1),
         help=(
             "Set the maximum number of rows that can be loaded into Harlequin's "
-            "Results Viewer. Set to 0 for no limit. Default is 100,000"
+            "Results Viewer. Set to -1 for no limit. Default is 100,000"
```
tconbeer:
```suggestion
            f"Results Viewer. Set to -1 for no limit. Default is {DEFAULT_VIEWER_MAX_ROWS}"
```

src/harlequin/cli.py
```diff
@@ -276,13 +304,22 @@ def build_cli() -> click.Command:
...
+        ),
+    )
+    @click.option(
+        "--limit",
+        type=click.IntRange(min=-1),
+        help=(
+            "Limit every query to this many rows, as if you had typed it into "
+            "the Run Query Bar and checked the box: fewer rows leave the "
+            "database. The same limit [bold]hsql --limit[/] applies. Unset, queries "
+            "fetch every row."
```
tconbeer: change to: "Default value for the limit control; if set, the limit will be applied by default. If unset, queries fetch all rows"

CHANGELOG.md
```diff
@@ -12,13 +12,32 @@ All notable changes to this project will be documented in this file.
...
+
+### Breaking Changes
+
+- `limit` and `--limit` now limit what the IDE **fetches**, where they used to cap what the Results Viewer **displayed** ([#1026](https://github.com/tconbeer/harlequin/issues/1026)). A profile with `limit = 100_000` used to fetch every row and show the first 100,000; it now fetches 100,000 and says `(Showing 100,000 of >100,000 Records)`. On screen the two are nearly the same thing — what changes is what leaves the database, which is the point. To keep the old meaning, rename the key to `viewer_max_rows`; the number does not change. Nothing warns at start-up because nothing needs to: the limit is visible in the Run Query Bar, checked, with the number in it.
+- `harlequin -l` is gone; the option is `--limit`, spelled out ([#1026](https://github.com/tconbeer/harlequin/issues/1026)). `-l` is psql's "list databases", and the two meanings are too far apart to leave sitting on the same letter. Losing the shorthand in the same release that changes what the option does is deliberate: a script that passes `-l` now fails loudly instead of quietly fetching fewer rows than it used to.
```
tconbeer: make this max 2 bullets, much shorter.

src/harlequin/cli.py
tconbeer: Applied, with `:,` added — `{DEFAULT_VIEWER_MAX_ROWS}` renders `100000`, and the help text read `100,000` before. Say the word if you'd rather have the bare digits.

---

PR #1032, opened 2026-08-11
perf: take three fixed costs off every hsql invocation

tconbeer:
No existing issue — this came out of profiling `hsql` cold start. Related to #524.

Profiling `hsql -c "select 1"` put ~90% of a 274ms cold start in imports. Most of that is duckdb and pyarrow, which the query genuinely needs — a bare `python -c "import click, duckdb, textual_fastdatatable.backend, pyarrow.dataset"` is 238ms on the same machine, so harlequin's own code was only ~45ms on top. Three costs on that path weren't buying anything:

1. **Config is now read with `tomllib` and only written with tomlkit.** Start-up reads config, and `_search_cwd` picks up whatever `pyproject.toml` is in the working directory — so running `hsql` from any Python project paid tomlkit's style-preserving parser to read a file that usually has no `[tool.harlequin]` section at all. On this repo's own 10KB `pyproject.toml` that parse is 46ms against tomllib's 1.5ms, plus tomlkit's own ~12ms import.
2. **`harlequin.layout` defers `wcwidth`** to the first non-ASCII cell it measures. `_width()` already had an `isascii()` fast path that returns without calling `wcswidth`, but the module-scope import was paid by every run regardless (~25ms).
3. **`hsql --version` takes the early exit `--help` already had.** A version string doesn't depend on an adapter, so it shouldn't wait for one to import.

Measured warm, interleaved against `main`, `select 1`:

| | main | this PR |
|---|---|---|
| from a dir with a `pyproject.toml` | 328ms | **258ms** |
| from a dir without one | 274ms | 264ms |
| `hsql --version` | 166ms | **93ms** |

The tomlkit split is the one with a real design question, because writes have a requirement reads don't: a user's comments and key order have to survive `harlequin --config`. So rather than dropping tomlkit, `ConfigFile` now re-reads the file through it on the first `update()`, and nothing imports tomlkit until something writes. `harlequin --config` and the keymap editor produce byte-identical output to `main` — I diffed it. `tomli` is added as a dependency on 3.10 only; it's `tomllib` before it was stdlib.

For `--version` I deliberately mirrored the existing `--help` rule rather than short-circuiting unconditionally: `hsql --version -a postgres` still loads postgres, so an invocation carrying that adapter's options still parses.

Two things I profiled and decided against:

- **Deferring `pyarrow.compute`/`types`/`lib` in `textual-fastdatatable`.** I prototyped it upstream; it saved ~2ms. `pyarrow.dataset`, pulled in later by `duckdb.from_arrow()` inside `_cast_to_text`, imports `pyarrow.compute` anyway.
- **Escaping pyarrow entirely on the text-layout path** — the remaining ~120ms. That means a non-Arrow path for small result sets, which is exactly the second normalizer the `harlequin.query` and `text_columns()` docstrings argue against, and it would put `-F table` and `-F csv` at risk of disagreeing about a timestamp. Not worth it for a fixed 120ms.

One unrelated fix rode along, found while writing the tests: a config file that sets `default_profile` and defines no `profiles` table raised an unhandled `KeyError` instead of being reported as a config error.

- [x] No.

- [x] Yes.

The write path this touches turned out to have **no test coverage at all** — a first draft of this change broke `ConfigFile.write()` outright and all 499 tests stayed green. So this adds round-trip tests for it (comment preservation, the pyproject case, creating a file that doesn't exist yet), plus run-time guards in `test_import_hygiene.py` for each deferral, since the static graph can't see one. I mutation-tested all four guards: each fails when its deferral is reverted. The non-ASCII test also asserts the column is still padded to the width the ideographs actually occupy, so a deferral that never fires would show up as the misalignment bug it would be rather than as a passing test.

Worth flagging one limit these tests document rather than fix: comments *within* a rewritten table still don't survive a write, because `relevant_config` hands back plain data and `update()` therefore replaces a whole table rather than editing inside it. That's pre-existing — `main` behaves identically, byte for byte — and out of scope here.

509 unit tests, 103 functional tests, ruff, mypy and all four import-linter contracts pass.

- [x] I have added an entry to `CHANGELOG.md`, under the `[Unreleased]` section heading. There's no issue for this one, so the entries don't reference one.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QBTrywAgMM9ce4dSrQXKG5)_


CHANGELOG.md
```diff
@@ -36,13 +36,21 @@ All notable changes to this project will be documented in this file.
...
+- A config file that sets `default_profile` and defines no profiles at all is now reported as the config error it is, instead of escaping as an unhandled `KeyError`.
 
 ### Performance
 
 - `harlequin.exception` no longer imports `rich` ([#524](https://github.com/tconbeer/harlequin/issues/524)). It is on every headless path — `harlequin.config` imports it — so `hsql --help` was paying 20ms for a panel it never draws.
+- Config files are now **read** with `tomllib` and only **written** with tomlkit. Start-up reads config, and `_search_cwd` picks up whatever `pyproject.toml` is in the working directory — so running `hsql` from any Python project paid tomlkit's style-preserving parser to read a file that usually has no `[tool.harlequin]` section at all. On this repo's own 10KB `pyproject.toml` that parse is 46ms against tomllib's 1.5ms. Writes still go through tomlkit, because a user's comments have to survive `harlequin --config`; nothing imports it until something writes.
```
tconbeer: Collapse this to one line, reference the PR

tconbeer: Done in 2f8c12d — the three bullets and the timing paragraph are now one line referencing this PR, and it says what a user gets rather than which parser reads what:

> - `hsql` starts faster ([#1032](https://github.com/tconbeer/harlequin/pull/1032)): `hsql -c "select 1"` goes from 328ms to 258ms run from a directory with a `pyproject.toml` in it, and from 274ms to 264ms without one; `hsql --version` goes from 166ms to 93ms.

Also dropped the internals from the two other entries I'd added and pointed both at the PR, since neither has an issue: the bug fix now just says the config error is reported "instead of crashing", and the dependency line is down to "Adds `tomli` on Python 3.10 only". Left the `harlequin.exception`/`rich` entry above alone, since it isn't mine.

---

PR #1038, opened 2026-08-13
Plan: M2 technical plan for hsql's verbs and safety flags (#524)

tconbeer:
Implementation plan for M2 -- self-description and safety -- modeled on the
M1 plan, and written against measurements taken on this checkout rather than
against estimates.

What it settles:

- Catalog listings are result sets, so `--format`, `-o`, `-t/-A`,
  `--display-rows` and `--stats` already apply to them; M2 adds two layouts
  (`compact`, `tree`) and no output subsystem.
- Five additive fields on the adapter contract, because four headline
  features are unimplementable without them: `CatalogItem.type_name` and
  `child_count`, `IMPLEMENTS_READ_ONLY`, `IMPLEMENTS_VALIDATE_SQL`, and
  `AbstractOption.secret`.
- Capabilities are declared, not detected -- reflection reports SQLite as
  validating SQL, since it defines `validate_sql` and raises from it.
- Three releases: catalog/describe, then self-description, then the safety
  flags, which are the only group that needs an ecosystem rollout.

Measurements behind it: a listing costs one round trip per path segment plus
one (not the "exactly one fetch_children" the product plan assumes), and
`--depth 2` over a 400-relation schema is 403 round trips and 2.3s against a
local DuckDB file; a cancelled query returns zero rows with no error and
would exit 0, so a timeout has to attribute its own cancellation, and exiting
while a worker sits in DuckDB aborts with 134 rather than the documented 4;
DuckDB's `validate_sql` treats a missing table as valid; config files merge
per top-level key, so a cwd file with any profile hides every profile in the
home file.

Also applies this plan's corrections to the product plan, as amendments in
the style of M1's.

### DIFF
--- a/docs/plans/harlequin-for-agents.md
+++ b/docs/plans/harlequin-for-agents.md
@@ -151,6 +151,14 @@ Everything in Workstream B lands as `hsql` subcommands: `hsql catalog`,
 `hsql mcp`, `hsql open`. Bare `hsql` with no arguments prints help; it never launches
 the TUI.
 
+> **Amended in M2's plan.** Subcommands need the collision rule this section raises
+> against them, because `hsql` takes `CONN_STR` positionally: a database file named
+> `catalog` is ambiguous with the verb. The rule is that the first non-option argument is
+> a verb *if it exactly matches one*, and `./catalog` is the escape. The set is seven
+> short words and the fallback is a command with required flags rather than a
+> full-screen app, which is what makes the ambiguity tolerable here and not on
+> `harlequin`.
+
 ### Consequences
 
 - **The `-f` collision disappears, and with it the plan's only breaking change.**
@@ -301,6 +309,21 @@ identically whether stdout is CSV or Parquet.
 **Errors are plain.** No Rich panels, no ANSI, no box drawing:
 `hsql: error: relation "usres" does not exist` on stderr. The panel is a TUI affordance.
 
+> **Amended in M2's plan**, on two of the safety flags above.
+>
+> **`--timeout` has to attribute the cancellation itself.** A cancelled DuckDB query comes
+> back as an empty result set with no error — `fetchall()` swallows the interrupt and
+> returns `None`, which is also what an empty result looks like — so a timeout that trusts
+> the adapter prints "no rows" and exits 0. `hsql` knows it fired the deadline, discards
+> that result, and exits 4. It also refuses up front on an adapter that can't cancel,
+> since there is otherwise no way to stop the work.
+>
+> **`--dry-run` is a parse check, not an existence check**, on the adapter it will be used
+> with most. DuckDB's `validate_sql()` treats every non-parser error as valid, so a query
+> against a table that doesn't exist dry-runs clean. M2 adds a `PREPARE` pass to the
+> in-tree adapter for statements that parse and aren't DDL, which catches it; other
+> adapters keep whatever they have, and the docs say per adapter what is checked.
+
 **Exit codes:**
 
 | Code | Meaning |
@@ -428,6 +451,27 @@ hsql find <TERM> [--in tables|columns|all]
 - **`describe`** returns one object in full — columns, types, comments, row counts where
   the adapter can supply them. One round trip.
 
+> **Amended in M2's plan**, on three things this section assumes the catalog contract can
+> already do.
+>
+> **A listing is one round trip per path segment, plus one** — four for
+> `catalog mydb.analytics.orders`, not one. An item can only be reached through its
+> ancestors, so the resolve in front of the listing is itself a walk. The listing stays
+> single-level and cheap; the resolve is what a network database charges for.
+>
+> **`--format compact` needs a field that doesn't exist.** `CatalogItem.type_label` is a
+> 1–3 character label for a tree column (`##`, `s`, `ts`), so today's honest rendering is
+> `total #.#`, not `total DECIMAL(18,2)`. The full type is fetched and discarded by the
+> in-tree adapters already; M2 adds an optional `type_name` to keep it, and falls back to
+> the short label for adapters that never populate it.
+>
+> **Child counts need one too** — an optional `child_count`, where `None` means unknown
+> rather than zero, so an agent can't read a missing count as an empty schema.
+>
+> Measured while writing that plan: `--depth 2` on a schema with 400 relations is 403
+> round trips and 2.3 seconds against a *local* DuckDB file. The node budget and the child
+> counts are not garnish.
+
 ### `hsql find` — the real gap
 
 What an agent most often needs isn't "list this schema," it's *"where does `orders`
@@ -486,6 +530,13 @@ redacted.
 Capability flags matter: they let an agent know *not* to try `--dry-run` on an adapter
 that can't validate, instead of trying and failing.
 
+> **Amended in M2's plan.** The flags have to be *declared* on the adapter class, not
+> detected on the connection. Reflection is measurably wrong: `HarlequinSqliteConnection`
+> defines `validate_sql` and raises `NotImplementedError` from it, so an override check
+> reports SQLite as validating SQL. Declaring them is also what lets `hsql info --json`
+> answer while the database is unreachable, which is when a diagnostic is worth most — so
+> `info` imports adapters but opens no connection.
+
 - **B5.** As an agent debugging a human's setup, I run `hsql info --json` and can tell
   them exactly which config file is winning and why.
 - **B6.** As an agent, I check capability flags before offering the human a feature their
@@ -540,6 +591,22 @@ and a wrong guess produces a confusing runtime error. The config is shared betwe
    agents and CI ever touching a config that involves secrets.
 6. **A formal spec page on the site**: discovery order, merge semantics, precedence,
    every key with its type and default, and the reserved/invalid names.
+
+> **Amended in M2's plan**, on two things this workstream would otherwise document as they
+> stand.
+>
+> **The merge is per top-level key, not per profile.** `_merge_config_files()` is a
+> `dict.update()` per file, so a cwd config file that defines any profile hides *every*
+> profile in the home file. Nobody writes a project-local profile in order to lose their
+> personal ones; M2 makes the merge per-key and treats the change as a bug fix, and the
+> spec page should describe that rather than what exists today.
+>
+> **`${VAR}` interpolation has to be a read-path transform.** `harlequin --config` reads
+> the config and writes it back, so resolving values too early bakes a plaintext password
+> into the user's file on their next wizard run. The syntax is `${VAR}` and
+> `${VAR:-default}` — a bare `{` never triggers it, so nothing anyone has in a password
+> today needs escaping — and an unset variable with no default is a config error naming
+> the variable and its file, never an empty string.
 
 ### Secrets: #667, #898, and redaction are one feature
 
@@ -804,7 +871,15 @@ Because the CLI is a separate command, **every milestone here is purely additive
 [the M1 technical plan](./m1-hsql-technical-plan.md): the module architecture, the
 refactors that have to land first, and an eight-PR sequence across two releases. A few
 figures and defaults in this document were corrected against measurements taken while
-writing it; §8 there lists them.
+writing it; §8 there lists them. PRs 1–6 shipped and PR 7, the docs topic, is in flight;
+PR 8, the agent eval suite, was not built.
+
+**On M2.** Likewise in [the M2 technical plan](./m2-hsql-technical-plan.md), across three
+releases rather than one: the catalog verbs first, since they need nothing outside this
+repo; then self-description; then the safety flags, which need additive fields on the
+adapter contract and therefore an ecosystem rollout. The amendments marked above are the
+corrections that came out of measuring it. `--single-transaction` belongs to M2 as well —
+M1 deferred it here, though this table's M2 row omits it.
 
 **On M0.** Name availability is the only thing in this plan that someone else can take
 while we deliberate, which is why it's a milestone rather than a task. It's also cheap
--- a/docs/plans/harlequin-for-agents.md
+++ b/docs/plans/harlequin-for-agents.md
@@ -471,6 +471,11 @@ hsql find <TERM> [--in tables|columns|all]
 > so "report child counts wherever the adapter can get them cheaply" cannot be implemented
 > as written. What a listing can report is what it fetched, which it already does.
 >
+> **`describe` is `catalog` on a relation, so there is no separate verb.** One level below
+> a relation *is* its columns, with their names, real types and quoted identifiers. All a
+> second mode could add is detail the contract does not carry (comments) or that costs a
+> table scan (row counts).
+>
 > **`--format compact` needs a field that doesn't exist.** `CatalogItem.type_label` is a
 > 1–3 character label for a tree column (`##`, `s`, `ts`), so today's honest rendering is
 > `total #.#`, not `total DECIMAL(18,2)`. The full type is fetched and discarded by the
@@ -892,14 +897,14 @@ writing it; §8 there lists them. PRs 1–6 shipped and PR 7, the docs topic, is
 PR 8, the agent eval suite, was not built.
 
 **On M2.** Likewise in [the M2 technical plan](./m2-hsql-technical-plan.md), across three
-releases rather than one: the catalog modes first, since they need nothing outside this
-repo; then self-description; then the safety flags, which need additive fields on the
-adapter contract and therefore an ecosystem rollout. The amendments marked above are the
-corrections that came out of measuring it. Two additions and several subtractions to this
-table's M2 row: `--single-transaction` belongs to M2 (M1 deferred it here), the
-introspection surface is mode options rather than subcommands, and `fmt`, `--dry-run`,
-recursive `--depth`, the node budget and child counts are all cut — each for a reason
-given in §7 of that plan.
+releases rather than one: the catalog first, since it needs nothing outside this repo; then
+self-description; then the safety flags, which need additive fields on the adapter contract
+and therefore an ecosystem rollout. The amendments marked above are the corrections that
+came out of measuring it. One addition and several subtractions to this table's M2 row:
+`--single-transaction` belongs to M2 (M1 deferred it here), the introspection surface is
+mode options rather than subcommands, and `describe`, `fmt`, `--dry-run`, recursive
+`--depth`, the node budget and child counts are all cut — each for a reason given in §7 of
+that plan.
 
 **On M0.** Name availability is the only thing in this plan that someone else can take
 while we deliberate, which is why it's a milestone rather than a task. It's also cheap
--- a/docs/plans/harlequin-for-agents.md
+++ b/docs/plans/harlequin-for-agents.md
@@ -317,6 +317,13 @@ identically whether stdout is CSV or Parquet.
 > that result, and exits 4. It also refuses up front on an adapter that can't cancel,
 > since there is otherwise no way to stop the work.
 >
+> **`--single-transaction` is cut too.** Adapters name their transaction modes freely —
+> SQLite and Postgres offer "Auto"/"Manual", DuckDB overrides nothing and reports none, and
+> a third-party adapter may name its modes anything and default to whatever its driver
+> does — so core cannot ask for "the manual one" through that interface without first
+> reworking what a transaction mode is. Story A6's guarantee comes from `begin` and
+> `commit` in the caller's own script until then.
+>
 > **`--dry-run` is cut**, from M2 and possibly for good. Only DuckDB implements
 > `validate_sql` at all, and there it is a parse check rather than an existence check:
 > `select * from no_such_table` validates clean, because the adapter treats every
@@ -476,11 +483,14 @@ hsql find <TERM> [--in tables|columns|all]
 > second mode could add is detail the contract does not carry (comments) or that costs a
 > table scan (row counts).
 >
-> **`--format compact` needs a field that doesn't exist.** `CatalogItem.type_label` is a
-> 1–3 character label for a tree column (`##`, `s`, `ts`), so today's honest rendering is
-> `total #.#`, not `total DECIMAL(18,2)`. The full type is fetched and discarded by the
-> in-tree adapters already; M2 adds an optional `type_name` to keep it, and falls back to
-> the short label for adapters that never populate it.
+> **`--format compact` is cut, and the type it wanted is added.** `CatalogItem.type_label`
+> is a 1–3 character label for a tree column (`##`, `s`, `ts`), so the type an agent needs
+> is not in the contract at all — M2 adds an optional `type_name`, which the in-tree
+> adapters already fetch and discard, falling back to the short label where an adapter
+> never populates it. The *format* goes: `parent(child TYPE, …)` only reads for a relation
+> and its columns, and what a level is belongs to the adapter, so a database's schemas
+> would render as a signature they are not. `-tA --format csv` over the same rows is as
+> cheap and is true at every level.
 
 ### `hsql find` — the real gap
 
@@ -623,6 +633,14 @@ and a wrong guess produces a confusing runtime error. The config is shared betwe
 > `tomllib` change introduced. M2 makes the merge per-key and treats it as a bug fix; the
 > spec page should describe that rather than what exists today.
 >
+> **Validation runs on the merged document, so no error can name a file.** `load_config()`
+> merges every discovered file and then validates the result, which means a diagnostic can
+> only ever point at a key in a structure nobody wrote. M2 validates each file as it is
+> read and merges only valid ones, which is what makes deliverable 2's file-and-line
+> diagnostics possible. It also adds `hsql --config list-profiles`, since the first
+> question a human asks is what they can pass to `-P`, and a missing name is how the merge
+> bug above becomes visible.
+>
 > **`${VAR}` interpolation has to be a read-path transform.** `harlequin --config` reads
 > the config and writes it back, so resolving values too early bakes a plaintext password
 > into the user's file on their next wizard run. The syntax is `${VAR}` and
@@ -650,9 +668,12 @@ Combine that with `${VAR}` interpolation (#898) and the secrets story closes:
   redaction everywhere else.
 - **Never prompt in headless mode.** A prompt that blocks on stdin is the *worst*
   failure mode for an agent: no output, no exit code, the whole turn burned until
-  something times out. `hsql` must fail fast with a message pointing at the profile, the
-  env var, or `--password-stdin` — and `--password-stdin` conflicts with `-f -`, so that
-  pair needs an explicit error rather than a mysterious hang.
+  something times out. `hsql` must fail fast with a message pointing at the profile or the
+  env var.
+
+  > **Amended in M2's plan: no `--password-stdin` either.** Between profiles, `${VAR}`
+  > interpolation and the variables every driver already reads, it fills no gap worth a
+  > flag — and the gap it does fill, it fills by consuming the stream `-f -` wants.
 - **`spec --json` marks secret options.** This is what teaches an agent *not* to
   construct `hsql --password hunter2`, which would leak through `ps` and shell history
   to every other user on the box. The skill should say it too, but the machine-readable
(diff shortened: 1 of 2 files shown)


docs/plans/harlequin-for-agents.md
```diff
@@ -151,6 +151,14 @@ Everything in Workstream B lands as `hsql` subcommands: `hsql catalog`,
 `hsql mcp`, `hsql open`. Bare `hsql` with no arguments prints help; it never launches
 the TUI.
 
+> **Amended in M2's plan.** Subcommands need the collision rule this section raises
```
tconbeer: I'm not sure this is the right approach. It might be better to have option flags for the subcommands, like hsql --catalog. This would make it consistent with harlequin, eg harlequin --config

docs/plans/harlequin-for-agents.md
```diff
@@ -301,6 +309,21 @@ identically whether stdout is CSV or Parquet.
...
+> **Amended in M2's plan**, on two of the safety flags above.
+>
+> **`--timeout` has to attribute the cancellation itself.** A cancelled DuckDB query comes
+> back as an empty result set with no error — `fetchall()` swallows the interrupt and
+> returns `None`, which is also what an empty result looks like — so a timeout that trusts
+> the adapter prints "no rows" and exits 0. `hsql` knows it fired the deadline, discards
+> that result, and exits 4. It also refuses up front on an adapter that can't cancel,
+> since there is otherwise no way to stop the work.
+>
+> **`--dry-run` is a parse check, not an existence check**, on the adapter it will be used
```
tconbeer: I don't like --dry-run enough to justify this complexity. In practice only duckdb implements validate anyway. Let's punt on this feature to a future milestone, or never.

docs/plans/harlequin-for-agents.md
```diff
@@ -428,6 +451,27 @@ hsql find <TERM> [--in tables|columns|all]
...
+> **`--format compact` needs a field that doesn't exist.** `CatalogItem.type_label` is a
+> 1–3 character label for a tree column (`##`, `s`, `ts`), so today's honest rendering is
+> `total #.#`, not `total DECIMAL(18,2)`. The full type is fetched and discarded by the
+> in-tree adapters already; M2 adds an optional `type_name` to keep it, and falls back to
+> the short label for adapters that never populate it.
+>
+> **Child counts need one too** — an optional `child_count`, where `None` means unknown
+> rather than zero, so an agent can't read a missing count as an empty schema.
+>
+> Measured while writing that plan: `--depth 2` on a schema with 400 relations is 403
```
tconbeer: I think depth 2 would just fetch schema names, right?

docs/plans/harlequin-for-agents.md
```diff
@@ -486,6 +530,13 @@ redacted.
 Capability flags matter: they let an agent know *not* to try `--dry-run` on an adapter
 that can't validate, instead of trying and failing.
 
+> **Amended in M2's plan.** The flags have to be *declared* on the adapter class, not
+> detected on the connection. Reflection is measurably wrong: `HarlequinSqliteConnection`
+> defines `validate_sql` and raises `NotImplementedError` from it, so an override check
```
tconbeer: This isn't a bug, harlequin handles the NotImplementedError and short-circuits the validation call. But generally you're right and it's a better pattern to set the classvar

docs/plans/harlequin-for-agents.md
```diff
@@ -541,6 +592,22 @@ and a wrong guess produces a confusing runtime error. The config is shared betwe
 6. **A formal spec page on the site**: discovery order, merge semantics, precedence,
    every key with its type and default, and the reserved/invalid names.
 
+> **Amended in M2's plan**, on two things this workstream would otherwise document as they
+> stand.
+>
+> **The merge is per top-level key, not per profile.** `_merge_config_files()` is a
+> `dict.update()` per file, so a cwd config file that defines any profile hides *every*
+> profile in the home file. Nobody writes a project-local profile in order to lose their
```
tconbeer: I don't think this is true. Test it.

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,883 @@
+# M2 Technical Plan — self-description and safety
+
+Implementation plan for milestone M2 of [Harlequin for agents](./harlequin-for-agents.md),
+following [the M1 technical plan](./m1-hsql-technical-plan.md). The product plan says what
+`hsql` is for; M1 built the command and the execution core under it. This one is about the
+verbs that let an agent write a correct query before it runs one, and the three flags that
+let a human hand an agent a warehouse credential.
+
+**M2's scope, from the roadmap:** `catalog` (one level below the match, child counts, node
+budget), `describe`, `info --json`, `spec --json`, `fmt`, `config validate/show/schema/init`,
```
tconbeer: I want to remove fmt, I don't really see the point.

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,883 @@
...
+But there is no way to *reach* an item except by fetching its ancestors. To call
+`fetch_children()` on `mydb.analytics.orders` you need that item object, which only
+`SchemaCatalogItem.fetch_children()` can hand you, which needs the schema item, which needs
+the database item. I built a DuckDB file with 400 relations in one schema and counted the
+round trips:
+
+| Call | Children | Round trips | Local ms |
+| --- | --- | --- | --- |
+| `hsql catalog` | 1 database | 1 | 1.9 |
+| `hsql catalog wide` | 2 schemas | 2 | 10.2 |
```
tconbeer: This prototype syntax doesn't make sense. We still need a conn str. So this should really be:
hsql --catalog -c "wide"

I think it makes sense to reuse -c here but if agents would find that confusing then --path or --relation could work.

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,883 @@
...
+| `hsql catalog wide.analytics` | 400 relations | 3 | 12.8 |
+| `hsql catalog wide.analytics.t000` | 3 columns | 4 | 22.8 |
+| `hsql catalog wide.analytics --depth 2` | 400 relations + 1200 columns | **403** | **2303** |
```
tconbeer: Oh now I understand the comment in the design doc. Maybe we don't allow --depth 2 and force the client to fetch one relation's children at a time?

Alternatively, maybe fetching async/concurrently can help? Maybe hsql should establish multiple connections simultaneously for --depth queries

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,883 @@
...
+which tells an agent almost nothing and would send it back to `select * limit 0` to find
+out what it is joining on.
+
+The data is right there and discarded. `DuckDbConnection._get_columns()` selects
+`column_name, data_type` from `information_schema.columns` and then does
+`type_label=self.connection._short_column_type(column_type)`, dropping `column_type` on the
+floor. SQLite's is the same shape. So this is a field on `CatalogItem`, not a new query
+(§3.2).
+
+Neither is there anywhere to hang a child count. `400 relations` next to a schema is what
```
tconbeer: Nor should there be? hsql can build that count when doing --depth 2 but there isn't any other way to get the count then to fetch the children.

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,883 @@
...
+`config_wizard` reads `ConfigFile.relevant_config` and writes it back through `update()`, so
+**anything that transforms values on read will be written back transformed** unless the read
+path used for writing stays raw. That is the whole implementation risk in env interpolation
+(§3.5).
+
+Schema errors (`_raise_on_bad_schema`) are good prose with no machine channel and no
+positions: `tomllib` reports a line for a *parse* error inside its message, and neither
+parser will tell you where key `profiles.prod.limitt` sits in the file.
+
+### 1.7 sqlfmt is a dependency, and the headless path is contractually forbidden to import it
```
tconbeer: Yeah just cancel the hsql fmt plan

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,883 @@
...
+Two more things I measured while establishing what a deadline can do:
+
+- **Exiting while a worker thread is inside DuckDB aborts the process.** `sys.exit(4)` with
+  a live daemon thread in `to_arrow_table()` produces `terminate called without an active
+  exception` and exit code **134**, which is not in the documented table. `os._exit(4)`
+  after an explicit flush exits 4 with stdout intact.
+- So a deadline is implementable, but only as: cancel → wait → flush → `os._exit`. And with
+  an adapter that cannot cancel, there is no way to stop the work at all.
+
+### 1.10 `--dry-run` on DuckDB catches syntax errors and nothing else
```
tconbeer: Let's cancel dry run also.

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,883 @@
...
+schema generation are all things to do to the config we already read.
+
+---
+
+## 3. Target architecture
+
+```
+harlequin/
+  adapter.py      MOD  + IMPLEMENTS_READ_ONLY, IMPLEMENTS_VALIDATE_SQL, capability reporting
+  catalog.py      MOD  + CatalogItem.type_name, CatalogItem.child_count
```
tconbeer: No child count, unless it's a derived prop of len(children) that hydrates later after children are fetched.

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,883 @@
...
+  navigate.py     NEW  locating and listing catalog items by path, under a budget
+  options.py      MOD  + secret=..., + to_dict()
+  config.py       MOD  per-key merge, provenance, ${VAR} interpolation
+  config_schema.py NEW the JSON Schema for a config file, generated from what is installed
+  redact.py       NEW  what must never be printed, in one place
```
tconbeer: "What must never be printed" are specific options, right? Is this utilities for redaction?

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,883 @@
...
+  query.py        MOD  text_columns() returns string data unchanged
+  layout.py       MOD  + compact, tree
+  hsql/
+    cli.py        MOD  three-phase parse: verb, then adapter, then the command
```
tconbeer: No thanks. Let's use options instead of sub commands.

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,883 @@
...
+    verbs.py      NEW  the verb table -- names and help lines, importing nothing
+    deadline.py   NEW  a wall clock over a run, and how it exits
```
tconbeer: Why not timeout.py ?

tconbeer (commented):
Through S 3.1. Comments need addressing before it makes sense to dig into more details.

docs/plans/harlequin-for-agents.md
tconbeer: Tested — it holds, and it predates the `tomllib` refactor (same failure from the pre-refactor `config.py`). The full reproduction is in [the reply on the other thread](https://github.com/tconbeer/harlequin/pull/1038#discussion_r3781067274); short version is that the cwd file replaces the home file's whole `profiles` table while its `default_profile` survives, so both commands exit with `HarlequinConfigError: Config files set the default_profile to personal, but do not define a profile with that name`. The amendment now says that rather than the milder "hides your profiles".

---

docs/plans/m2-hsql-technical-plan.md
tconbeer: You're right that the syntax was wrong — a conn_str is always needed. Rewritten throughout as mode options, and the table rows now read `hsql --catalog --path wide.analytics wide.db`.

On the path argument I took `--path` rather than reusing `-c`, of the two you offered. `-c` is `--command` and means *SQL* here, in psql and in duckdb, and principle 4 is that a flag shouldn't change meaning with context. An agent that has learned `-c` will eventually type `hsql --catalog -c "select 1"`, and the right answer to that is an error, not a lookup for a relation named `select 1`. `--path` also reads correctly for `--describe`, which needs the same argument, and it leaves `-c` free if a later milestone ever wants a catalog call and a query in one invocation.

---

docs/plans/m2-hsql-technical-plan.md
tconbeer: Took your first option: **M2 ships one level and no `--depth` at all.** The reason the flag can't be made safe is that its cost depends on the level it starts from — depth 2 from the root is one round trip per database, depth 2 from a schema is 403 — so no single number is safe everywhere, and the level where it's worst is the one an agent most wants. Recursion comes back with `fetch_descendants()`, which turns it into one query for adapters that can express it; a `--depth` on top of a walking default would just be the same cliff with a nicer name.

Dropping recursion also removed the node budget: one level is one round trip, and a 400-row listing is a *display* problem that `--display-rows` and the `(40 of 400 rows)` footer already solve.

On concurrency — I don't think it's the first answer. The adapter contract says nothing about a connection being usable from two threads, and offers no way to open a second one from an existing connection, so `hsql` would have to `connect()` N times (re-authenticating each time) and hope; it also multiplies load on the database for a call the agent shouldn't be making in bulk. It's recorded in §3.8 as worth revisiting if `fetch_descendants` lands and is still too slow.

---

docs/plans/harlequin-for-agents.md
tconbeer: Yes — from the root, depth 2 is schema names, one round trip per database, and databases are few. The 403 was depth 2 from a *schema*, where it's one round trip per relation. That's the whole problem with the flag: one number over levels with wildly different costs. Both docs now say which level they mean, and M2 ships one level with no `--depth` at all.

---

docs/plans/harlequin-for-agents.md
tconbeer: Fair — corrected. The amendment no longer calls it a bug; it now says nothing is broken today because the app catches the error at the call site, and makes the narrower point: no probe can tell a raising stub from a real implementation, so the classvar is the pattern that's readable *before* the call, and the only one `--info` can read while the database is unreachable.

---

docs/plans/m2-hsql-technical-plan.md
tconbeer: Agreed, and the field is gone. §1.3 now says it: no adapter can produce a count without fetching the children, so a `child_count` would be `None` everywhere or a fetch in disguise. The count that's knowable is `len(children)` once they're loaded, which is what a listing's own footer already reports. It was only ever proposed as a mitigation for the recursion M2 no longer ships, so it goes with it.

---

docs/plans/m2-hsql-technical-plan.md
tconbeer: Right on both counts — it's redaction utilities, and *what* to hide is specific options. The line now reads "redaction helpers, driven by the options that declare themselves secret," and §3.7 leads with the split: the module holds three helpers (`redact_profile`, `redact_conn_str`, `redact_text`), and the list of what's sensitive lives on the option as `secret=True`, so an adapter declares its own once and `--info`, `--spec`, `--config show`, the wizard's masked input and error rendering all inherit it.

---

docs/plans/m2-hsql-technical-plan.md
tconbeer: No reason — renamed to `timeout.py`.

---

tconbeer:
Revised in `00cbb2d` for the whole review. Threads with a question or a divergence have replies; this covers the rest, all of which are straight compliance:

| Your comment | What changed |
| --- | --- |
| Options, not subcommands (×2) | Whole plan rewritten around mode options: `--catalog`, `--describe`, `--info`, `--spec`, `--config show\|validate\|schema\|init`, with `--path` for the catalog path. §3.1 is now "Modes are options, not subcommands"; `verbs.py` and the `commands/` package are gone, and so is the verb-vs-CONN_STR collision rule the previous draft had to invent. |
| Remove `fmt` (×2) | Cut, in §7, with the reason: an agent that wants formatted SQL can run `sqlfmt`, and this was the only headless mode that would have forced a 196ms import into an import graph that's otherwise an enforced contract. Also amended in the product plan's §6. |
| Cancel `--dry-run` (×2) | Cut, in §7, with the measurements that justify it (only DuckDB implements `validate_sql`, and there `select * from no_such_table` validates clean). `IMPLEMENTS_VALIDATE_SQL` survives as a declaration `--info` reports and the app can consult — no `hsql` consumer. |
| No child count | Field dropped; §1.3 rewritten. |
| `deadline.py` → `timeout.py` | Renamed. |

Two knock-on simplifications worth flagging, both from dropping recursion:

- **No node budget.** One level is one round trip, so there's nothing left to bound; a 400-row listing is a display problem, and `--display-rows` plus the `(40 of 400 rows)` footer already handle it. `--max-nodes` is gone.
- **Three fewer moving parts in `navigate.py`** — no `Budget`, no `stopped_at`, no depth bookkeeping. It's `resolve()`, `list_children()`, and a round-trip counter for `--stats`.

Net effect on the milestone: **one** new field on the catalog contract (`type_name`), two capability classvars, `secret` on options, and 17 PRs across the three releases instead of 19.

Still open in §6, and both are yours to call whenever: whether `--describe` earns its own mode at all (it resolves and lists exactly as `--catalog` does — the difference is meant to be per-object detail most adapters may not supply), and whether `--catalog` should reuse the IDE's catalog cache.

---

docs/plans/harlequin-for-agents.md
tconbeer: Follow-up from your last message, in `096f9ef`: `--describe` is gone. `--catalog --path mydb.analytics.orders` already returns the relation's columns with names, real types and quoted identifiers in one round trip below the resolve, which is all a describe was going to print — a second mode could only add comments (not in the contract) or row counts (a table scan).

That also closes both of §6's open questions: no describe means no question about its row counts, and `--catalog` does not reuse `catalog_cache.py` — noted in the plan as kept for compatibility rather than a live cache to read or warm. Repeat navigation re-resolves its prefix, at one round trip per segment.

Release A is now three PRs plus docs; the milestone is 16.

---

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,935 @@
...
+columns asks for one relation at a time, which is what the IDE's own catalog does and what
+Ted proposed in review. Recursion comes back when an adapter can answer it in one query
+(§3.8), and it arrives with the capability rather than ahead of it.
+
+**A trailing glob is sugar; an interior one is a different mode.** `--path analytics.ord*`
+filters one resolved parent's children in the client and stays one round trip.
+`--path *.orders` cannot be evaluated without fetching every candidate level, so it belongs
+to `--find` (§3.8) and is refused here rather than quietly walked.
+
+**Round trips are counted and reportable.** `--stats` gains `round_trips`, so the cost of a
```
tconbeer: Without depth we don't need this. Remove the round trip counting.

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,935 @@
...
+vertical|parquet`, `-o PATH`, `-t/-A/--no-header`, `--display-rows` and the byte-for-byte
+determinism M1's snapshots already pin. The alternative is a second renderer for the same
+job, which is how `--format json` and a catalog's JSON end up disagreeing about how to spell
+null.
+
+Two new layouts, both of which read the path column and neither of which knows any types:
+
+- **`compact`** — `analytics.orders(id BIGINT, customer_id BIGINT, total DECIMAL(18,2))`,
+  the token-efficient form, roughly an order of magnitude cheaper than pretty JSON.
+- **`tree`** — indented, for a human reading over an agent's shoulder.
```
tconbeer: I don't think we need tree - we're fetching one node's children only

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,935 @@
...
+`ResultSet` — `create_backend()` takes a sequence of rows and the column names, with no
+database involved — means `--catalog` inherits `--format csv|json|jsonl|markdown|table|
+vertical|parquet`, `-o PATH`, `-t/-A/--no-header`, `--display-rows` and the byte-for-byte
+determinism M1's snapshots already pin. The alternative is a second renderer for the same
+job, which is how `--format json` and a catalog's JSON end up disagreeing about how to spell
+null.
+
+Two new layouts, both of which read the path column and neither of which knows any types:
+
+- **`compact`** — `analytics.orders(id BIGINT, customer_id BIGINT, total DECIMAL(18,2))`,
```
tconbeer: This doesn't really work except for tables and columns but we said above that the type of node is up to the adapter...

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,935 @@
...
+them means loading the adapter class. That is the one cost this mode pays — 309ms for four
+installed adapters, against 51ms for names and versions alone (§1.4) — and it is the right
+trade for a once-per-task lookup whose entire job is that the agent stops guessing.
+`hsql --info -a duckdb` imports one (114ms) for the common case of asking about the adapter
+you are using, and an adapter that fails to import is reported with its capabilities as
+`"unknown"` and the import error alongside — never as `false`, because guessing false about
+`implements_read_only` is the direction that gets someone hurt.
+
+### 3.5 Config: one merge, provenance, `${VAR}`, and a schema
+
```
tconbeer, line 580: For the config implementation, I want to explore using a Pydantic model instead of rolling this ourselves. Interested in both perf and simplifying code. (To be clear, the Pydantic config module won't work, but a Pydantic model could declare the config schema, provide validation, and grow a method to merge itself). 

Also, the current validation workflow is backwards. We merge configs and then validate the merged dict. Instead we should validate every file individually so we merged well-structured and valid configs only. This should make good error messsages much easier.

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,935 @@
...
+  own test.
+- Shell-command interpolation (asked for on #898) is out: it is arbitrary code execution
+  driven by a file discovered from the working directory.
+
+**`--config validate`** reports `{file, key_path, message, line?}` per problem. Line numbers
+only where the parser gives one — `tomllib` puts a line in its parse-error message, and
+neither parser reports positions for a key we merely dislike. Saying "line unknown" is better
+than a made-up number and better than not shipping the mode.
+
+**`--config schema`** generates a JSON Schema for `.harlequin.toml` from the config model
```
tconbeer, line 679: We would get this for free with a pydantic model.

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,935 @@
...
+   diagnostic is `hsql: error: timed out after 30s`, and the exit code is 4 — the number
+   `ExitCode.TIMEOUT` has been reserving since M1.
+5. If the grace period expires too, flush both streams and `os._exit(ExitCode.TIMEOUT)`.
+   Plain `sys.exit` with a live worker inside DuckDB aborts the interpreter and exits 134
+   (§1.9), which would be a documented-exit-code violation caused by the flag that exists to
+   make failure legible.
+
+`SIGINT` takes the same path and exits 130, which it already does.
+
+**`--single-transaction`** switches the connection to its manual transaction mode, runs the
```
tconbeer: Right now adapters can name their transaction modes whatever they want, and default ti undefined transaction behavior. This will require a different approach to transaction modes. I don't think it's worth that complexity when the user can just wrap their query in begin/commit.

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,935 @@
...
+Two things the declaration cannot cover, both with a helper of their own above:
+
+- **Connection strings are positional**, so no option type describes
+  `postgres://user:pw@host/db`. `redact_conn_str` is DSN-aware — and
+  [#354](https://github.com/tconbeer/harlequin/issues/354) is evidence people really do put
+  passwords there.
+- **A driver exception that echoes a DSN** does not pass through our option layer, so
+  `redact_text` is the backstop applied to error messages before they are printed.
+
+**`--password-stdin`** reads one line from stdin and assigns it to the adapter's option named
```
tconbeer: I don't like this. Take it out.

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,935 @@
...
+recursive walk would help wall-clock time and nothing else, and it costs more than it looks:
+the adapter contract says nothing about a connection being usable from two threads, nor does
+it offer a way to open a second one from an existing connection, so `hsql` would have to
+re-`connect()` (re-authenticating N times) and hope. It also multiplies load on the database
+for a call the agent should not be making in bulk. Worth revisiting if `fetch_descendants`
+lands and is still too slow; not worth it as the first answer.
+
+---
+
+## 4. Sequencing
```
tconbeer, line 798: No, do config/self description first. Those address existing bugs and require no adapter changes. Then catalog, including search implementations for the duckdb and SQLite adapters, then safety.

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,935 @@
...
+
+```python
+@dataclass(frozen=True)
+class Provenance:
+    value: Any
+    path: Path  # which file this key's winning value came from
+    overrode: list[Path]
+```
+
+`--config show` prints the effective config with a `# from ~/.harlequin.toml` per key, and
```
tconbeer, line 645: Let's add a --config list-profiles that prints the names of all found profiles.

docs/plans/m2-hsql-technical-plan.md
```diff
@@ -0,0 +1,935 @@
...
+## 5. Testing
+
+The M1 guards stay and grow; the new ones are the same kind — deterministic, and failing
+identically on every machine.
+
+- **Import hygiene, extended per mode.** `hsql -c "select 1"` imports no `tomlkit`, no
+  `questionary` and (still) no `sqlfmt`; `hsql --help` imports no adapter; `hsql -a sqlite
+  --catalog` imports no `duckdb`. Each is a subprocess reading `sys.modules`, and each
+  deferral has a matching `ignore_imports` entry with a comment at the import site.
+- **A counting fake connection** — a `HarlequinConnection` whose `fetch_children` increments
```
tconbeer: This is unnecessary

docs/plans/m2-hsql-technical-plan.md
tconbeer: Both halves taken, but the second one comes with a measurement that argues against Pydantic specifically.

**Validation before the merge: yes, unreservedly.** `load_config()` now reads as `merge([validate(f) for f in files])`. It costs nothing, it means every diagnostic can name the file it came from, and the merge only ever combines inputs already known to be well-formed. It's PR 1 of Release A.

**On the model.** The cost is model *construction*, not import, and config is read on every invocation of both commands — so it lands on the query path with no lazy way out, because the read path is the validating path:

| | over a bare interpreter | modules |
| --- | --- | --- |
| `import pydantic` | +52ms | 99 |
| defining `Config`/`Profile` as `BaseModel`s | **+135ms** | 207 |
| validating, or `model_json_schema()` | free once defined | — |
| `import msgspec` | +37ms | 104 |
| defining the same as `msgspec.Struct`s | free once imported | 104 |
| `msgspec.convert()`, `msgspec.json.schema()` | free | 104 |

+135ms takes `hsql -c "select 1"` from 257ms to ~390ms, past the 300ms §13 tracks in CI. So Pydantic is a no on perf, which was one of the two things you wanted from it.

`msgspec` gets all three of the things you named — declared schema, path-aware validation errors, and the JSON Schema for free — for +37ms, requires Python ≥3.10 (our floor exactly) and ships cp310–cp314 wheels (our matrix exactly). Still 14% of a `hsql -c` run, so it's a trade rather than a free win, and it's your call: I've left it as §6's one open question, to settle in PR 1's review rather than later, since PR 1's validation and PR 6's schema both read whatever it decides.

---

tconbeer:
Second review pass applied in `926615f`. One reply in-thread on the Pydantic question, since that one needed numbers; the rest, briefly:

- **Releases reordered.** Config and self-description first (PRs 1–8), then the catalog with `--find` implemented for DuckDB and SQLite (9–12), then safety (13–17). §4's rationale now leads with "fix what is broken before adding what is new."
- **Round-trip counting removed** — `Listing.round_trips`, the `--stats` key, and the test that asserted it. Without recursion the number is `len(path) + 1`.
- **`--single-transaction` cut.** Your point about adapter-named modes and undefined defaults is in §1.8 as the evidence and §7 as the decision: core can't ask for "the manual one" through an interface where the label is the adapter's choice.
- **`--password-stdin` cut.** §3.7 keeps "never prompt" and adds why no new channel is needed either.
- **`--config list-profiles` added**, and validation now runs per file ahead of the merge — which is also what lets `--config validate` name a file.
- **`redact.py`** is described as redaction helpers, with the list of what's sensitive living on the options as `secret=True`. **`timeout.py`** renamed.

One judgment call worth your eyes: you questioned `compact` on the grounds that it only works for tables and columns while node types are the adapter's. I took that as fatal to the format rather than fixable and **cut `compact` along with `tree`** — a database's schemas rendered as `mydb(analytics sch, main sch)` is a signature they aren't. `-tA --format csv` over the same rows is as cheap and stays true at every level, so nothing is lost but the name. Say the word if you'd rather keep it scoped to relation→column.

---

PR #1041, opened 2026-08-14
fix: merge profiles and keymaps by name across config files (#1040)

### LINKED ISSUE #1040: Config files merge per top-level key, so a second file's `profiles` table replaces the first's
Discovered config files are merged with a shallow `dict.update()` per file, so the whole `profiles` table is replaced rather than merged. A project-local config file that defines one profile hides every profile in the user's home file — and because `default_profile` is a *different* top-level key, it survives from the home file and points at a profile that is no longer there, so both commands refuse to start.

## Reproduction

`~/.harlequin.toml`:

```toml
default_profile = "personal"

[profiles.personal]
adapter = "duckdb"
theme = "nord"

[profiles.shared]
adapter = "duckdb"
theme = "gruvbox"
```

`./.harlequin.toml`, in a project directory:

```toml
[profiles.project]
adapter = "sqlite"
```

Then, from that directory:

```
$ harlequin
HarlequinConfigError: Config files set the default_profile to personal, but do not
define a profile with that name.
```

`hsql` fails the same way, with exit code 2.

The merged config shows what happened — `default_profile` from the home file, `profiles` only from the cwd file:

```python
>>> from harlequin.config import _find_config_files, _merge_config_files
>>> _merge_config_files(_find_config_files(None))
{'default_profile': 'personal', 'profiles': {'project': {'adapter': 'sqlite'}}}
```

Even without a `default_profile`, `harlequin -P personal` from that directory reports the profile as nonexistent.

## Expected

Profiles from every discovered config file are available, with later files winning key by key: a cwd file that sets `limit` in a profile the home file defines should override that key and leave the rest of the profile — and the other profiles — alone. That is what "later wins" reads as, and it is the merge the docs' precedence order implies.

## Cause

`_merge_config_files()` in `src/harlequin/config.py`:

```python
def _merge_config_files(paths: list[Path]) -> Config:
    config: Config = {}
    for p in paths:
        config_file = ConfigFile(p)
        config.update(config_file.relevant_config)
    return config
```

`dict.update()` replaces the value at each top-level key, so the second file's `profiles` table displaces the first's entirely. `keymaps` has the same shape and the same problem.

## Not caused by the recent parser change

This predates the `tomllib` read path from #1032: running the same reproduction against the pre-refactor `config.py` (from `1af8cca`, which read through tomlkit) produces the identical error. `dict.update()` has always replaced the table rather than merging it.

## Fix

Merge per profile name, and per key within a profile; likewise for `keymaps`. Planned in the M2 technical plan (`docs/plans/m2-hsql-technical-plan.md`, §1.6 and §3.5), where it lands alongside `hsql --config show`, which reports per-key provenance and is what makes a wrong merge visible. Worth pulling forward independently of that milestone — the failure above is a hard stop for anyone whose home config sets a default profile and who then adds a project-local one.

DSeaStar:
Closes #1040

`_merge_config_files()` used `dict.update()`, so a later file's `profiles` (or `keymaps`) table replaced the earlier one entirely. A project-local file that defined one profile therefore hid every profile in the home file — and because `default_profile` is a different top-level key, it survived and pointed at a name that was no longer there.

The merge is now by name: later files still win, but they win *key by key* inside a profile, and they add or replace individual keymaps rather than the whole table. Other top-level keys (`default_profile`) keep later-wins.

This is the merge #1040 asked for, pulled forward on its own rather than waiting for the M2 `hsql --config show` work. I did not deep-merge keymap *bindings* (they are a list, not a table of keys); a later keymap of the same name replaces the earlier one, which matches "later wins" without inventing a binding-identity rule.

- [x] No.

- [x] Yes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ All notable changes to this project will be documented in this file.
 
 ### Bug Fixes
 
+- Config files now merge `profiles` and `keymaps` by name, so a project-local file no longer hides profiles defined in an earlier file ([#1040](https://github.com/tconbeer/harlequin/issues/1040)).
 - The Results Viewer no longer reports a truncated fetch as an exact total ([#1026](https://github.com/tconbeer/harlequin/issues/1026)).
 - The SQLite adapter now honors a limit of `1` or `0`.
 - A config file that sets `default_profile` and defines no profiles at all is now reported as a `HarlequinConfigError`, instead of crashing ([#1032](https://github.com/tconbeer/harlequin/pull/1032)).
--- a/src/harlequin/config.py
+++ b/src/harlequin/config.py
@@ -325,11 +325,69 @@ def _search_home() -> list[Path]:
 
 
 def _merge_config_files(paths: list[Path]) -> Config:
+    """Merge discovered config files. Later paths win.
+
+    `profiles` and `keymaps` are merged by name rather than replaced, so a
+    project-local file that defines one profile does not hide every profile
+    in an earlier file. Within a profile, later files win key by key. A
+    later keymap of the same name replaces the earlier one.
+    """
     config: Config = {}
     for p in paths:
-        config_file = ConfigFile(p)
-        config.update(config_file.relevant_config)
+        _apply_config(config, ConfigFile(p).relevant_config)
     return config
+
+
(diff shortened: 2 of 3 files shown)


tconbeer:
Thanks, this isn't quite the right solution. Proper fix coming soon.

PR #1044, opened 2026-08-15
fix: escape --null-string in markdown output

VXNCXNX:
Closes: I could not find a matching existing open issue. Happy to add a reference if there is one I missed.

`--null-string` is printed into markdown output unescaped, so a `|` or a newline in it breaks the table.

```
$ hsql -a sqlite -c "select null as x" --format markdown --null-string "a|b"
| x    |
| ---- |
| a|b  |
```

One column in the header, two cells in the row. With a newline the row splits in two:

```
$ hsql -a sqlite -c "select null as x, 1 as y" --format markdown --null-string "$(printf 'a\nb')"
| x      | y   |
| ------ | --- |
| a
b    | 1   |
```

`_MarkdownLayout.write` escapes headers and values, and `_column_widths` is already called with `_escape(self.null)`. But `_BaseLayout._cell` renders `self.null` as it finds it, so the width was measured from the escaped string while the unescaped one was printed. That is also why the column misaligns.

The escape now happens once in `_MarkdownLayout.__init__`. `_column_widths` then takes the already-escaped value and the two agree by construction.

After:

```
| x    |          | x      | y   |
| ---- |          | ------ | --- |
| a\|b |          | a<br>b | 1   |
```

The other layouts are byte-identical before and after, still printing `a|b` raw:

```
 x   | y            -[ RECORD 1 ]-        x,y
-----+---           x | a|b               a|b,1
 a|b | 1            y | 1
```

`_cell` is shared with the table and vertical layouts, which must keep printing the null string raw, so escaping at the render site was not an option. Doing it once at construction is the only place where the width calculation and the render can be made to agree by construction rather than by coincidence.

Verification: `test_a_null_string_cannot_escape_its_column`, directly beside `test_a_cell_cannot_escape_its_column`, whose docstring already states the invariant: "A `|` or a newline in a value doesn't make a table hard to read; it makes it a different table". The sibling gives the mutation check for free, since it passes today: with the fix reverted the new test fails and the sibling still passes.

`make check` passes: 641 tests, mypy clean over 105 files, ruff clean, 4 contracts kept.

- [x] No. (bug fix to markdown escaping, no documented behaviour changes)

- [x] Yes.

- [x] I have added an entry to `CHANGELOG.md`, under the `[Unreleased]` section heading. That entry references the issue closed by this PR. (entry added under `[Unreleased]`; it references no issue because I could not find a matching open one)

*Disclosure: written with AI assistance (Claude Code). I reproduced the issue, ran the change and the verification myself.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- `hsql --format markdown` no longer breaks the table when `--null-string` contains a `|` or a newline.
+
 ## [2.9.0] - 2026-08-15
 
 ### Breaking Changes
--- a/src/harlequin/layout.py
+++ b/src/harlequin/layout.py
@@ -243,6 +243,13 @@ class _MarkdownLayout(_BaseLayout):
     MIN_WIDTH = 3
     """The shortest delimiter row GFM accepts is `---`."""
 
+    def __init__(self, options: LayoutOptions) -> None:
+        super().__init__(options)
+        # `_cell()` is shared with the other layouts and prints `self.null` as
+        # it finds it, so a null string a caller chose is escaped once, here,
+        # rather than at every place that renders one.
+        self.null = _escape(self.null)
+
     def write(self, result: "ResultSet", out: TextIO) -> None:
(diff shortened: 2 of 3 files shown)


tconbeer:
Couldn't push to your branch; going to merge this as #1051. Thank you for your contribution!

PR #1045, opened 2026-08-15
fix: --limit 0 footer reads (0 of >0 rows)

VXNCXNX:
Closes: I could not find a matching existing open issue.

`--limit 0` prints a footer that contradicts itself.

```
$ hsql -a sqlite -c "select 1 as a" --limit 0
 a
---
(0 of >0 rows)
note: results truncated at --limit 0; pass --limit -1 for all rows
```

`0 of >0` puts two zeroes on either side of a word that promises they differ. `>0` asserts a row exists while the count beside it says none does.

The bug is only in the rendering. `_footer` takes the truncated branch with `shown = 0` and `rows = fetch_limit - 1 = 0`, and formats a ratio out of two zeroes.

`(>0 rows)` when the kept count is zero, ratio dropped, bound kept:

```
$ hsql -a sqlite -c "select 1 as a" --limit 0
(>0 rows)
note: results truncated at --limit 0; pass --limit -1 for all rows
```

Everything else is unchanged:

```
--limit 5 over 1 row     (1 row)
--limit 5 over 6 rows    (5 of >5 rows)   + truncation note
--limit -1               (1 row)
empty result             (0 rows)
```

`query.py` is untouched. The `truncated` and `fetched_row_count` pair is already right; only the string built from them was wrong.

The alternative was to treat `max_rows == 0` as not-truncatable, so it reads `(0 rows)`. I did not take it because the docstring right above already argues the opposite case: `>N` is chosen over `N+` precisely because the limit-plus-one fetch **proves** a further row exists, and `>N` is the stronger claim. Suppressing truncation here would discard that proof and make a non-empty table read identically to an empty one. The ratio is the part that carries no information when both sides are zero, so that is the part that goes.

Verification: three tests in `TestFooter`, including one across all three layouts, plus one in `test_query.py` pinning that limit 0 over a genuinely empty result is not truncation, since that is where the invariant lives rather than in the rendering.

`test_limit_zero_fetches_a_header_and_no_rows` asserted the old string, so it is updated to the new one. That is the one existing test this changes, and it is the assertion the bug was hiding behind.

With the fix reverted, six tests fail. `make check` passes: 647 tests, mypy clean over 105 files, ruff clean, 4 contracts kept.

- [x] No. (the footer string is not reproduced in the docs)

- [x] Yes.

- [x] I have added an entry to `CHANGELOG.md`, under the `[Unreleased]` section heading. That entry references the issue closed by this PR. (entry added under `[Unreleased]`; it references no issue because I could not find a matching open one)

*Disclosure: written with AI assistance (Claude Code). I reproduced the issue, ran the change and the verification myself.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- `hsql --limit 0` no longer prints a footer that reads `(0 of >0 rows)`; a result truncated to no rows now reads `(>0 rows)`.
+
 ## [2.9.0] - 2026-08-15
 
 ### Breaking Changes
--- a/src/harlequin/layout.py
+++ b/src/harlequin/layout.py
@@ -191,10 +191,16 @@ class _BaseLayout:
         `at least N` and would be the weaker statement of the two. The noun
         agrees with the total rather than with what was printed, which is why a
         truncated result is always plural.
+
+        `--limit 0` is the one place the ratio degenerates: nothing was kept and
+        the limit is the total's lower bound, so `0 of >0` puts two zeroes on
+        either side of a word that promises they differ. The claim is still the
+        one worth making -- the probe row proves a first row exists -- so the
+        ratio goes and the bound stays, as `>0 rows`.
         """
         rows = result.fetched_row_count
         if result.truncated:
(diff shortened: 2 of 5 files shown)


tconbeer:
That statement is true. You are displaying zero rows but the table you are querying contains more than 0 rows. So zero of more than zero, or 0 of >0

PR #1046, opened 2026-08-15
Validate config per file, then merge per profile, nearest first

### LINKED ISSUE #1040: Config files merge per top-level key, so a second file's `profiles` table replaces the first's

tconbeer:
Closes #1040

M2 PR 1, per [the M2 technical plan](``https://github.com/tconbeer/harlequin/blob/claude/m2-pr1-hsql-aw1u7s/docs/plans/m2-hsql-technical-plan.md#35-config-validate-each-file-then-merge-with-provenance),`` plus the reversed read order and early stopping asked for on top of it. **`config.py` was rewritten after the first review** — the per-comment replies are on the threads; the summary is at the bottom.

- **Discovery is reversed, and there is one generator.** `_discover_config_files()` yields paths highest priority first — an explicit `--config-path`, then cwd, the user config dir, home, and within each directory `harlequin.toml` before `.harlequin.toml` before `pyproject.toml`. Every caller uses it, including `get_highest_priority_existing_config_file()`, which is now "the first one with a section of ours".
- **The merge is per profile, and the nearest file wins.** `profiles` merges by name and `keymaps` by keymap name, so a project-local file no longer replaces the whole `profiles` table while leaving the `default_profile` naming one of them behind — the combination that made both commands refuse to start.
- **`load_profile()` stops at the file that answers the question**, and never opens the ones behind it; `hsql -P None` opens none. `load_config()` is the whole document, for the IDE's keymaps and for `--config show` in PR 2. The module docstring carries the table of which caller reads how much.
- **`Config` is the TypedDict *and* the msgspec model**, so there is one declaration of what a config file may say, `msgspec.convert(raw, Config)` is the whole of a file's shape check, and what comes back is the plain dict everything already used. Validation runs per file, before the merge, so an error names the file and the key: `Expected table, got string, at profiles.`
- **`parse_profile_options()` parses, rather than merely refusing.** It builds a struct from the adapter's own `ADAPTER_OPTIONS` and converts the profile's remaining keys into it, so a value reaches the adapter as the type its adapter declared. Each command passes the keys it reads for itself — its own click params plus `TUI_ONLY_KEYS` (`hsql`) or `HSQL_ONLY_KEYS` (the IDE) — so no third list of profile keys exists to drift.
- **Resolving a name is separate from reading the document.** A `default_profile` that names no profile is raised where the name is used, so `-P other` and `-P None` start rather than being refused over a key they overrode.

- **Per-option merging within a profile is gone, and that is the price of early stopping.** A profile's keys are not known to be complete until every file has been read, so the two features are incompatible; a profile is now one file's, whole. Half a connection from each of two files is not a connection either of them describes. Breaking Changes entry.
- **Parsing option values is a behavior change, deliberately.** `port = 5432` arrives as `"5432"` and `read_only = "true"` as `True`; `mode = "RO"` matches a declared choice by casefold, as click does. Two things that used to pass now fail, and both were silently doing something wrong: `extension = "httpfs"` for a list option (it was being iterated one character at a time) and `port = true` for a text option.
- **The first version had msgspec beside the data rather than as it** — a file model and a core-profile model that were both discarded after validating, a `_ReadFile` that was `Config` plus a path, a hand-maintained key list, and lru_cached model factories. The rewrite deleted all of that. What stayed hand-written is three name rules msgspec cannot express (a profile called `None`, `read-only` for `read_only`, `keymap_names` for `keymap_name`), the unknown top-level key (msgspec drops what a TypedDict does not declare rather than refusing it — the allowed names are read off `Config.__annotations__`), and the "did you mean" for a misspelled adapter option, which is computed in the `except` branch so it costs nothing on the happy path.
- **Cost.** msgspec is ~30ms to import and is deferred into the functions that use it, so an invocation that finds no config file pays nothing. Measured on the real console script, best of 11: with four candidate files present, and with exactly one, both `hsql --help` and `hsql -c "select 1"` are within noise of `main` (±3ms) — the files no longer read pay for the import. `hsql -P None -c "select 1"` is a few ms faster.
- **Not done here:** `harlequin` still `ep.load()`s every installed adapter to build its command (309ms with four). It needs `hsql`'s two-phase build shared between the two commands and it changes `harlequin --help`, so it wants its own PR — sketch and numbers are in [that thread](https://github.com/tconbeer/harlequin/pull/1046#discussion_r3790462195).

- [x] Yes; I haven't opened a PR, but the gist of the change is: the config topic needs its precedence section rewritten — files are read nearest first, a profile comes whole from the nearest file that defines it, and profiles/keymaps merge by name so a project-local file adds to your home config rather than replacing it. Worth saying that a profile key the adapter does not declare is now an error, that option values are parsed to their declared types, and that `harlequin.toml` outranks `.harlequin.toml` outranks `pyproject.toml` within a directory. M2's PR 8 is where the wider config-spec page lands.

- [x] Yes.

`TestMergingConfigFiles` pins the new semantics, with #1040's exact combination as its own test. `TestReadingNoMoreThanItMust` proves the early stop by putting invalid TOML in the file that must not be reached — a file that is opened at all takes the test down with it. `TestParsingAgainstAnAdapter` covers the second pass: what a profile may hold, what each value parses to, and what is refused (with the words the message has to contain). Plus an `hsql` end-to-end test that a profile option the adapter never declared exits 2 before connecting, and an import-hygiene subprocess test that a config with nothing in it imports no msgspec.


src/harlequin/cli.py
```diff
@@ -419,9 +420,24 @@ def inner_cli(
...
         config = merge_profile_with_cli(
-            profile=config, cli_values=kwargs, explicitly_set=explicitly_set
+            profile=profile_config, cli_values=kwargs, explicitly_set=explicitly_set
         )
 
+        # the second validation pass: the profile's own keys, against the
+        # options the adapter it names declares. Only the profile's -- click has
+        # already vetted everything typed on the command line.
+        adapter_name = str(config.get("adapter", None) or DEFAULT_ADAPTER)
+        if (declaring := adapters.get(adapter_name)) is not None:
```
tconbeer, line 431: Does harlequin still load/import all installed adapters? Let's fix that and have it only import the adapter found in the profile here.

src/harlequin/cli.py
```diff
@@ -405,7 +406,7 @@ def inner_cli(
         """
         # load config from any config files
         try:
-            config, user_defined_keymaps = get_config_for_profile(
+            profile_config, user_defined_keymaps = get_config_for_profile(
```
tconbeer: Can't harlequin also use get_profile here?

src/harlequin/config.py
```diff
@@ -46,6 +83,44 @@
 `LC_ALL` would be output a caller could not predict.
 """
 
+TOP_LEVEL_KEYS = ("default_profile", "profiles", "keymaps")
+"""Everything a config file may define at its top level."""
+
+CORE_PROFILE_KEYS = frozenset(
```
tconbeer: This is a long list to keep in sync manually. Did you consider other options for deriving this list?

src/harlequin/config.py
```diff
@@ -175,20 +250,19 @@ def write(self) -> None:
...
-            "exist in any discovered config files.",
-            title="Harlequin couldn't load your profile.",
-        )
-    else:
-        profile = config["profiles"][profile_name]
+    Keymaps come from every file, so this reads them all. A caller that only
+    wants the profile should call `get_profile()`, which stops at the file that
+    defines it.
+    """
+    config, named_by = _load(config_path)
```
tconbeer: named_by is a terrible name.

src/harlequin/config.py
```diff
@@ -199,6 +273,97 @@ def get_config_for_profile(
...
+
+    Files are read nearest first, so the first one to define the wanted profile
+    is the one that wins -- and the files behind it are never opened. A profile
+    the caller did not name is only wanted once some file says it is the
+    default, so the files read before that one are held until it does.
+    """
+    if profile_name == "None":
+        return {}
+
+    wanted = profile_name
```
tconbeer: These cute names like "wanted" and "held" are not acceptable. Use explicit names like "requested_profile"

src/harlequin/config.py
```diff
@@ -199,6 +273,97 @@ def get_config_for_profile(
...
+) -> None:
+    """Check a profile's keys against the options its adapter declares.
+
+    The second pass, and the only place that knows both halves: core cannot
+    reject a key it does not recognize, because every adapter's options live in
+    the same table, and an adapter takes supersets of what it declares -- so
+    `reed_only = true` is otherwise dropped in silence by the constructor, and
+    the connection a user believed was read-only is not.
+
+    Names and declared choices, not types: the adapter contract says values may
```
tconbeer: I don't see a lot of downside in validating types of the adapter declares them. Some existing config files might break but that should be a very easy fix for users.

I think it probably makes sense to widen the types from the files, though. So if adapter says "port" is a string and the file gives and int, no real harm in casting to a string?

Do the adapter options define types? Would it be much more complex to validate types also?

src/harlequin/config.py
```diff
@@ -199,6 +273,97 @@ def get_config_for_profile(
...
+    arrive uncast, so `port = 5432` and `port = "5432"` both have to pass.
+    """
+    declared = {sluggify_option_name(o.name): o for o in options or []}
```
tconbeer: Can we not use msgpack for this validation?

src/harlequin/config.py
```diff
@@ -199,6 +273,97 @@ def get_config_for_profile(
...
+            return file.profiles[wanted]
+
+    if wanted is None:
+        # nothing named a default, which is not an error: it is what a config
+        # file that only defines keymaps, or no config file at all, looks like
+        return {}
+    raise _no_such_profile(wanted, path=names_default)
+
+
+def validate_profile_options(
```
tconbeer: This would be better as "parse_profile_options" I think... the goal should be to return a struct from msgspec. Or am I missing something?

src/harlequin/config.py
```diff
@@ -256,10 +421,44 @@ def refuse(detail: str) -> HarlequinConfigError:
 
 
 def load_config(config_path: Path | None) -> Config:
-    paths = _find_config_files(config_path)
-    config = _merge_config_files(paths)
-    _raise_on_bad_schema(config)
-    return config
+    """Every discovered config file, merged nearest first.
```
tconbeer: I need a little overview of which harlequin or hsql commands call which functions to load their configs. These are subtly different and it's very confusing. I know that, for example, hsql -P foo -c select 1 can return faster than hsql --config --show but it's hard to reason about the correctness and duplication without all of that being laid out explicitly, maybe in a docstring at the top of this file. These function names could be much better also.

src/harlequin/config.py
```diff
@@ -256,10 +421,44 @@ def refuse(detail: str) -> HarlequinConfigError:
...
+    A file merges into the ones behind it one key deeper than it used to: a
+    `profiles` table no longer replaces another file's outright, so a
+    project-local file that defines one profile leaves the rest of them --
+    and the `default_profile` that names one of them -- alone.
+    """
+    return _load(config_path)[0]
+
+
+def _load(config_path: Path | None) -> tuple[Config, Path | None]:
+    """The merged config, and which file named the default profile.
```
tconbeer: This docstring is a nightmare. Be concise.  Don't tell me what the function *doesnt* do

src/harlequin/config.py
```diff
@@ -269,153 +468,346 @@ def get_highest_priority_existing_config_file() -> Path | None:
...
+        return path
     return None
 
 
 def sluggify_option_name(raw: str) -> str:
     return raw.strip("-").replace("-", "_")
 
 
+@dataclass(frozen=True)
+class _ReadFile:
```
tconbeer: Pick a better name. This is also very close to the definition of Config, which is confusing. Do we need both? Can the names more clearly relate them?

src/harlequin/config.py
```diff
@@ -269,153 +468,346 @@ def get_highest_priority_existing_config_file() -> Path | None:
...
+        if raw:
+            yield _validate_file(raw, path=path)
+
+
 def _find_config_files(config_path: Path | None) -> list[Path]:
     """
     Returns a list of candidate config file paths, to be read and
     merged. Returns an empty list if none already exist. Order matters:
-    the last item will have highest priority.
+    the first item will have highest priority.
```
tconbeer: Why duplicate this? Why can't we use one generator to spit off config file paths, whether or not we're stopping early?

src/harlequin/config.py
```diff
@@ -269,153 +468,346 @@ def get_highest_priority_existing_config_file() -> Path | None:
...
+        )
+    return _bad_config(
+        path,
+        f"Config files set the default_profile to {name}, but do not define a "
+        "profile with that name.",
+    )
+
+
+def _validate_file(raw: Mapping[str, Any], *, path: Path) -> _ReadFile:
+    """Everything core owns in one file, checked before it meets another file.
```
tconbeer: Why isn't msgspec used here?

src/harlequin/config.py
```diff
@@ -269,153 +468,346 @@ def get_highest_priority_existing_config_file() -> Path | None:
...
+    values: dict[str, Any] = dict(profile)
+    for key, option in declared.items():
+        value = values.get(key, None)
+        if not isinstance(value, str):
+            continue
+        choices = _declared_choices(option)
+        if not choices or value in choices:
+            continue
+        for choice in choices:
+            if choice.lower() == value.lower():
```
tconbeer: Use casefold not lower

src/harlequin/config.py
```diff
@@ -269,153 +468,346 @@ def get_highest_priority_existing_config_file() -> Path | None:
...
+
+
+def _adapter_profile_model(
+    adapter: str, declared: Mapping[str, AbstractOption]
+) -> type[msgspec.Struct]:
+    """One adapter's declared options, as a model to validate a profile with.
+
+    Every option is a field so that the model describes the adapter (which is
+    what a JSON Schema of it will want), but only a declared set of choices
+    constrains a value: the contract promises values may arrive uncast, so
```
tconbeer: "May arrive uncast" doesn't mean "we won't validate options".

tconbeer (commented):
config.py is a disaster. There is confusing duplication and near-duplication everywhere. msgspec is adding complexity instead of hiding it. Names are terrible and only make it more confusing. The whole module has to be rewritten to make it much more clear and easy to use and maintain.

src/harlequin/config.py
tconbeer: Gone. The list existed because `config.py` was trying to answer "is this key a command's or the adapter's" on its own, which it cannot: the answer is different for each command.

Each command now passes the names it reads for itself, and neither is a new list. `hsql` takes `{param.name for param in cmd.params}` off its own command before the adapter's options are attached, plus `TUI_ONLY_KEYS`; the IDE does the same with `inner_cli.params`, plus a new `HSQL_ONLY_KEYS` derived from the `HSQL_ONLY_OPTIONS` spellings that were already there and already pinned against the real `hsql` command by a test.

---

src/harlequin/config.py
tconbeer: Gone with the thing it was for. Reading a config no longer tracks which file set `default_profile` at all — the error is now "Config files set the default_profile to foo, but no config file defines a profile with that name", which is what is actually true of a merged document and needs no path carried through three functions to say. Naming the file per key is provenance, and that is PR 2's job.

---

src/harlequin/config.py
tconbeer: Both gone, and `held` turned out to be a symptom rather than a naming problem: it existed because the walk kept a side list of files read before anything named a default. Merging into the accumulator as we go makes that list *be* the accumulator, so the whole loop is now:

```python
config: Config = {}
for from_file in _read_config_files(config_path):
    _merge(from_file, into=config)
    name = profile_name or config.get("default_profile", None)
    profiles = config.get("profiles", {})
    if name is not None and name in profiles:
        return profiles[name]  # and the files behind this one go unread
return _select_profile(config, requested=profile_name)
```

---

src/harlequin/config.py
tconbeer: Yes — the options do carry types, and it is now a parse rather than a name check. Every subclass implies one: `FlagOption` → `bool`, `ListOption` → `list[str]`, `SelectOption` → its choices, everything else → `str`. The per-adapter struct is built from those, and the profile is converted into it, so a value reaches the adapter as the type its adapter declared.

Widening happens first, in `_as_declared()`, and does the two roundings click already does for a value typed on the command line:

| written | parsed | why |
| --- | --- | --- |
| `port = 5432` | `"5432"` | your example — a text option, holding the type TOML reaches for |
| `read_only = "true"` | `True` | msgspec's lax mode |
| `mode = "RO"` | `"ro"` | a declared choice, matched by casefold |
| `read_only = "yes"` | refused | not a boolean in any spelling |
| `extension = "httpfs"` | refused, naming the array | it was being iterated one character at a time |
| `port = true` | refused | a flag's value under a text option is a mistake, not a spelling |

The last two are the "some existing config files might break" cases, and both were silently doing something wrong before. Changelog entry says so.

---

src/harlequin/config.py
tconbeer: We can, and now do — msgspec's `forbid_unknown_fields=True` on the per-adapter struct is what refuses the key. The hand-written loop was there for the *message*, and it no longer runs on the happy path: the set difference and `difflib.get_close_matches` are computed in the `except` branch, once we already know the convert failed, so the good message costs nothing when there is nothing wrong.

Same shape at the top level, with one thing msgspec can't do for us: `msgspec.convert(raw, Config)` **drops** a key a TypedDict doesn't declare rather than refusing it (Structs have `forbid_unknown_fields`; TypedDict targets don't), so "Found unexpected key in config: foo" is a three-line check — and it reads the allowed names off `Config.__annotations__`, so there's still only one declaration.

---

src/harlequin/config.py
tconbeer: You're not missing anything — that was the flaw. It built a struct, threw it away, and returned `None`, which is why msgspec looked like it was adding machinery: it validated and then nothing used the result.

It is `parse_profile_options()` now, and it returns the parsed profile. The struct comes back with the values as the adapter declared them, and the coerced values are merged over the profile the caller passes on, so the adapter is handed `"5432"` and `True` rather than `5432` and `"true"`. Both commands now use the return value instead of ignoring it.

(A `Profile` rather than the struct itself: adapters take `**kwargs`, and only the keys the profile actually set may be passed — a struct's unset fields would hand every adapter a `None` for every option it declares.)

---

src/harlequin/config.py
tconbeer: Added, as the first thing in the module:

| caller | function | files read |
| --- | --- | --- |
| `hsql` | `load_profile()` | up to the one that defines the profile |
| `harlequin`, `--keys` | `load_profile_and_keymaps()` | all: keymaps merge across every file |
| the IDE's debug screen | `load_config()` | all |
| `harlequin --config` | `ConfigFile`, at the path `get_highest_priority_...()` returns | one, and unvalidated (it is about to be written back) |

The names now say which is which — `load_config` is the document, `load_profile` is the one profile and is the only one that stops early, `load_profile_and_keymaps` is the IDE's because keymaps are what force it to read everything. `get_config_for_profile` and `get_profile` are gone.

They are also no longer subtly different: both selection paths merge with the same `_merge()` and end at the same `_select_profile()`, so the only difference between them is where they stop. `--config show` (PR 2) will be `load_config()`, which is why it will be slower than `hsql -P foo -c "select 1"` — same code, more files.

---

src/harlequin/config.py
tconbeer: Fair. That function is gone, and I cut the rest of them back — every docstring in the module is now one line unless there is a reason a reader would otherwise get it wrong, and none of them says what the function doesn't do.

---

src/harlequin/config.py
tconbeer: No, we don't need both — `_ReadFile` was `Config` plus a `path`, and the path was only ever used to name a file in an error. It's deleted.

The bigger fix behind it: `Config` is now the msgspec model *and* the TypedDict, so there is one declaration of what a config file may say. `msgspec.convert(raw, Config)` returns the plain dict everything already uses, instead of building a parallel model and throwing it away. That is what made msgspec look like it was adding structure rather than removing it.

The reader yields `Config`s now, and the path stays where it is needed: inside `_read_config_files()`, which is where the error is raised.

---

src/harlequin/config.py
tconbeer: There is one now. `_discover_config_files()` is a generator of paths, and it is what every caller uses — the full load, the profile load that stops early, and `get_highest_priority_existing_config_file()`, which is now just "take the first one that has a section of ours" instead of its own reversed scan. `_read_config_files()` is the one layer on top of it, adding the read and the validation, and it is a generator for the same reason.

The three `_search_home/_search_config/_search_cwd` functions collapsed into one `_search_directories()`, which is also the seam `tests/conftest.py` patches to keep a developer's own config files out of the test suite — one target instead of three, and `--config-path` still goes around it.

---

src/harlequin/config.py
tconbeer: Most of what was there is now msgspec's: the whole shape of a file is `msgspec.convert(raw, Config)`, and the per-profile type checking that function was doing is gone entirely (it duplicated `parse_row_count()`, which already rejects `limit = true` where the value is used).

What is left hand-written in `_check_profile_names()` is three rules msgspec can't express, because they are about names TOML allows and we don't: a profile called `None`, an option written `read-only` instead of `read_only`, and `keymap_names` for `keymap_name`. Each exists to produce a specific "did you mean" that a schema violation can't. The alternative is `Dict[Literal[...], ...]`, which cannot work — profile names and adapter option names are open sets.

---

src/harlequin/config.py
tconbeer: Done — `casefold()` in `_as_declared()`, which is the one place a choice is matched.

---

src/harlequin/config.py
tconbeer: Agreed, and the sentence is gone along with the behavior it was excusing. The contract line means an adapter must not *assume* a value arrives cast — it does not mean nobody may cast it, and we are the layer that knows what the type should be. Options are now parsed into their declared types, and the widening in `_as_declared()` is what keeps `port = 5432` working while `port = true` is refused.

---

src/harlequin/cli.py
tconbeer: It does — `build_cli()` opens with `load_adapter_plugins()`, which `ep.load()`s every installed adapter. Measured with four installed: 309ms, against 114ms for one and 51ms for names-and-versions. I have **not** changed that here, because it is a change to how the IDE's command is built rather than to how config is read, and it changes `harlequin --help`. Happy to do it next, or in this PR if you'd rather — say which.

What it takes, concretely. The IDE needs all adapters for exactly three things: the `--adapter` Choice, the per-adapter rich-click option groups, and the merged set of connection options on the command. `hsql` already solves this with a resilient pre-parse of `-a/-P/--config-path` that names the one adapter and attaches only its options, and `adapter_names()` covers the Choice without importing anything. So:

- move `_preflight()` and `_attach_adapter_options()` somewhere both commands can reach (`harlequin.cli` may not import `harlequin.hsql.cli`), and have the IDE build its command the same two-phase way;
- `--adapter` becomes `click.Choice(adapter_names())`, no imports;
- one option group, for the adapter in play; `AbstractOption.merge()` stops being needed on this path, since collisions only exist between two adapters.

The cost is that `harlequin --help` documents the adapter you are about to use rather than every adapter installed — `harlequin --help -a postgres` for another one, as `hsql` does. That is the part I did not want to decide for you inside a config PR. There is also snapshot/test churn in `test_cli.py`, which asserts on adapter options being present in help.

---

src/harlequin/cli.py
tconbeer: Do we need to maintain this set manually? Can we derive it?

src/harlequin/cli.py
tconbeer: Write this up as an issue. I'd like to make this change but in a separate PR

For --help, we can customize the behavior to still import all adapters, so that won't get a speedily, but I think it's worth it for the more typical path when someone is actually opening the IDE.

src/harlequin/config.py
```diff
@@ -255,167 +352,273 @@ def refuse(detail: str) -> HarlequinConfigError:
...
+        raw = ConfigFile(path).relevant_config
+        if raw:
+            yield _parse_config(raw, path=path)
+
+
+def _parse_config(raw: Mapping[str, Any], *, path: Path) -> Config:
+    """One file's config, or a HarlequinConfigError naming that file."""
+    import msgspec
+
+    # msgspec drops a key `Config` does not declare rather than refusing it, so
```
tconbeer: Why not make Config a msgspec struct with forbid_unknown_keys? Then it will raise

src/harlequin/config.py
```diff
@@ -255,167 +352,273 @@ def refuse(detail: str) -> HarlequinConfigError:
...
+    # this is the one thing the model cannot say for itself
+    if unknown := sorted(raw.keys() - Config.__annotations__.keys()):
```
tconbeer: Why not make Config a msgspec struct with forbid_unknown_keys=True? That way it will raise.

src/harlequin/config.py
```diff
@@ -255,167 +352,273 @@ def refuse(detail: str) -> HarlequinConfigError:
...
-                title="Harlequin couldn't load your config file.",
+def _check_profile_names(profiles: Mapping[str, Profile], *, path: Path) -> None:
+    """Names TOML allows in a profile table and Harlequin does not."""
+    for profile_name, profile in profiles.items():
+        if profile_name == "None":
+            raise _refuse(
+                path,
+                "Config file defines a profile named 'None', which is not allowed.",
+            )
+        for option_name in profile:
```
tconbeer: I don't think we need this check now that the exact schema is validated.

src/harlequin/config.py
```diff
@@ -255,167 +352,273 @@ def refuse(detail: str) -> HarlequinConfigError:
...
-        )
 
-    if (default := config.get("default_profile", None)) is not None and not isinstance(
-        default, str
-    ):
+
+def _merge(from_file: Config, *, into: Config) -> None:
+    """Add what a lower-priority file defines and a higher-priority one did not.
+
+    One key deeper than the `dict.update()` this replaced, which is the whole
```
tconbeer: This part of the docstring is irrelevant to the future and should be removed.

src/harlequin/cli.py
tconbeer: Filed as #1047, with your `--help` decision written into it: it keeps importing everything so its output doesn't change, and the two-phase build is for the path that actually opens the IDE. Nothing in this PR touches it.

---

src/harlequin/config.py
tconbeer: Done — `Config` is a `msgspec.Struct` with `forbid_unknown_fields=True`, and the check beside it is gone. The struct is better than the check in a second way I hadn't weighed: the merge and the selection read attributes instead of `.get("profiles", {})` chains.

```python
def _merge(from_file: Config, *, into: Config) -> None:
    if into.default_profile is None:
        into.default_profile = from_file.default_profile
    for profile_name, profile in from_file.profiles.items():
        into.profiles.setdefault(profile_name, profile)
    for keymap_name, bindings in from_file.keymaps.items():
        into.keymaps.setdefault(keymap_name, bindings)
```

The cost is that msgspec is now imported at module scope, since defining the class needs it. Measured, best of 11 on the real console script: **+13ms on `hsql --help` when no config file is discovered at all** (95.5 → 108.7ms), and within noise of `main` on any invocation that does find one — the files early stopping no longer reads pay for the import. I dropped the import-hygiene test that asserted msgspec stayed out of an empty-config run, since that is no longer true by design.

One thing it costs elsewhere: the debug screen renders the merged config as TOML, so it now calls `Config.to_dict()` (the keys that are set, and nothing else) rather than dumping the mapping directly.

---

src/harlequin/config.py
tconbeer: Right — deleted. The dash rule and the `keymap_names` rule were both checking names the second pass already refuses, and it refuses them better: the suggestion comes from `difflib` against the real key set rather than from one hand-written rule per spelling.

```
Profile defines an option 'read-only', which is not an option of the duckdb adapter. Did you mean 'read_only'?
Profile defines an option 'keymap_names', ...                                        Did you mean 'keymap_name'?
Profile defines an option 'reed_only', ...                                           Did you mean 'read_only'?
```

The `keymap_names` one needed a fix to work: the suggestion was drawn only from the adapter's declared options, so a near miss on one of the *command's* keys got no suggestion. It now draws from both sets.

What is left of `_check_profile_names` is one rule, inlined into `_parse_config`: a profile may not be called `None`, because that is the name a caller passes to mean "none of them".

One consequence worth naming: a dashed key is now caught where the adapter is known, so `load_config()` alone no longer reports it. That is the same trade as everywhere else in this PR — validate what you are about to use — and `--config validate` (PR 3) is what will sweep every profile.

---

src/harlequin/config.py
tconbeer: Cut — `_merge`'s docstring is one line now, about what it does rather than what it replaced. I went through the module for the same thing; nothing else was narrating history.

---

src/harlequin/cli.py
tconbeer: Derived, and the set is gone. `hsql.cli` grew one function:

```python
def bare_command() -> click.Command:
    """This command with no adapter's options on it, and no config file read."""
    return build_cli(["-P", "None", "--help"])
```

`-P None` is what keeps it cheap and correct: no config file is read, so no profile can name an adapter, and `--help` is what stops the fallback to the default one — 9.5ms in a process that has already imported the IDE, and no adapter imported, so the surface it reports is hsql's own rather than whatever happens to be installed.

The IDE reads two things off it, in `harlequin/cli.py`:

- `hsql_spellings()` — for the "did you mean `hsql -c`?" hint. And the *difference* between the two commands turned out to be unnecessary: click has already established the spelling is not ours before raising `NoSuchOption`, so all that is left to ask is whether hsql has it.
- `hsql_profile_keys()` — `param.name` for each, which is exactly the profile keys hsql reads, with no `sluggify` guessing from spellings. This is what `parse_profile_options()` needs so the IDE doesn't mistake `format = "csv"` in a shared profile for an adapter option.

Both are lazy behind an `lru_cache`, so `--help` and `--version` never build it. `test_the_hsql_option_list_is_hsqls` (which pinned the copy) is replaced by one asserting the derivation is hsql's own surface and not an adapter's — the property that would break it.

---

src/harlequin/config.py
```diff
@@ -46,26 +96,31 @@
...
+    """A binding's own keys are `HarlequinKeyMap.from_config()`'s to check."""
 
-class Config(TypedDict, total=False):
-    default_profile: str | None
-    keymaps: dict[str, list[RawKeyBinding]]
-    profiles: dict[str, Profile]
+    def to_dict(self) -> dict[str, Any]:
+        """What this config would look like written back to a file."""
+        return {
+            key: value for key, value in msgspec.structs.asdict(self).items() if value
```
tconbeer: `value` could be falsy but is still meaningful. fix this guard.

tconbeer: Fixed. It now drops `None` alone, which is the one thing TOML has no way to write:

```python
    def to_dict(self) -> dict[str, Any]:
        """What this config would look like written back to a file."""
        return {
            key: value
            for key, value in msgspec.structs.asdict(self).items()
            # the one thing TOML cannot write. Everything else is kept, however
            # falsy: `default_profile = ""` is a mistake worth being able to see.
            if value is not None
        }
```

`default_profile = ""` was exactly the case being swallowed, and the debug screen is where someone would go looking for it. An empty `profiles` or `keymaps` now renders as a bare `[profiles]` header instead of vanishing, which reads as "nothing defined these" rather than implying there is nothing to say.

I went back through the guards I added for the same mistake; the other truthiness tests are all on mappings where empty and absent are the same thing (`if raw:` for a file with no section of ours, `if not given:` for a profile with no adapter options in it, `if not choices:` for an option that declares none). `_merge` and `_select_profile` were already testing `is None`.

Also rebased onto `origin/main` (a1ae510).

---

PR #1052, opened 2026-08-18
feat: view a result cell's full value in a modal

### LINKED ISSUE #1011: Add data table interaction for viewing a cell's contents in a new modal/editor
> Would be awesome if a double click or something would open up a modal to display all of the data, with a scrollbar!
> 
> Although, I realize this is a new feature not directly related to this issue :) 

 _Originally posted by @rsullenb in [#894](https://github.com/tconbeer/harlequin/issues/894#issuecomment-3499682693)_

ChrisJr404:
Closes #1011.

Adds a modal that shows the full value of the focused Results Viewer cell, so a long string or a blob of JSON that gets clipped in the grid is actually readable. Press `space` on a cell and you get a scrollable pane with the whole value; arrows/PgUp/PgDn scroll it, `c` copies it, and esc/enter/q close.

Wired it the normal way so it stays rebindable: a `results_viewer.view_cell` entry in `HARLEQUIN_ACTIONS`, an `action_view_cell` on `ResultsTable`, and a default `space` binding in the vscode keymap.

I went with a keybinding rather than the double-click the issue floated. #987 is already using double-click for in-place edits, and a key works fine when you're driving from the keyboard. Can add double-click too if you'd prefer.

One thing this surfaced: `ErrorModal` dismisses on any key but never stopped the event, so the keypress could fall through and trigger a binding on the table underneath (dismiss an error with `space`, and you'd immediately pop this modal open). Made the error modal consume the key.

For tests I added `test_view_cell_modal`: runs a query with a long string, focuses the table, presses `space`, checks the modal is up and shows the cell value in full, then esc pops it off. Plain pilot assertions, no new snapshot. Ran the results-viewer, keymap, help-screen and app suites locally and they're green; the error-modal fix is covered by the existing `test_query_errors` cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
   - `list-profiles` lists every profile you can pass to `-P`, its adapter, and which one is the default. It is rows, so `--csv`, `-o` and `-t`/`-A` apply.
   - `show` prints the merged config with the file each value came from, and the files it overrode — so you can see which file is winning. `--json` for JSON.
   - `validate` reports every problem in every discovered config file — the file, the key, what is wrong, and the line where the TOML parser knew one — instead of stopping at the first, as a run does. It checks each profile against the options its adapter declares, so a misspelled `reed_only` is reported rather than dropped in silence, and it checks your keymaps too. It exits `2` when it found anything, so `hsql --config validate --format none` is the whole check for a script. It is rows, so `--csv`, `-o` and `-t`/`-A` apply.
+- The Results Viewer can open the value under the cursor in a scrollable modal, so a long string or a blob of JSON that gets clipped in the grid can be read in full (press `space`, or copy it out with `c`) ([#1011](https://github.com/tconbeer/harlequin/issues/1011)).
 
 ### Breaking Changes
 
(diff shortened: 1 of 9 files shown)


tconbeer:
This is good, thanks so much for this PR

image

I don't love the styling of the modal. I'd rather it match the Help and Error modals, with the primary color as the outer border, and a faded color as the inner border:
image

Also, a click outside the modal should close it (as it does with the Help and error modals).

So I would say that you could probably be reusing more components to bring the styling and behavior of this new modal in line with the existing ones.

ChrisJr404:
Good call, that's much cleaner. I reworked the modal to reuse the shared `#modal_outer`/`#modal_inner` containers, so it now gets the same primary outer border and faded inner border as the Help and Error modals instead of its own bespoke styling. Dropped the custom `#cell_view_*` rules entirely.

It also closes on a click outside the box now, matching the Help/Error behavior. I wrapped the outer container in the same `VerticalSuppressClicks` the Help screen uses so clicks inside the modal (scrollbar, text) don't fall through and dismiss it. Extended the existing test to cover the click-to-close path.

Local suite is green and ruff is clean; no snapshots changed.

tconbeer:
Couldn't push to this branch, so made some tweaks and will ship in #1062. Thanks for your contribution!

PR #1055, opened 2026-08-18
Add `AbstractOption.to_dict()` and `hsql --spec`

tconbeer:
Implements PR4 of M2 of the hsql plan.

1. **`AbstractOption.to_dict()`** — an option as plain data: `name`, `type`, `label`, `description`, `short_decls`, `default`, `choices`, `multiple`. Concrete rather than abstract, since `AbstractOption` is public API that third-party adapters subclass; a subclass that predates the method still answers, and a new option type only sets the new `option_type` class attribute to name itself. The keys are the same for every type — one that does not apply is null rather than missing.
2. **`hsql --spec`** — a new mode option that dumps the installed CLI surface as JSON and exits: hsql's own options and arguments, plus every installed adapter's connection options keyed by adapter name. `-a NAME` narrows it to one adapter.
3. **`harlequin.plugins.adapter_versions()`** — the distribution version behind each installed adapter, importing none of them.
4. **Mode mutual exclusion** — `--spec` and `--config MODE` together is a usage error naming both, and the "a mode does not run SQL" refusal moved out of `_report_config` so both modes share it.
5. **`reserved_spellings()`** — the rule `_attach_adapter_options` applies when an adapter option collides with one of hsql's own flags, extracted so `--spec` reads it instead of keeping a second copy.
6. The IDE's debug screen now reads `to_dict()` in place of duck-typing at `opt.name`, `opt.short_decls` and `getattr(opt, "default", None)`.

**One vocabulary over two sources.** hsql's own options are click parameters and an adapter's are `AbstractOption`s, but a document where half the entries said `help` and the other half said `description` would make a reader learn which half they were in. Both are reported under the CLI's terms — `name` is the parameter name (and so its key in a profile), `decls` is every spelling, `default` is what the command uses when the flag is absent.

**The document reports what a caller can type, not what an adapter declared.** `_attach_adapter_options` already drops a spelling that collides with an hsql flag, and drops an option whose name is taken. A spec that advertised `-c` for an adapter option that will never receive it would send a caller to write something that means `--command`. Rather than copy that rule, I pulled it into `reserved_spellings()` and had both callers read it. Same reasoning for a flag's `default`: `FlagOption`'s own docstring says a non-`False` default is GUI-only, so the spec reports `false` and `to_dict()` keeps reporting the declaration faithfully.

**An adapter that will not import is a fact about the installation, not a reason to answer nothing.** It appears in the document with `options: null` and the error, plus a line on stderr; the other adapters are still reported and the exit code is still 0.

**`bare_command()` rather than the running command** as the source for hsql's own options, so `hsql --spec` and `hsql --spec -a duckdb` differ only in the half they were asked to differ in. Read through `get_params()` rather than `.params`, because click keeps `--help` out of the latter and a caller reading this to learn the surface should be told about the flag that would have shown it to them.

**Tradeoffs.** `--spec` imports every installed adapter — ~380ms with four of them here — which is the cost of reading an `ADAPTER_OPTIONS` off a class. That is the right trade for a once-per-task lookup and the wrong one for the query path, which is why the mode lives in its own module that the callback imports only when the flag is passed. It opens no connection and reads no config file, so it answers when the database is unreachable or the config is broken.

**One correction to the plan.** §3 of the M2 plan says "`--spec` and `--config` import no database driver". That cannot hold for `--spec`: reading an adapter's options means importing the adapter, which imports its driver — and §3.1's own table and §1.4's measurement both say as much. The guard shipped here is the one that is true: `--spec` opens no connection, and under `-a` it imports exactly one adapter.

- [ ] No.
- [x] Yes; I haven't opened a PR, but the gist of the change is: the Headless & Agents topic needs a `--spec` section (what the document holds, `-a` to narrow it, and that it neither connects nor reads config), and the adapter-author page needs a note on `to_dict()` and `option_type`. The M2 plan puts both in PR 8 of Release A, alongside the `--config` docs.

- [x] Yes.

42 new tests:
- `tests/unit_tests/test_options.py` (new) — the serialization's shape across every option type, a `PrehistoricOption` that implements only the three renderings the ABC demanded before `to_dict()` existed, a subclass inheriting its parent's `option_type`, and every option both bundled adapters declare.
- `tests/unit_tests/test_hsql.py` — the document's contents (defaults, choices, spellings, repeatable options, envvars, the positional's `nargs`), that click's "no default" sentinel never reaches JSON, that a colliding adapter spelling is dropped from the report, `-a` narrowing, two modes as a usage error naming both, `-o`/`--format none`/an unreachable format, and that it answers over an unreadable config file and a profile that is not there.
- `tests/unit_tests/test_import_hygiene.py` — a subprocess proving `hsql --spec -a sqlite` imports `harlequin_sqlite` and no `duckdb`, `pyarrow` or `tomlkit`.

Full `make check` equivalent is green: ruff format/check, mypy over 109 files, `lint-imports` (4 contracts kept), 771 passed on 3.10 and the 3 py12 tests on 3.12.

- [x] I have added an entry to `CHANGELOG.md`, under the `[Unreleased]` section heading.

https://claude.ai/code/session_01LDd4r9U41wKsxXsH9gaNt7

---
_Generated by [Claude Code](https://claude.ai/code/session_01LDd4r9U41wKsxXsH9gaNt7)_


src/harlequin/components/debug_info.py
```diff
@@ -154,11 +154,12 @@ def parse_info(self) -> List[DebugWidget]:
...
             table = ["| Flag(s) | Value |", "|---|---|"]
             for opt in self.adapter_options:
-                flags = f"--{opt.name}"
-                if getattr(opt, "short_decls", []):
-                    flags += " " + " ".join(opt.short_decls)
-                value = getattr(opt, "default", None)
-                table.append(f"| `{flags}` | `{value}` |")
+                # to_dict() rather than reaching for attributes an option type
+                # may not have: `default` belongs to some of them and not
+                # others, and the option is the only thing that knows which.
```
tconbeer: don't comment on code you're removing.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file.
...
+- Adds `hsql --spec`, which dumps hsql's own options and every installed adapter's connection options as JSON, so an agent can read the CLI surface instead of parsing `--help` ([#524](https://github.com/tconbeer/harlequin/issues/524)). Every option reports each way to spell it, what it takes, the values a choice accepts, its default, and whether it is repeatable — and an adapter option reports both spellings it has, `--read-only` for the command line and `read_only` for a profile. `-a NAME` narrows it to one adapter. It connects to nothing and reads no config file, so it answers even when your database is unreachable or your config is broken. The `harlequin` command's own options are not covered, and the document says so instead of reading as an exhaustive list.
```
tconbeer: Both of these entries are too long. Shorten by half at least.

src/harlequin/hsql/modes/spec.py
```diff
@@ -0,0 +1,280 @@
...
+
+    command = _hsql_command()
+    # `get_params` rather than `params`, because click keeps `--help` out of the
+    # latter and adds it at parse time -- and a caller reading this to learn the
+    # surface should be told about the flag that would have shown it to them
+    params = command.get_params(click.Context(command, info_name="hsql"))
+    document = {
+        "program": "hsql",
+        "version": version("harlequin"),
+        "covers": COVERS,
```
tconbeer: Let's call this "scope" instead of "covers"

src/harlequin/components/debug_info.py
tconbeer: Removed — that belonged in the PR description, not beside the code. `declared = opt.to_dict()` now stands on its own.

---

CHANGELOG.md
tconbeer: Cut both to a line each — roughly a quarter of what they were:

- Adds `hsql --spec`, a machine-readable `--help`: hsql's options and every installed adapter's connection options, as JSON. `-a NAME` narrows it to one adapter (#524).
- Adds `AbstractOption.to_dict()` to the adapter API, which serializes an option as plain data.

---

src/harlequin/hsql/modes/spec.py
tconbeer: Renamed — the JSON key is `"scope"`, the constant behind it is `SCOPE`, and the module docstring and test follow.

---

PR #1096, opened 2026-08-30
fix: don't split statements inside dollar-quoted bodies (#1019)

### LINKED ISSUE #1019: Query splitting breaks inside a dollar-quoted body (for some statements)

visheshgubrani:
Closes #1019 (an existing open issue)

This fixes statement splitting inside dollar-quoted function bodies. Semicolons inside `$$ ... $$` were being treated as query separators, which caused one function definition to be split into multiple invalid statements.

The splitter now captures dollar-quote delimiters along with semicolons. It pairs matching delimiters and ignores semicolons found inside the quoted body.

The Query Editor and scripts share the same splitter, so the fix applies to both. Tests cover the original issue, labeled delimiters, delimiter boundary cases, unmatched delimiters, multiple bodies, Unicode text, and the Query Editor path.

Using the delimiters reported by tree-sitter avoids adding a separate SQL lexer just for dollar quoting. This also keeps the existing handling of strings, comments, and quoted identifiers unchanged.

Unmatched delimiters do not hide any semicolons, so invalid or incomplete SQL keeps its existing recovery behaviour. The change is limited to dollar-quoted bodies. Other function body formats are unchanged.

* [x] No.
* [ ] Yes, but I have not opened a PR yet. The required change is: ...

* [x] Yes.
* [ ] No, I believe tests are not necessary.

* [x] I have added an entry to `CHANGELOG.md` under the `[Unreleased]` section. That entry references the issue closed by this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- A semicolon inside a dollar-quoted function body (e.g. `create function f() as $$ select 1; $$`) is no longer treated as a statement separator, in the Query Editor or in scripts ([#1019](https://github.com/tconbeer/harlequin/issues/1019)).
+
 ## [2.12.0] - 2026-08-29
 
 ### Features
--- a/src/harlequin/statements.py
+++ b/src/harlequin/statements.py
@@ -1,9 +1,11 @@
 """Locating statement boundaries in SQL text.
 
 Both front ends split a script here, through the same tree-sitter grammar and
-the same one-line query the Query Editor has always used, so a script run with
-`-f` and the same script in the editor cannot disagree about where a statement
-ends.
+the same query the Query Editor has always used, so a script run with `-f` and
+the same script in the editor cannot disagree about where a statement ends. A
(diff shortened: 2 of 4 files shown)


tconbeer (approved):
Thank you! Nice solution: simpler than I had assumed.

visheshgubrani:
Thanks, Ted! Glad the approach worked out.

PR #1107, opened 2026-09-01
fix: config wizard no longer prefills bad defaults (#1105)

### LINKED ISSUE #1105: Config wizard is too eager to fill in bad defaults
Config wizard fills in defaults in a few places it shouldn't:
- limit gets `-1`; it should be blank. Help text says "Enter -1 for no limit", but Help should say "Leave blank for app defaults; enter -1 for no limit".
- For adapter options (e.g., postgres host, user, etc.) the word "None" is auto-filled, when it should just be blank.
- For adapter secret options (e.g., postgres password), four asterisks ("****") are auto-filled (presumably this is also "None"), but this should also start blank.

visheshgubrani:
The config wizard prefilled `-1` for an unset query limit and the literal string `"None"` for unset adapter options. For secrets, `"None"` appeared as `****` and could be written into the profile if accepted.

This change keeps unset values unset, while preserving declared defaults and existing profile values. The limit prompt now starts blank and explains how to use the app default or explicitly request no limit.

Closes #1105

* Treat `None` as an absent prompt value in `TextOption`, `PathOption`, and `SelectOption`, rather than converting it to the string `"None"`.
* Allow the query limit prompt to be left blank and omit the `limit` key in that case.
* Skip blank adapter-option answers when writing the profile, including clearing existing values.
* Compare blank answers with `value == ""` so an explicit `False` from a `FlagOption` is still written.
* Preserve declared option defaults and existing profile values.
* Add regression coverage for limit handling, text/path/select defaults, secrets, lists, cleared values, and explicit false flags.

The `"None"` value originated in Harlequin before questionary received it, so the conversion is fixed in the option classes rather than worked around in questionary or only in the wizard. A shared helper keeps the three affected option types consistent.

Blank answers are omitted at the wizard's write boundary so empty strings and `[""]` do not become configuration values. The check is intentionally limited to `""` a general truthiness check would incorrectly discard an explicit `False` from a flag option.

Existing values, declared defaults, `viewer_max_rows`, and non-wizard config paths are unchanged.

* [x] No.
* [ ] Yes; I haven't opened a PR, but the gist of the change is: ...

* [x] Yes.
* [ ] No, I believe tests aren't necessary.

Testing performed:

* 57 config-wizard and option unit tests passed.
* `mypy` passed.
* Ruff formatting and lint checks passed.
* Import-linter contracts passed.

* [x] I have added an entry to `CHANGELOG.md`, under the `[Unreleased]` section heading. That entry references the issue closed by this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- The config wizard no longer pre-fills the unset query limit with `-1` or unset adapter options with `None` (`****` for secrets). Leave the limit blank for app defaults, or enter `-1` for no limit ([#1105](https://github.com/tconbeer/harlequin/issues/1105)).
+
 ## [2.12.2] - 2026-08-31
 
 ### Bug Fixes
--- a/src/harlequin/config_wizard.py
+++ b/src/harlequin/config_wizard.py
@@ -98,15 +98,14 @@ def _wizard(config_path: Path | None) -> None:
 
     # two questions, because they are two limits: what leaves the database,
     # and what the Results Viewer holds of it.
-    limit = int(
-        questionary.text(
-            message="How many rows should each query fetch from the database?",
-            instruction="Enter -1 for no limit.",
-            validate=_validate_int,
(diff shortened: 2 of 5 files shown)


tconbeer (approved):
awesome, thanks!

Issue #545, opened 2024-05-04
SSH tunneling


tconbeer:
Need to think through whether this is an adapter feature or a harlequin feature. Maybe it's a library that adapters can use. 

Would be nice to provide CLI options for configuring an ssh tunnel, and then intercept/rewrite the host/port/etc options passed through to the adapter. 

See:

https://stackoverflow.com/questions/22046708/psycopg2-access-postgresql-database-on-remote-host-without-manually-opening-ssh

golgor:
One similar topic that would be nice is to add support for Cloud SQL. This is used for any database hosted in google Cloud for example, maybe other providers as well.

I have some working examples, I can try to find it and add some info here. Maybe also try to implement something if I find the time.

tconbeer:
A major hurdle for me on this one is that the packages I would use don't seem very well-maintained. e.g., https://github.com/pahaz/sshtunnel/ looks perfect, but hasn't seen a release in 4 years.

golgor:
The package for Cloud SQL connection is well maintained by Google: https://github.com/GoogleCloudPlatform/cloud-sql-python-connector

I implemented my solution with SQL Alchemy, I haven't checked the code for Harlequin that much yet, so not sure if the same approach would work for you.

Bascially you create a "Connector" using the Google Cloud SQL package, and then pass that into the create_engine of SQL Alchemy.

The example below would give you a better idea of how it works.

```python
@dataclass
class DatabaseConnectionSpec:
    """A dataclass to hold the specification for a database connection."""

    user: str
    password: str
    database: str
    gcloud_project_str: str


class MysqlSessionContext:
    """Class to handle a session to a MySQL database."""

    def __init__(self: Self, connection_spec: DatabaseConnectionSpec) -> None:
        """Initialize the object and setup the database connection.

        The connection is setup using cloud_sql_proxy to connect to the database. The user

        Args:
            connection_spec (DatabaseConnectionSpec): The specification for the database connection.
        """
        self.spec = connection_spec

        self.engine = create_engine("mysql+pymysql://", creator=self._get_connector, echo=False)
        self.session = Session(bind=self.engine)

    def __enter__(self: Self) -> Session:
        """Enter the context manager and return the session.

        Returns:
            Session: A session to use for database operations.
        """
        return self.session

    def __exit__(self: Self, exc_type: type | None, exc_val: Exception | None, exc_tb: TracebackType | None) -> None:
        """Exit the context manager and close the session.

        Args:
[...]

PR #1112, opened 2026-09-01
SSH tunnels 1/5: the ssh child that holds a local forward open

tconbeer:
Part 1 of 5 of `docs/plans/ssh-tunnels-design.md`, phased as §7 asks. Nothing here is reachable from either command yet; 2/5 adds the flags.

Stack: **1/5 (this)** → 2/5 options → 3/5 keepalive & reporting → 4/5 reconnect → 5/5 changelog.

Part of #545

- `harlequin/ssh.py`, one class. `SshTunnel` runs `ssh -N` as a child process and `start()` blocks until the local ends of its forwards accept connections; `stop()` terminates and then kills it, from a context manager and an `atexit` backstop.
- `build_argv()` — the whole of what Harlequin says to `ssh`: `-N`, `-o ExitOnForwardFailure=yes`, one `-L` per forward, and the destination. Nothing a user typed is parsed.
- `resolve_config()` / `parse_config()` — the `ssh -G` probe, run with the argv we are about to run, and its output. It is the only way to know which local port to wait on, since the motivating case puts the forward in a `Host` block where Harlequin never sees a port number. Two whitespace-separated fields per `localforward`, each a port, a `[host]:port` or a socket path.
- `HarlequinSshError`, and `harlequin.ssh` joins the "adapter API is reachable without the TUI" import-linter contract.

The design doc has the alternatives (§8 — rewriting the adapter's host/port, `sshtunnel`, paramiko, SOCKS); the notable choices inside this module:

- **A child, not `ssh -f`.** A forked `ssh` signals readiness by exiting, but the surviving process is not ours to kill — which is the orphaned tunnel this feature removes. Staying in the foreground also keeps the terminal, so passphrase and 2FA prompts still work.
- **stderr is captured, drained by a thread, and quoted verbatim in errors.** `ssh` writes its prompts to `/dev/tty` rather than to stderr, so capturing costs no interactivity, and its diagnostics are better than any we would write.
- **A port answering is not proof our forward is up.** The readiness loop records which ports were already bound *before* the child started; when none was, it returns as soon as they all answer, and when some were, it waits out a one-second settle window for `ExitOnForwardFailure` to kill the child. That is what makes the reuse test behavioral (`allow_reuse` — the child failed, but the ports answer) rather than a match against ssh's wording. Half-open — some ports answering — fails either way.
- **One guard beyond the design.** A `--ssh-host` or `--ssh-forward` value starting with a dash is refused: `ssh` has no `--`, so such a value reaches it as a flag, and config files are discovered in the working directory. That is the same threat that got `--ssh-option` rejected in §8; a cloned repo's `pyproject.toml` must not be able to smuggle in `-oProxyCommand=`.
- **`ssh -G` output it cannot read degrades** to no poll, no forwards-nothing error and a short grace period, rather than to a refusal. A `hostname` line is what tells `-G`'s output from anything else.

- [x] Yes, and it comes with 5/5 of this stack (the user-facing docs go to [tconbeer/harlequin-web](https://github.com/tconbeer/harlequin-web); nothing in this PR is user-visible on its own).

- [x] Yes.

`tests/unit_tests/test_ssh.py`, and a fake `ssh` at `tests/data/unit_tests/ssh/ssh` that understands `-G`, `-V` and `-L` and binds the local end of each forward. No SSH server and no `online` marker. Lifecycle tests run it through `sys.executable` so they cover all three OSes; the ones that need it on `PATH` are skipped on Windows. One test asks a real `ssh` whether it accepts the argv, and skips if none is installed.

- [x] I have added an entry to `CHANGELOG.md` — in 5/5 of this stack, where the feature becomes user-visible.

---

### DIFF
--- a/src/harlequin/ssh.py
+++ b/src/harlequin/ssh.py
@@ -339,7 +339,12 @@ class SshTunnel:
 
     def _child_failed(self) -> HarlequinSshError:
         assert self._process is not None
-        detail = self._stderr.text() if self._stderr is not None else ""
+        detail = ""
+        if self._stderr is not None:
+            # the child has exited, but what it wrote on the way out may still
+            # be in the pipe; an error that quotes ssh has to wait for it
+            self._stderr.settle()
+            detail = self._stderr.text()
         return HarlequinSshError(
             f"ssh exited with code {self._process.returncode} without opening "
             f"the forward.\n\n{detail}".rstrip(),
@@ -369,6 +374,16 @@ class _Stderr:
 
     def text(self) -> str:
         return b"".join(self._chunks).decode("utf-8", errors="replace").strip()
+
+    def settle(self) -> None:
+        """Wait for the drain to reach the end of what the child wrote.
+
+        `Popen.wait()` returns when the child exits, which is before the pipe it
+        wrote to has been read to the end -- so a caller that quotes ssh calls
+        this first, or quotes nothing. Only useful once the child is gone;
+        while it runs there is no end to wait for.
+        """
+        self._thread.join(timeout=_TERMINATE_SECONDS)
 
     def close(self) -> None:
         self._thread.join(timeout=_TERMINATE_SECONDS)
--- a/src/harlequin/ssh.py
+++ b/src/harlequin/ssh.py
@@ -252,7 +252,8 @@ class SshTunnel:
         "to authenticate, visit ..." has to appear to be worth anything.
         """
         self._process: subprocess.Popen[bytes] | None = None
-        self._stderr: _Stderr | None = None
+        self._stderr: _Output | None = None
+        self._stdout: _Output | None = None
 
     @property
     def endpoints(self) -> tuple[tuple[str, int], ...]:
@@ -291,7 +292,9 @@ class SshTunnel:
                 argv,
                 # ssh's prompts go to the terminal it inherits, not to this
                 # pipe, so capturing stderr does not swallow them
-                stdout=subprocess.DEVNULL,
+                # both, because ssh is not the only program writing here: a
+                # `ProxyCommand` helper inherits these and picks its own
+                stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE,
             )
         except OSError as e:
@@ -299,7 +302,8 @@ class SshTunnel:
                 f"Harlequin could not run {argv[0]}: {e}", title=ERROR_TITLE
             ) from e
         self._process = process
-        self._stderr = _Stderr(process.stderr, echo=self.echo_stderr)
+        self._stderr = _Output(process.stderr, echo=self.echo_stderr)
+        self._stdout = _Output(process.stdout, echo=self.echo_stderr)
         atexit.register(self.stop)
         try:
             self._wait_until_ready(already_bound=already_bound)
@@ -321,8 +325,9 @@ class SshTunnel:
                 process.kill()
                 with contextlib.suppress(subprocess.TimeoutExpired):
                     process.wait(timeout=_TERMINATE_SECONDS)
-        if self._stderr is not None:
-            self._stderr.close()
+        for stream in (self._stderr, self._stdout):
+            if stream is not None:
+                stream.close()
 
     def __enter__(self) -> SshTunnel:
         self.start()
@@ -373,8 +378,9 @@ class SshTunnel:
         assert self._process is not None
         # the child has exited, but what it wrote on the way out may still be
         # in the pipe -- and it is the whole of why this failed
-        if self._stderr is not None:
-            self._stderr.settle()
+        for stream in (self._stderr, self._stdout):
+            if stream is not None:
+                stream.settle()
         return HarlequinSshError(
             f"ssh exited with code {self._process.returncode} without opening "
             f"the forward.\n\n{self._unseen()}".rstrip(),
@@ -384,9 +390,10 @@ class SshTunnel:
     def _timed_out(
         self, *, already_bound: Collection[tuple[str, int]] = ()
     ) -> HarlequinSshError:
-        if self._stderr is not None:
-            # a helper's last line may have no newline of its own
-            self._stderr.flush()
+        for stream in (self._stderr, self._stdout):
+            if stream is not None:
+                # a helper's last line may have no newline of its own
+                stream.flush()
         if already_bound:
             listed = ", ".join(f"{host}:{port}" for host, port in sorted(already_bound))
             reason = (
@@ -413,19 +420,28 @@ class SshTunnel:
         Nothing, where it was echoed as it arrived: an error that repeated it
         would print a helper's instructions twice.
         """
-        if self._stderr is None or self.echo_stderr:
+        if self.echo_stderr:
             return ""
-        return self._stderr.text()
+        said = [
+            stream.text()
+            for stream in (self._stderr, self._stdout)
+            if stream is not None and stream.text()
+        ]
+        return "\n".join(said)
 
 
-class _Stderr:
-    """What ssh wrote to stderr: passed through as it arrives, and kept.
+class _Output:
+    """One of ssh's streams: passed through to *our stderr* as it arrives, and kept.
 
     Passed through because the thing a user most needs to see arrives while the
     tunnel is still opening and nothing has failed yet -- Tailscale's "to
-    authenticate, visit ..." is written here, and a helper's instructions are
+    authenticate, visit ..." among them -- and a helper's instructions are
     worthless after the timeout they caused. `echo=False` is for a caller that
     no longer owns the screen; it keeps the text for an error to quote instead.
+
+    Always onto stderr, whichever of ssh's streams this is: `hsql`'s stdout
+    carries result sets and nothing else, and a helper that writes to ssh's
+    stdout must not be able to corrupt a caller's csv.
 
     A line at a time, so `redact_text()` sees whole lines: ssh is third-party
     output, and a driver's DSN can reach it.
--- a/src/harlequin/ssh.py
+++ b/src/harlequin/ssh.py
@@ -2,23 +2,24 @@
 
 Harlequin runs `ssh` and touches nothing else. The connection details already
 name the local end of a forward -- `host = "localhost"`, `port = 15439` -- so no
-adapter is told a tunnel exists, and every adapter works, including the ones
-nobody in this org maintains. What lives here is the child's lifetime: build the
-argv, ask `ssh -G` what it is about to forward, wait for those ports to answer,
-and kill it on the way out.
+adapter is told a tunnel exists. What lives here is the child's lifetime: build
+the argv, ask `ssh -G` what it is about to forward, wait for those ports to
+answer, and kill it on the way out.
 
-Nothing a user typed is parsed. `ssh` owns the syntax of a destination and of a
-forward spec, and its error messages are better than any we would write, so both
-values reach it verbatim and its stderr is what an error quotes. The one thing
-this module reads is `ssh -G`'s own output (`parse_config()`), which says which
-local port to wait on -- the motivating case puts the forward in `~/.ssh/config`,
-where Harlequin never sees a port number at all.
+Nothing a user typed is parsed: `ssh` owns the syntax of a destination and of a
+forward spec, so both values reach it verbatim and its stderr is what an error
+quotes. They are checked for the characters ssh would read as an option or
+expand into a shell (`_refuse_unsafe_value()`), because a config file is
+discovered in the working directory. The one thing this module reads is `ssh
+-G`'s own output (`parse_config()`), which says which local port to wait on: a
+forward declared in `~/.ssh/config` never reaches Harlequin as a port number.
 """
 
 from __future__ import annotations
 
 import atexit
 import contextlib
+import os
 import re
 import socket
 import subprocess
@@ -26,12 +27,12 @@ import threading
 import time
 from collections import deque
 from dataclasses import dataclass
-from typing import Any, Sequence
+from typing import Any, Collection, Sequence
 
 from harlequin.exception import HarlequinSshError
 
 SSH = "ssh"
-"""The client, found on PATH: whichever one the user's `~/.ssh/config` is for."""
+"""The client, found on PATH."""
 
 DEFAULT_TIMEOUT = 10.0
 """Seconds to wait for the forwards, when nothing says otherwise."""
@@ -42,25 +43,36 @@ _POLL_INTERVAL = 0.05
 _CONNECT_TIMEOUT = 0.5
 """How long one probe of one local port may take."""
 
-_SETTLE_SECONDS = 1.0
-"""How long a port that was already bound gets to prove it was ssh's to take.
-
-Without this, a listener that answered before `ssh` even started would read as
-the forward being up -- the one case where a port answering proves nothing.
-"""
-
 _GRACE_SECONDS = 1.0
-"""What a tunnel with no port to poll waits instead, before trusting the child."""
+"""What a tunnel with no port to poll waits, before trusting the child."""
 
 _TERMINATE_SECONDS = 2.0
 """How long a terminated child has to exit before it is killed."""
 
+_STDERR_SETTLE_SECONDS = 0.5
+"""How long a caller quoting ssh waits for the drain to catch up.
+
+Bounded rather than open-ended: the pipe reaches EOF only when every write end
+is closed, and a `ProxyJump` helper or an askpass inherits ssh's stderr and
+holds one open after ssh itself is gone.
+"""
+
 _STDERR_LIMIT = 8192
 """Bytes of ssh's stderr kept for an error to quote."""
 
-_LOOPBACK = "127.0.0.1"
-_WILDCARD_HOSTS = {"", "*", "0.0.0.0", "localhost"}
+_ANY_INTERFACE = {"", "*", "0.0.0.0", "::"}
+"""Bind addresses that are not themselves an address to connect to."""
+
 _KEYWORD = re.compile(r"^([a-z][a-z0-9]*)(?:\s+(.*))?$")
+
+_UNSAFE_IN_VALUE = re.compile(r"""[\x00-\x20\x7f`$\\"'|&;<>(){}]""")
+"""Characters no destination or forward spec has, and that ssh may expand.
+
+A destination reaches the user's own `ssh_config` as `%h`, `%r` and `%p`, which
+a `ProxyCommand` there runs through a shell (CVE-2023-51385, CVE-2025-61984).
+Since a config file is discovered in the working directory, the value can come
+from a cloned repository's `pyproject.toml` rather than from its user.
+"""
 
 
 @dataclass(frozen=True)
@@ -68,8 +80,7 @@ class Forward:
     """One local forward, as `ssh -G` printed it: what listens, where it goes.
 
     Two whitespace-separated fields, each a port, a `[host]:port`, or a socket
-    path. Kept as text because that is all a notice needs; only the listening
-    side is ever read for a number.
+    path. Only the listening side is ever read for a number.
     """
 
     listen: str
@@ -90,7 +101,7 @@ class SshConfig:
 
     One list whether a forward came from `--ssh-forward`, a `Host` block, or
     both: `-G` echoes the command line's `-L` flags back along with the config
-    file's, so there are never two sources to keep in agreement.
+    file's.
     """
 
     forwards: tuple[Forward, ...]
@@ -106,14 +117,14 @@ def build_argv(
 
     `ExitOnForwardFailure` is the only option Harlequin imposes on its own: a
     forward that silently did not happen is the one failure a user cannot
-    diagnose. Notably not imposed is `ServerAliveInterval`, which a command-line
-    `-o` would take away from a `Host` block that already sets it.
+    diagnose.
 
-    Raises: HarlequinSshError if a value would reach `ssh` as a flag.
+    Raises: HarlequinSshError for a value ssh would read as an option or expand
+    into a shell.
     """
-    _refuse_a_flag("--ssh-host", host)
+    _refuse_unsafe_value("--ssh-host", host)
     for forward in forwards:
-        _refuse_a_flag("--ssh-forward", forward)
+        _refuse_unsafe_value("--ssh-forward", forward)
     argv = [SSH, "-N", "-o", "ExitOnForwardFailure=yes"]
     if batch_mode:
         argv += ["-o", "BatchMode=yes"]
@@ -158,10 +169,10 @@ def resolve_config(argv: Sequence[str], *, timeout: float) -> SshConfig | None:
     """What `ssh -G` says the argv about to run resolves to, or None.
 
     None is the degraded answer -- no client, no `-G`, or output this cannot
-    read -- and every caller reads it as "ask ssh nothing else": no poll, no
-    forwards-nothing error, and the adapter's own connection is the test.
+    read -- and every caller reads it as "ask ssh nothing else": no poll and no
+    forwards-nothing error.
     """
-    probe = [argv[0], "-G", *argv[1:]]
+    probe = [_client_path(argv[0]), "-G", *argv[1:]]
     try:
         completed = subprocess.run(
             probe,
@@ -206,10 +217,8 @@ def parse_config(text: str) -> SshConfig | None:
 class SshTunnel:
     """An `ssh` child process holding one or more local forwards open.
 
-    A child rather than `ssh -f`: the process that survives a fork is not ours
-    to kill, which is the orphaned tunnel this feature exists to remove. It also
-    keeps the terminal, so a passphrase or 2FA prompt still reaches a human --
-    which is why both commands start it before anything owns the screen.
+    It inherits the terminal, so a passphrase or 2FA prompt still reaches a
+    human, and it dies with the process that started it.
     """
 
     def __init__(
@@ -259,19 +268,21 @@ class SshTunnel:
         if self.running:
             return
         self.reused = False
-        already_bound = any(_accepts(endpoint) for endpoint in self.endpoints)
+        already_bound = frozenset(
+            endpoint for endpoint in self.endpoints if _accepts(endpoint)
+        )
+        argv = [_client_path(self.argv[0]), *self.argv[1:]]
         try:
             process = subprocess.Popen(
-                self.argv,
+                argv,
                 # ssh's prompts go to the terminal it inherits, not to this
-                # pipe, so capturing stderr costs no interactivity and buys an
-                # error message written by the program that failed.
+                # pipe, so capturing stderr does not swallow them
                 stdout=subprocess.DEVNULL,
                 stderr=subprocess.PIPE,
             )
         except OSError as e:
             raise HarlequinSshError(
-                f"Harlequin could not run {self.argv[0]}: {e}", title=ERROR_TITLE
+                f"Harlequin could not run {argv[0]}: {e}", title=ERROR_TITLE
             ) from e
         self._process = process
         self._stderr = _Stderr(process.stderr)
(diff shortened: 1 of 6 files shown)


src/harlequin/ssh.py
```diff
@@ -0,0 +1,458 @@
...
+            answering = all(_accepts(endpoint) for endpoint in self.endpoints)
+            if exited:
+                # `ExitOnForwardFailure=yes`, so ssh exits rather than connect
+                # without the forward. Ports that answer anyway are someone
+                # else's tunnel, which is a thing to use only when asked.
+                if answering and already_bound and self.allow_reuse:
+                    self.reused = True
+                    return
+                raise self._child_failed()
+            if answering and (not already_bound or time.monotonic() >= settled):
```
tconbeer: **Blocker: `allow_reuse=False` does not fail on a bound port.**

§5.3 mandates *default → fail*, on the grounds that "a port that answers is not proof the right tunnel is behind it." This gate does not deliver that.

`_SETTLE_SECONDS` is a wall clock, but **OpenSSH binds local forwards only after authentication**. Any auth slower than 1s — passphrase, 2FA push, `ProxyJump`, slow bastion — elapses the window while the child is still alive and has bound nothing. The state for that whole interval is `exited=False, answering=True (the foreign listener), already_bound=True`, and the gate opens.

Reproduced with an accepting foreign listener and a client that reaches its bind after 3s:

```
SUCCESS after 1.01s | reused=False | running=True
```

The adapter is then pointed at another process's socket and sends it the database credentials. Three compounding effects:

- **`reused` stays `False`** even with `allow_reuse=True`, so the §5.3 notice never prints on this path and 3/5's reporting will be wrong.
- **The reuse branch at line 330 is unreachable whenever the client is slow**, because this line returned before the child exited.
- **`timeout <= 1.0` collapses the window** onto the deadline, and this check precedes the deadline check below. Verified: `timeout=0.5` with a pre-bound port returns success at 0.51s instead of timing out.

With `already_bound`, the only sound terminators are *the child exiting* or *the deadline* — a fixed wall clock cannot express "ssh has had its chance to fail." `ExitOnForwardFailure=yes` means a real client losing the bind race does exit, so requiring that (or the deadline) converts every one of these into `_child_failed()`.

Note the suite cannot currently see this: the fake `ssh` binds at t≈0, so `test_a_port_that_is_already_bound_fails` passes on timing a real client does not have. `FAKE_SSH_DELAY=1.5` plus the `listener` fixture reproduces it in-suite.

---

src/harlequin/ssh.py
```diff
@@ -0,0 +1,458 @@
...
+
+    def start(self) -> None:
+        """Start it and block until the forwarded ports accept connections.
+
+        Raises: HarlequinSshError, quoting ssh's stderr.
+        """
+        if self.running:
+            return
+        self.reused = False
+        already_bound = any(_accepts(endpoint) for endpoint in self.endpoints)
```
tconbeer: TOCTOU on this sample, same root cause as the gate at line 334. `already_bound` is read once, before `Popen`; the window between it and ssh's own bind is the entire SSH handshake. Anything that binds the port inside that window gets `not already_bound` → the readiness loop returns on the first affirmative with no settle at all.

Also worth noting this is `any(...)` collapsed to one bool, so a single pre-bound port among several makes every endpoint wait out the settle window. Per-endpoint state would fix that and read more directly.

---

src/harlequin/ssh.py
```diff
@@ -0,0 +1,458 @@
...
+
+def _refuse_a_flag(option: str, value: str) -> None:
+    """Refuse a value `ssh` would read as an option rather than as a value.
+
+    `ssh` has no `--`, so a destination or a forward spec starting with `-`
+    reaches it as a flag -- and a config file is discovered in the working
+    directory, where a cloned repo must not be able to smuggle in an
+    `-oProxyCommand=`.
+    """
+    if value.startswith("-"):
```
tconbeer, line 518: **The dash check is too narrow for a value that can arrive from a discovered config file.**

Placement is right — every value reaching argv passes through `build_argv`, and `resolve_config`'s probe derives from an already-validated argv. The character set is the problem.

`--ssh-host` becomes `%h`/`%r` (and `%p` from an `ssh://…:port` destination). If the user's own `~/.ssh/config` has a `ProxyCommand` referencing them — `ProxyCommand ssh -W %h:%p bastion` is canonical, and MDM-managed configs are full of them — a **non-dash-leading** destination reaches OS command injection:

- **CVE-2023-51385** — shell metacharacters (backticks, `$(…)`) expanded unquoted into `ProxyCommand` / `LocalCommand` / a `Match exec` predicate. OpenSSH < 9.6.
- **CVE-2025-61984** — bypass of that fix: control characters, notably `\n`, were not in the 9.6 ban, and a newline in the username splits the command the shell runs. OpenSSH < 10.1. TOML writes `\n` trivially.

Both were originally reported against *git submodule URLs* — the identical threat shape, since `harlequin.config` discovers `harlequin.toml` / `.harlequin.toml` / `pyproject.toml` in `Path.cwd()` (`config.py:701`) and a repo-local file can supply both the profile and `default_profile`. `cd hostile-repo && hsql -c 'select 1'` is the whole exploit.

Suggest making this an allow-list rather than a dash check: reject control characters (`\n`, `\r`, `\0`, anything < 0x20) and shell metacharacters in `--ssh-host` and `--ssh-forward`. Harlequin has no legitimate destination or forward spec containing them, and it is the one guard that does not depend on which OpenSSH the user happens to have.

(I did not verify the two version boundaries against NVD or OpenSSH source — they come from the published write-ups. The recommendation doesn't depend on them being exact.)

---

src/harlequin/ssh.py
```diff
@@ -0,0 +1,458 @@
...
+import subprocess
+import threading
+import time
+from collections import deque
+from dataclasses import dataclass
+from typing import Any, Sequence
+
+from harlequin.exception import HarlequinSshError
+
+SSH = "ssh"
```
tconbeer, line 36: **Windows resolves this against the current directory before `PATH`.**

`subprocess.Popen(["ssh", …])` with no `executable=` and `shell=False` reaches `CreateProcess` with a NULL `lpApplicationName`, whose documented search order includes the current directory ahead of `PATH`. A repo that ships `ssh.exe` in its root gets arbitrary code execution the first time a Windows user runs Harlequin there with an `ssh_host` set — no config-file involvement needed beyond a profile they already have.

`shutil.which("ssh")` is not a fix on Windows either; it prepends `os.curdir`. Resolving against `PATH` with `os.curdir` removed, or setting `NoDefaultCurrentDirectoryInExePath` in the child env, closes it.

Lands with 2/5 in practice, but the fix belongs in this module.

(From documented search-order behavior — I did not execute this on Windows.)

---

src/harlequin/ssh.py
```diff
@@ -0,0 +1,458 @@
...
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.PIPE,
+            )
+        except OSError as e:
+            raise HarlequinSshError(
+                f"Harlequin could not run {self.argv[0]}: {e}", title=ERROR_TITLE
+            ) from e
+        self._process = process
+        self._stderr = _Stderr(process.stderr)
+        atexit.register(self.stop)
```
tconbeer, line 307: **`atexit` is the only backstop, and hsql's own hot path bypasses it.**

`atexit` does not run on `SIGTERM`/`SIGHUP` (no handler is installed anywhere), and it does not run under `os._exit()`. `hsql`'s `timeout._halt()` (`src/harlequin/hsql/timeout.py:124`) calls `os._exit()` on exactly the path a tunnel makes likely — a query that outlives `--timeout` plus its grace period.

So a timed-out `hsql` run leaves the `ssh` child alive, holding an unauthenticated local port forwarded to the internal database, for as long as the machine is up. On a shared host any local user reaches production through it. That is the orphaned tunnel §8 says this feature exists to remove.

The child is in Harlequin's process group, so a terminal Ctrl-C reaches it, but nothing else does. Wants a `stop()` before `_halt()` flushes, plus `SIGTERM`/`SIGHUP` handling around the `ExitStack` in the later PR.

---

src/harlequin/ssh.py
```diff
@@ -0,0 +1,458 @@
...
+    if host in _WILDCARD_HOSTS:
+        # a forward bound to every interface is reachable on the loopback, which
+        # is the interface the adapter is pointed at
+        host = _LOOPBACK
+    elif host == "::":
+        host = "::1"
+    return host, int(port)
+
+
+def _pretty(field: str) -> str:
```
tconbeer, line 591: **A `0.0.0.0` forward is *reported* as loopback.**

`_endpoint()` rewrites the wildcard bind to `127.0.0.1` so it can be polled, and `_pretty` / `Forward.__str__` reuse it. Verified:

```python
str(Forward("[0.0.0.0]:15439", "[db.internal]:5439"))
# '127.0.0.1:15439 -> db.internal:5439'
```

Since `ssh_forward` will be a profile key, a repo-local config can open `0.0.0.0:15439:prod-db:5432` — exposing the user's internal database to their whole LAN — and the one line §4 says "tells a user which database they are actually looking at" actively conceals it. §8 asserts that `0.0.0.0` "should take deliberate effort"; via a discovered config file it takes none from the user.

Keep the loopback rewrite for polling; print the real bind address.

Related, for 2/5: `ssh_allow_reuse` as a profile key means §5.3's fail-closed default can be turned off by a file the user did not write. Worth considering honoring that one from the command line only.

---

src/harlequin/ssh.py
```diff
@@ -0,0 +1,458 @@
...
+        self._thread.join(timeout=_TERMINATE_SECONDS)
+
+    def close(self) -> None:
+        self._thread.join(timeout=_TERMINATE_SECONDS)
+        with contextlib.suppress(OSError):
+            self._stream.close()
+
+    def _drain(self) -> None:
+        with contextlib.suppress(OSError, ValueError):
+            while chunk := self._stream.read(1024):
```
tconbeer: **`read(1024)` on a `BufferedReader` blocks until 1024 bytes or EOF, so the timeout error can never quote ssh.**

`_timed_out()` (line 355) correctly does *not* call `settle()` — the child is alive, there is no EOF to wait for. But that means this thread is still parked inside `read(1024)` and `_chunks` is empty. Every real ssh diagnostic is far under 1KB, so `detail` is always `""` on the timeout path.

Verified — child writes `Enter passphrase for key '/home/tco/.ssh/id_ed25519':` (51 bytes) and hangs, `timeout=1.5`:

```
ssh did not open the forward within 1.5s. It is most likely waiting for a
passphrase, a password, or confirmation of a host key; ...
```

Nothing appended. The same child writing >1024 bytes does produce the detail, which pins the cause. Confirmed in isolation that `read1(1024)` returns the 51 bytes immediately where `read(1024)` blocks indefinitely.

This is precisely the path where the user most needs ssh's own words — *which* credential, *which* host key. `read1(1024)` fixes it without changing anything else about the drain.

---

src/harlequin/ssh.py
```diff
@@ -0,0 +1,458 @@
...
+
+    def settle(self) -> None:
+        """Wait for the drain to reach the end of what the child wrote.
+
+        `Popen.wait()` returns when the child exits, which is before the pipe it
+        wrote to has been read to the end -- so a caller that quotes ssh calls
+        this first, or quotes nothing. Only useful once the child is gone;
+        while it runs there is no end to wait for.
+        """
+        self._thread.join(timeout=_TERMINATE_SECONDS)
```
tconbeer: EOF on the pipe requires *every* write end closed, not just the child's. A `ProxyCommand`/`ProxyJump` helper, an askpass, or a backgrounded `ControlPersist` master inherits ssh's stderr and holds one open. Then this join stalls the full `_TERMINATE_SECONDS` and `_child_failed()` still quotes nothing.

Reproduced — child writes `Permission denied (publickey).`, spawns a grandchild inheriting stderr, exits: `settle took 2.00s; text=''`.

`ProxyJump` is an advertised supported case (§3.3), so this is reachable in normal use. Reading until the *child is reaped* rather than until EOF avoids both the stall and the loss.

---

src/harlequin/ssh.py
```diff
@@ -0,0 +1,458 @@
...
+"""What a tunnel with no port to poll waits instead, before trusting the child."""
+
+_TERMINATE_SECONDS = 2.0
+"""How long a terminated child has to exit before it is killed."""
+
+_STDERR_LIMIT = 8192
+"""Bytes of ssh's stderr kept for an error to quote."""
+
+_LOOPBACK = "127.0.0.1"
+_WILDCARD_HOSTS = {"", "*", "0.0.0.0", "localhost"}
```
tconbeer: `localhost` in this set collapses to `127.0.0.1` with no other-family fallback, unlike `socket.create_connection(("localhost", port))`, which tries every `getaddrinfo` result.

On a host where `localhost` resolves only to `::1` (IPv6-only container images), the default local forward — `localforward 15439 …`, listen host absent, which is the motivating case — binds `::1`, the probe of `127.0.0.1:15439` refuses forever, and a working tunnel dies at the deadline with the misleading "most likely waiting for a passphrase" message.

Also note `localhost` isn't a wildcard the way the other three are; the comment below this ("a forward bound to every interface") doesn't describe it.

---

src/harlequin/ssh.py
```diff
@@ -0,0 +1,458 @@
...
+        """Block until every forwarded port answers, or say why it never did."""
+        assert self._process is not None
+        started = time.monotonic()
+        deadline = started + self.timeout
+        settled = started + min(_SETTLE_SECONDS, self.timeout)
+        if not self.endpoints:
+            # degraded, or forwarding to a socket path: there is no port to wait
+            # on, so the child gets a moment to fail outright and the adapter's
+            # connection is the real test.
+            time.sleep(min(_GRACE_SECONDS, self.timeout))
```
tconbeer: Minor, and not reachable at this head: a negative `timeout` makes this `time.sleep()` raise `ValueError`, which escapes `start()` past the `except BaseException: self.stop(); raise`. Once `--ssh-timeout` / `ssh_timeout` lands in 2/5, `--ssh-timeout -1` is a traceback rather than an error modal.

`resolve_config` already tolerates it (`TimeoutExpired` → degraded `None`), so validating at the option boundary is probably the cheaper fix.

---

src/harlequin/ssh.py
```diff
@@ -0,0 +1,458 @@
...
+                forwards.append(Forward(listen=fields[0], destination=fields[1]))
+    if not is_ssh_output:
+        return None
+    return SshConfig(forwards=tuple(forwards))
+
+
+class SshTunnel:
+    """An `ssh` child process holding one or more local forwards open.
+
+    A child rather than `ssh -f`: the process that survives a fork is not ours
```
tconbeer: `AGENTS.md`: *"Never record a decision, a tradeoff, or feedback from a PR comment"* / *"never reference code that was removed or refactored."*

> A child rather than `ssh -f`: the process that survives a fork is not ours to kill, which is the orphaned tunnel this feature exists to remove.

This argues the implementation against an alternative and states the feature's motivation — PR-description material. The trailing *"which is why both commands start it before anything owns the screen"* also legislates callers in other modules.

Suggest:

```python
"""An `ssh` child process holding one or more local forwards open.

It inherits the terminal, so a passphrase or 2FA prompt still reaches a human.
"""
```

---

src/harlequin/ssh.py
```diff
@@ -0,0 +1,458 @@
...
+    host: str,
+    forwards: Sequence[str] = (),
+    *,
+    batch_mode: bool = False,
+) -> list[str]:
+    """The `ssh` command that holds `forwards` open, and runs nothing.
+
+    `ExitOnForwardFailure` is the only option Harlequin imposes on its own: a
+    forward that silently did not happen is the one failure a user cannot
+    diagnose. Notably not imposed is `ServerAliveInterval`, which a command-line
```
tconbeer: > Notably not imposed is `ServerAliveInterval`, which a command-line `-o` would take away from a `Host` block that already sets it.

A sentence about code that isn't there — a road not taken, which the reader has no way to act on. Suggest cutting it entirely; the preceding sentence about `ExitOnForwardFailure` already says what *is* imposed and why.

---

src/harlequin/ssh.py
```diff
@@ -0,0 +1,458 @@
+"""An `ssh` child process holding local forwards open, for either command.
+
+Harlequin runs `ssh` and touches nothing else. The connection details already
+name the local end of a forward -- `host = "localhost"`, `port = 15439` -- so no
+adapter is told a tunnel exists, and every adapter works, including the ones
```
tconbeer: Two bits of the module docstring are advocacy for the design rather than statements about the code:

- line 5: *"and every adapter works, including the ones nobody in this org maintains"*
- line 11: *"and its error messages are better than any we would write"*

And line 14's *"the motivating case puts the forward in `~/.ssh/config`"* names the issue behind the PR rather than the behavior.

Suggest:

```
Harlequin runs `ssh` and touches nothing else. The connection details already
name the local end of a forward -- `host = "localhost"`, `port = 15439` -- so no
adapter is told a tunnel exists.
...
Nothing a user typed is parsed: `ssh` owns the syntax of a destination and of a
forward spec, so both values reach it verbatim and its stderr is what an error
quotes. The one thing this module reads is `ssh -G`'s own output
(`parse_config()`), which says which local port to wait on: a forward declared
in `~/.ssh/config` never reaches Harlequin as a port number.
```

Calibrated against `query.py` / `statements.py`, which carry comparably dense prose legitimately — the test is whether a sentence states something non-obvious about what the code *does*.

---

src/harlequin/ssh.py
```diff
@@ -0,0 +1,458 @@
...
+DEFAULT_TIMEOUT = 10.0
+"""Seconds to wait for the forwards, when nothing says otherwise."""
+
+ERROR_TITLE = "Harlequin could not open the SSH tunnel."
+
+_POLL_INTERVAL = 0.05
+_CONNECT_TIMEOUT = 0.5
+"""How long one probe of one local port may take."""
+
+_SETTLE_SECONDS = 1.0
```
tconbeer: Flagging this docstring specifically because it asserts the property my blocker comment at line 334 shows the code doesn't have:

> How long a port that was already bound gets to prove it was ssh's to take.

A port cannot prove that by answering after a fixed interval, since ssh hasn't necessarily reached its bind yet. Whatever shape the fix takes, this text needs to change with it.

---

src/harlequin/ssh.py
```diff
@@ -0,0 +1,458 @@
...
+            time.sleep(min(_GRACE_SECONDS, self.timeout))
+            if self._process.poll() is not None:
+                raise self._child_failed()
+            return
+        while True:
+            exited = self._process.poll() is not None
+            answering = all(_accepts(endpoint) for endpoint in self.endpoints)
+            if exited:
+                # `ExitOnForwardFailure=yes`, so ssh exits rather than connect
+                # without the forward. Ports that answer anyway are someone
```
tconbeer: Collecting the milder prose trims here rather than as separate threads — all the same convention, none blocking:

- **line 328-329** *"Ports that answer anyway are someone else's tunnel, which is a thing to use only when asked."* — the second clause restates the `and self.allow_reuse` on the line below. → `# ports that answer after ssh exited are someone else's tunnel`
- **line 316-318** *"…and the adapter's connection is the real test."* — the same sentence appears in `resolve_config`'s docstring (160) and `_GRACE_SECONDS` (53). Keep it in one place; here → `# no port to wait on: give the child a moment to fail outright`
- **line 269-271** *"…costs no interactivity and buys an error message written by the program that failed"* — cost/benefit framing. → `# ssh's prompts go to the terminal it inherits, not to this pipe, so capturing stderr does not swallow them`
- **line 91-93** *"…so there are never two sources to keep in agreement"* — justifies the shape against an unbuilt alternative. Keep the first half.
- **line 71-72** *"Kept as text because that is all a notice needs"* — drop the justification, keep "only the listening side is ever read for a number."
- **line 383-385** *"Only useful once the child is gone; while it runs there is no end to wait for."* — implied by the two sentences before it.
- **line 34** *"whichever one the user's `~/.ssh/config` is for"* — states nothing checkable. → `"""The client, found on PATH."""`
- **tests**: `tests/data/unit_tests/ssh/ssh:6` and `tests/unit_tests/test_ssh.py:5` carry the same sentence nearly verbatim ("every observable thing a real forward does…"). Keep it in the fake binary, where the claim is about that code.

Judged fine as-is, for contrast: `_refuse_a_flag`, `resolve_config`, `parse_config`, and the Windows/Unix wording note in `test_ssh.py:255` — each states a behavior a caller or reader must know.

---

tconbeer (commented):
Reviewed under three explicit lenses, run as separate passes so none contaminated the others: **correctness/logic**, **security** (unexpected access to user data), and **comments/docstrings** against `AGENTS.md` conventions. The two top findings were reproduced against running code before being written up.

**Verdict: changes needed before merge** (posting as a comment review — GitHub won't let me request changes on my own PR).

**One blocker**, at `ssh.py:334`: `allow_reuse=False` — the default §5.3 calls "the safe direction" — does not fail on a bound port. The settle window is a wall clock, but OpenSSH binds local forwards only *after* authentication, so any auth slower than 1s lets a foreign listener satisfy the readiness check. `start()` returns success at ~1.01s with `reused=False`, and the adapter is pointed at another process's socket with no signal to the user. Three variants compound it (TOCTOU on the pre-check, the unreachable reuse branch, and `timeout <= 1.0` collapsing the window).

Worth stating plainly: **CI being green across all 32 checks is not evidence against this.** The bug is invisible to the suite by construction — the fake `ssh` binds at t≈0, so `test_a_port_that_is_already_bound_fails` passes on timing a real client does not have. `FAKE_SSH_DELAY=1.5` plus the existing `listener` fixture reproduces it in-suite, and that regression test is worth having regardless of the fix's shape.

The rest, inline: a narrow argument-injection guard (`ProxyCommand` token expansion reached by a non-dash-leading value from a discovered `pyproject.toml`), Windows `CreateProcess` resolving `ssh` from the current directory, an orphaned tunnel via `hsql`'s `os._exit()` path, a `0.0.0.0` forward reported as loopback, and a stderr drain that makes the timeout error unable to quote ssh. Several land with 2/5–5/5 but are fixed in this module.

The design itself reads well, and the module is a clean seam — these are gaps between what §5.3 and §8 specify and what the code enforces, not objections to the approach.

**Not checked**, and worth knowing where the next bug is:

- I did not review 2/5–5/5. Where a finding concerns code those PRs add, I assessed only where the fix belongs.
- I did not run the suite locally (`duckdb` unavailable in my container; `conftest.py` won't import). I relied on this PR's CI for "the tests pass," and verified separately *why* they pass.
- I did not test against a real `ssh` client or a real SSH server. The auth-timing premise behind the blocker is OpenSSH's documented ordering, not an observed handshake; my repro simulates it with a delay.
- I did not verify the CVE version boundaries against NVD or OpenSSH source.
- I did not check Windows or macOS behavior directly.
- I did not audit the IDE's error-rendering path beyond confirming `pretty_print_error` applies no redaction — whether ssh's stderr should route through `redact_text()` is a pre-existing gap this module is the first to feed third-party output into, and a judgment call rather than a finding.
- I did not review the fake `ssh` helper as a test-quality matter, only for docstring duplication.

---

tconbeer:
Thank you — the blocker is real and I've fixed it. Answering the findings in one comment rather than fifteen; every one of them is addressed, in 1/5 unless noted.

## The blocker — `allow_reuse=False` did not fail on a bound port

You're right about the mechanism, and the consequence: with auth slower than the settle window, `start()` returned success and the adapter got another process's socket.

The settle window is gone. When a port was already bound, readiness now comes only from **the child exiting** — which `ExitOnForwardFailure=yes` guarantees when it cannot take the port — or from **the deadline**. A fixed clock can't express "ssh has had its chance to fail," as you said, so it no longer tries to.

Three follow-on changes:

- `already_bound` is a `frozenset` of endpoints rather than one bool, so a single pre-bound port among several is visible as itself.
- **A pre-bound port is not re-probed while the child runs.** Its answer would say nothing, and — found while writing your regression test — a listener whose owner never accepts has a backlog to fill: at 20 Hz the probes filled the test's `listen(8)` queue and the *reuse* path then failed for the wrong reason. One probe at the decision point.
- The `already_bound` timeout gets its own message, naming the port and `--ssh-allow-reuse`, instead of guessing at a passphrase. That covers the `timeout <= 1.0` variant, which now times out as it should.

Your repro is in the suite as three tests (`FAKE_SSH_DELAY=1.5` + the `listener` fixture): fail by default, reuse when asked, and the sub-second timeout.

**On CI being green: agreed, and the point stands generally.** The fake client binds at t≈0, so the suite was asserting a property of the fake's timing rather than of the code.

## Security

- **ProxyCommand token expansion.** Fixed, as a deny-list rather than an allow-list: control characters (`< 0x20`, `0x7f`) plus `` ` $ \ " ' | & ; < > ( ) { } ``, on both `--ssh-host` and `--ssh-forward`. An allow-list would have to admit the grammar of a unix socket path and the bracketed IPv6 forms, and I'd rather not be the one deciding what a legal socket path looks like; the deny set covers every primitive in both CVEs and no legitimate destination or forward spec. The dash check stays.
- **Windows `CreateProcess` and the working directory.** Fixed by resolving `ssh` against `PATH` explicitly, at spawn time. Not `shutil.which()` — as you noted it prepends `os.curdir`, and in 3.10 it does so even when `path=` is passed, so it is unusable here.
- **`ssh_allow_reuse` from a config file.** Taken — this landed in #1113. A key that turns off a fail-closed check shouldn't be answerable by a file discovered in the working directory, so `take_ssh_keys()` now takes the names that were actually typed and refuses the key from a profile, naming the flag.
- **`0.0.0.0` reported as loopback.** Fixed: the rewrite is for polling only, and `_pretty` prints the bind address ssh reported.
- **The orphaned tunnel.** `stop_all()` before `_halt()` was already in #1113; `SIGTERM`/`SIGHUP` handling now is too, as `stopping_on_signal()`, entered beside the tunnel in both commands so a run without one changes no signal disposition. The tunnels are stopped from the handler rather than by the exception it raises, since something further up may catch it.

## Correctness

- **`read(1024)` blocking.** Fixed with `read1()`. This was the best catch of the set — the timeout path is exactly where ssh's own words matter, and it could never have them. There's a test now with a 51-byte passphrase prompt.
- **The `settle()` stall.** Bound dropped to 0.5s with its own constant, explaining that EOF needs every write end closed and a `ProxyJump` helper holds one. With `read1()` the bytes are in hand almost immediately, so the join is a backstop rather than the mechanism.
- **`localhost` in the wildcard set.** Fixed: a bare port and the wildcards are polled as `localhost` so `create_connection` tries every family; a literal `localhost` is left alone for the same reason.
- **Negative `timeout` reaching `time.sleep()`.** Guarded here, and click plus `parse_seconds()` refuse it at the option boundary in #1113.

## What I did not close

**The TOCTOU you flagged at line 262 is narrowed, not closed.** `already_bound` is still read once before `Popen`, so a listener that appears *during* the handshake is still indistinguishable from our own forward. I don't think it's closeable without asking ssh what it bound — parsing `-v` output, which I'd rather not make load-bearing. It now requires an adversary racing the handshake rather than merely holding the port first, which is the part that was reachable by accident.

## Comments and docstrings

All applied, including the collected batch: the module docstring, the class docstring (your replacement, near enough verbatim), the `ServerAliveInterval` road-not-taken, `_SETTLE_SECONDS`' claim (the constant is gone), and the duplicated sentence between the fake client and the test module — kept in the fake, where the claim is about that code.

---

Validated locally at each step: the full suite on 1/5 (1391 passed), 2/5 (1411), 3/5 (1417) and 4/5 (1422), plus mypy, ruff and the import-linter contracts. The fixes are merged forward through 2/5–5/5, so all five heads carry them.

---

tconbeer:
Two fixes from your Tailscale SSH session — both real, and the second one shows my reasoning was wrong.

**The default wait is now 60s, not 10s.** The wait is on a person as often as on a network: an ssh that opens a browser for an identity provider, or asks for a hardware key, is a minute of someone reading a screen. `--ssh-batch-mode` is still how an unattended caller fails at the first prompt rather than waiting it out.

**ssh's stderr passes through to ours as it arrives.** I justified buffering it with "ssh writes its prompts to `/dev/tty`, so capturing costs no interactivity" — true of ssh's own passphrase prompt, and not at all of a helper's instructions, which go to stderr. Tailscale's authentication URL is exactly that case, and it appeared only inside the failure, after it had stopped being useful. Now:

```
# Tailscale SSH requires an additional check.
# To authenticate, visit: https://login.tailscale.com/a/l4fc5f072

[60s later, if nobody visited it]
╭─ Harlequin could not open the SSH tunnel. ─────────────────────────────╮
│ ssh did not open the forward within 60s. It is most likely waiting for │
│ a passphrase, a password, or confirmation of a host key; answer it, or │
│ pass --ssh-batch-mode to fail immediately instead of waiting.          │
╰────────────────────────────────────────────────────────────────────────╯
```

Details worth naming:

- **A line at a time**, so `redact_text()` sees whole lines. That closes the gap you flagged as "not checked" — ssh is third-party output, and a driver's DSN can reach it — without risking a secret split across a read boundary.
- **A last line with no newline is flushed** before the tunnel gives up, since a prompt has none.
- **The errors no longer quote what was echoed**, so a helper's instructions are not printed twice. `echo_stderr` goes false on a restart (4/6), where nobody is reading stderr, and the text goes back into the error for the modal.

`DEFAULT_SSH_TIMEOUT` is the shared constant behind both commands' `--ssh-timeout`, so the generated schema and CLI reference moved with it. Merged forward through the stack; 1434 tests pass on the tip.

There is also a new **6/6, #1123**: the config wizard now asks "Do you connect via SSH?" before the adapter options.

---

PR #1113, opened 2026-09-01
SSH tunnels 2/5: --ssh-host and friends, in both commands

tconbeer:
Part 2 of 5 of `docs/plans/ssh-tunnels-design.md`. **Based on #1112**, which is where `harlequin/ssh.py` lands; review that one first, and this diff on top of it.

Stack: 1/5 tunnel → **2/5 (this)** → 3/5 keepalive & reporting → 4/5 reconnect → 5/5 changelog.

Part of #545

Five options, in `harlequin` and in `hsql`, and one profile key each: `--ssh-host`, `--ssh-forward` (repeatable), `--ssh-batch-mode`, `--ssh-allow-reuse`, `--ssh-timeout`.

```toml
[profiles.redshift]
adapter = "postgres"
host = "localhost"
port = 15439
dbname = "prod"
ssh_host = "redshift_prod"
```

- `harlequin.config` names the five keys (`SSH_KEYS`) and `take_ssh_keys()` takes them off a merged config; `harlequin.ssh.open_tunnel()` turns what it took into a started child.
- The tunnel opens in the click callback, before Textual and before anything connects. `hsql` registers `tunnel.stop` with `ctx.call_on_close`; the IDE wraps `tui.run()` in an `ExitStack`.
- `hsql` writes the tunnel's line to stderr (`note: ssh: 127.0.0.1:15439 -> data.example.com:5439 via redshift_prod`) and exits 3 for a tunnel that would not open, quoting ssh. The IDE notifies instead, and gets the tunnel object as `Harlequin(ssh_tunnel=...)`.
- The ssh destination and its resolved forwards join the connection hash, via a new `through=` argument to `get_connection_hash()`.
- The generated `schemas/config-v1.json` and `docs/generated/hsql-reference.md` are regenerated, which is what makes the profile keys valid ones.

- **The adapter is never told a tunnel exists.** `conn_str` and every adapter option reach it exactly as typed — which is why any adapter works, including out-of-tree ones. That is §2 of the design, and there is a test for it.
- **Two exit codes, not one.** A destination that forwards nothing is a *usage* error (exit 2), because nothing has run yet and what is wrong is what was asked for; `build_tunnel()` raises `HarlequinConfigError` there. A tunnel that would not open is a database that cannot be reached (exit 3), which `exit_code_for()` now maps `HarlequinSshError` to.
- **The five options are declared in each command rather than shared.** Importing a shared module at CLI import time would put `harlequin.ssh` on the path of every invocation, and §6 promises it is imported only when `ssh_host` is set. A test pins the two declarations to the same spellings, in the spirit of the copies the two commands already keep of each other.
- **`--ssh-timeout` carries no click default**, so that "any ssh key set with no destination" is detectable and refused rather than silently connecting past a forward the user meant to use.
- **Two bastions fronting two databases both look like `localhost:15439`.** Without the cache-key change they would share a catalog cache and a query history. `through=` is omitted from the hashed material when there is no tunnel, so an untunneled connection keys exactly as it always has.
- **`hsql --timeout`'s hard exit now stops the tunnel** (`harlequin.ssh.stop_all()`): `os._exit()` runs no `atexit` handler, and an ssh child that outlives the run is the thing this feature removes.

- [x] Yes.

End-to-end tests drive the real CLI path for both commands against the fake `ssh` on `PATH` (Unix only, as the design says): a query runs through a tunnel and the child is gone with the run; the forwards reach the argv unchanged and in order; a mode that never connects opens nothing; a profile carries the tunnel; two bastions hash differently; the adapter is handed no `ssh_*` key. Plus unit tests for `take_ssh_keys()` and `SshOptions`.

---

### DIFF
--- a/src/harlequin/app.py
+++ b/src/harlequin/app.py
@@ -353,12 +353,15 @@ class Harlequin(AppBase):
         self.run_query_bar.apply_configured_limit()
 
         if self.ssh_tunnel is not None:
-            # which database this session is actually looking at, which the
-            # connection details alone no longer say
+            # which database this session is actually looking at
+            warnings = self.ssh_tunnel.warnings()
             self.notify(
-                self.ssh_tunnel.notice(),
+                "\n\n".join((self.ssh_tunnel.notice(), *warnings)),
                 title="SSH tunnel",
-                severity="warning" if self.ssh_tunnel.reused else "information",
+                severity=(
+                    "warning" if self.ssh_tunnel.reused or warnings else "information"
+                ),
+                markup=False,
             )
 
         self._connect()
--- a/src/harlequin/app.py
+++ b/src/harlequin/app.py
@@ -96,6 +96,7 @@ if TYPE_CHECKING:
     from textual.await_complete import AwaitComplete
 
     from harlequin.keymap import HarlequinKeyMap
+    from harlequin.ssh import SshTunnel
 
 
 class CatalogCacheLoaded(Message):
@@ -201,6 +202,7 @@ class Harlequin(AppBase):
         export_path: Path | str | None = None,
         viewer_max_rows: int | str | None = 100_000,
         query_limit: int | str | None = None,
+        ssh_tunnel: SshTunnel | None = None,
         driver_class: Union[Type[Driver], None] = None,
         css_path: Union[CSSPathType, None] = None,
         watch_css: bool = False,
@@ -217,6 +219,9 @@ class Harlequin(AppBase):
         self.history: History | None = None
         self.show_files = show_files
         self.show_s3 = show_s3 or None
+        # already started, by the command that built this app: `ssh` prompts for
+        # a passphrase on the terminal Textual is about to take.
+        self.ssh_tunnel = ssh_tunnel
         # kept as text: it is what the Data Exporter's path input starts with
         self.export_path = str(export_path) if export_path is not None else None
         # None is no cap: the viewer holds every row that was fetched. So are 0
@@ -346,6 +351,15 @@ class Harlequin(AppBase):
 
     async def on_mount(self) -> None:
         self.run_query_bar.apply_configured_limit()
+
+        if self.ssh_tunnel is not None:
+            # which database this session is actually looking at, which the
+            # connection details alone no longer say
+            self.notify(
+                self.ssh_tunnel.notice(),
+                title="SSH tunnel",
+                severity="warning" if self.ssh_tunnel.reused else "information",
+            )
 
         self._connect()
         self._load_catalog_cache()
--- a/src/harlequin/cli.py
+++ b/src/harlequin/cli.py
@@ -38,6 +38,7 @@ from harlequin.keys_app import HarlequinKeys
 from harlequin.locale_manager import set_locale
 from harlequin.options import AbstractOption
 from harlequin.plugins import adapter_names, load_adapter, load_adapter_plugins
+from harlequin.redact import hide_secrets_in
 from harlequin.windows_timezone import check_and_install_tzdata
 
 if TYPE_CHECKING:
@@ -284,8 +285,8 @@ def _adapter_option_group(
 def _open_tunnel(ctx: click.Context, ssh_config: dict[str, Any]) -> "SshTunnel | None":
     """The tunnel the IDE runs through, or exit having said why there is none.
 
-    Here rather than in the app: once Textual owns the terminal, `ssh` cannot
-    ask for a passphrase and a 2FA push has nowhere to print "check your phone".
+    Once Textual owns the terminal, `ssh` cannot ask for a passphrase and a 2FA
+    push has nowhere to print "check your phone".
     """
     if not ssh_config.get("ssh_host"):
         # nothing to open, and so nothing to import
@@ -589,6 +590,11 @@ def build_cli(argv: Sequence[str]) -> click.Command:
             profile=profile_config, cli_values=kwargs, explicitly_set=explicitly_set
         )
 
+        # what this process must not print, before anything can print it: an
+        # error panel quotes a driver, and ssh's stderr quotes whatever a
+        # config file pointed it at
+        hide_secrets_in(config, adapter_cls.ADAPTER_OPTIONS)
+
         # detect and install (if necessary) a tzdatabase on Windows
         if sys.platform == "win32" and not config.pop("no_download_tzdata", None):
             try:
@@ -643,9 +649,7 @@ def build_cli(argv: Sequence[str]) -> click.Command:
             )
             ctx.exit(2)
 
-        # off the config before the adapter is handed the rest of it: the
-        # connection details already name the local end of the forward, so an
-        # adapter is never told a tunnel exists
+        # off the config before the adapter is handed the rest of it
         try:
             ssh_config = take_ssh_keys(config, typed=explicitly_set)
         except HarlequinConfigError as e:
@@ -663,40 +667,42 @@ def build_cli(argv: Sequence[str]) -> click.Command:
             pretty_print_error(e)
             ctx.exit(2)
 
-        # before Textual owns the terminal, which is the only place a
-        # passphrase prompt or a 2FA push has anywhere to go
-        tunnel = _open_tunnel(ctx, ssh_config)
-
-        connection_id = (
-            adapter_instance.connection_id
-            if adapter_instance.connection_id is not None
-            else get_connection_hash(conn_str, config)
-        )
-        if tunnel is not None:
-            connection_id = get_connection_hash(
-                (connection_id,), {}, through=tunnel.cache_material()
-            )
-
-        tui = Harlequin(
-            adapter=adapter_instance,
-            profile_name=profile,
-            keymap_names=keymap_names,
-            user_defined_keymaps=user_defined_keymaps,
-            connection_hash=connection_id,
-            viewer_max_rows=viewer_max_rows,
-            query_limit=query_limit,
-            theme=theme,
-            show_files=show_files,
-            show_s3=show_s3,
-            export_path=export_path,
-            ssh_tunnel=tunnel,
-        )
         with contextlib.ExitStack() as stack:
-            if tunnel is not None:
+            if ssh_config.get("ssh_host"):
                 from harlequin.ssh import stopping_on_signal
 
-                stack.callback(tunnel.stop)
+                # entered before the child exists: `start()` blocks for the
+                # whole handshake, and a signal caught in that window would
+                # otherwise leave `ssh` running
                 stack.enter_context(stopping_on_signal())
+            tunnel = _open_tunnel(ctx, ssh_config)
+            if tunnel is not None:
+                stack.callback(tunnel.stop)
+
+            connection_id = (
+                adapter_instance.connection_id
+                if adapter_instance.connection_id is not None
+                else get_connection_hash(conn_str, config)
+            )
+            if tunnel is not None:
+                connection_id = get_connection_hash(
+                    (connection_id,), {}, through=tunnel.cache_material()
+                )
+
+            tui = Harlequin(
+                adapter=adapter_instance,
+                profile_name=profile,
+                keymap_names=keymap_names,
+                user_defined_keymaps=user_defined_keymaps,
+                connection_hash=connection_id,
+                viewer_max_rows=viewer_max_rows,
+                query_limit=query_limit,
+                theme=theme,
+                show_files=show_files,
+                show_s3=show_s3,
+                export_path=export_path,
+                ssh_tunnel=tunnel,
+            )
             tui.run()
 
     # this command's own options, before any adapter's are added to it
--- a/src/harlequin/cli.py
+++ b/src/harlequin/cli.py
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
+import contextlib
 import sys
 import warnings
 from functools import lru_cache
 from importlib.metadata import entry_points, version
 from pathlib import Path
-from typing import Any, Sequence
+from typing import TYPE_CHECKING, Any, Sequence
 
 import rich_click as click
 from rich_click.utils import OptionGroupDict
@@ -16,16 +17,19 @@ from harlequin.catalog_cache import get_connection_hash
 from harlequin.colors import GREEN, PINK, PURPLE, VALID_THEMES, YELLOW
 from harlequin.config import (
     DEFAULT_ADAPTER,
+    DEFAULT_SSH_TIMEOUT,
     Profile,
     load_profile_and_keymaps,
     merge_profile_with_cli,
     parse_profile_options,
     parse_row_count,
+    take_ssh_keys,
 )
 from harlequin.config_wizard import wizard
 from harlequin.exception import (
     HarlequinConfigError,
     HarlequinLocaleError,
+    HarlequinSshError,
     HarlequinTzDataError,
     pretty_print_error,
 )
@@ -35,6 +39,9 @@ from harlequin.locale_manager import set_locale
 from harlequin.options import AbstractOption
 from harlequin.plugins import adapter_names, load_adapter, load_adapter_plugins
 from harlequin.windows_timezone import check_and_install_tzdata
+
+if TYPE_CHECKING:
+    from harlequin.ssh import SshTunnel
 
 # configure defaults
 DEFAULT_VIEWER_MAX_ROWS = 100_000
@@ -139,6 +146,16 @@ HARLEQUIN_OPTION_GROUPS: list[OptionGroupDict] = [
             "--config-path",
             "--locale",
             "--no-download-tzdata",
+        ],
+    },
+    {
+        "name": "SSH Tunnel Options",
+        "options": [
+            "--ssh-host",
+            "--ssh-forward",
+            "--ssh-batch-mode",
+            "--ssh-allow-reuse",
+            "--ssh-timeout",
         ],
     },
     {
@@ -262,6 +279,27 @@ def _adapter_option_group(
         "name": f"{name} Adapter Options",
         "options": [f"--{option.name}" for option in adapter_cls.ADAPTER_OPTIONS or []],
     }
+
+
+def _open_tunnel(ctx: click.Context, ssh_config: dict[str, Any]) -> "SshTunnel | None":
+    """The tunnel the IDE runs through, or exit having said why there is none.
+
+    Here rather than in the app: once Textual owns the terminal, `ssh` cannot
+    ask for a passphrase and a 2FA push has nowhere to print "check your phone".
+    """
+    if not ssh_config.get("ssh_host"):
+        # nothing to open, and so nothing to import
+        return None
+    from harlequin.ssh import open_tunnel
+
+    try:
+        return open_tunnel(ssh_config)
+    except HarlequinConfigError as e:
+        pretty_print_error(e)
+        ctx.exit(2)
+    except HarlequinSshError as e:
+        pretty_print_error(e)
+        ctx.exit(3)
 
 
 def build_cli(argv: Sequence[str]) -> click.Command:
@@ -427,6 +465,47 @@ def build_cli(argv: Sequence[str]) -> click.Command:
         default=DEFAULT_KEYMAP_NAMES,
     )
     @click.option(
+        "--ssh-host",
+        help=(
+            "Open an SSH tunnel to this destination first, and connect through "
+            "it. A Host alias, host, user@host or ssh://user@host:port, passed "
+            "to ssh verbatim."
+        ),
+    )
+    @click.option(
+        "--ssh-forward",
+        multiple=True,
+        help=(
+            "A local forward, spelled as ssh -L takes one: LOCAL:HOST:REMOTE. "
+            "Repeat this option for more than one. Omit it when your ssh config "
+            "already has the LocalForward."
+        ),
+    )
+    @click.option(
+        "--ssh-batch-mode",
+        is_flag=True,
+        help=(
+            "Fail rather than prompt for a passphrase, a password or a host "
+            "key. ssh's own BatchMode."
+        ),
+    )
+    @click.option(
+        "--ssh-allow-reuse",
+        is_flag=True,
+        help=(
+            "When the local port is already bound, warn and connect through "
+            "the listener that has it instead of failing."
+        ),
+    )
+    @click.option(
+        "--ssh-timeout",
+        type=click.FloatRange(min=0, min_open=True),
+        help=(
+            "Seconds to wait for the tunnel's forwards. Default is "
+            f"{DEFAULT_SSH_TIMEOUT:g}"
+        ),
+    )
+    @click.option(
         "--config",
         help=(
             "Run the configuration wizard to create or update a Harlequin config file."
@@ -564,6 +643,15 @@ def build_cli(argv: Sequence[str]) -> click.Command:
             )
             ctx.exit(2)
 
+        # off the config before the adapter is handed the rest of it: the
+        # connection details already name the local end of the forward, so an
+        # adapter is never told a tunnel exists
+        try:
+            ssh_config = take_ssh_keys(config)
+        except HarlequinConfigError as e:
+            pretty_print_error(e)
+            ctx.exit(2)
+
         # instantiate the adapter, which was named and imported above -- the
         # key comes off either way, because what is left is its options
         config.pop("adapter", None)
@@ -575,11 +663,19 @@ def build_cli(argv: Sequence[str]) -> click.Command:
             pretty_print_error(e)
             ctx.exit(2)
 
+        # before Textual owns the terminal, which is the only place a
+        # passphrase prompt or a 2FA push has anywhere to go
+        tunnel = _open_tunnel(ctx, ssh_config)
+
         connection_id = (
             adapter_instance.connection_id
             if adapter_instance.connection_id is not None
             else get_connection_hash(conn_str, config)
         )
+        if tunnel is not None:
+            connection_id = get_connection_hash(
+                (connection_id,), {}, through=tunnel.cache_material()
+            )
 
         tui = Harlequin(
             adapter=adapter_instance,
@@ -593,8 +689,12 @@ def build_cli(argv: Sequence[str]) -> click.Command:
             show_files=show_files,
             show_s3=show_s3,
             export_path=export_path,
+            ssh_tunnel=tunnel,
         )
-        tui.run()
+        with contextlib.ExitStack() as stack:
+            if tunnel is not None:
+                stack.callback(tunnel.stop)
+            tui.run()
 
     # this command's own options, before any adapter's are added to it
     harlequin_options = {param.name for param in inner_cli.params}
--- a/src/harlequin/config.py
+++ b/src/harlequin/config.py
@@ -95,6 +95,22 @@ caller asks a database what a query's columns are, and an option that spent
 that spelling on "unlimited" would take the idiom away.
 """
 
+DEFAULT_SSH_TIMEOUT = 10.0
+"""Seconds both commands wait for an SSH tunnel's forwards, by default."""
+
+SSH_KEYS = (
+    "ssh_host",
+    "ssh_forward",
+    "ssh_batch_mode",
+    "ssh_allow_reuse",
+    "ssh_timeout",
+)
+"""The profile keys that describe an SSH tunnel, which both commands read.
+
+Named here rather than in `harlequin.ssh` so that an invocation with no tunnel
+can take them off a config without importing the module that opens one.
+"""
+
 TUI_ONLY_KEYS = (
     "theme",
     "keymap_name",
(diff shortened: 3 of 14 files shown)


src/harlequin/config.py
```diff
@@ -95,6 +95,37 @@
...
+    "ssh_forward",
+    "ssh_batch_mode",
+    "ssh_allow_reuse",
+    "ssh_timeout",
+)
+"""The profile keys that describe an SSH tunnel, which both commands read.
+
+Named here rather than in `harlequin.ssh` so that an invocation with no tunnel
+can take them off a config without importing the module that opens one.
+"""
```
tconbeer, line 118: **Security — the `ssh_allow_reuse` guard is bypassable from a config file via `ssh_timeout`.**

The direct route is airtight: I could not get `ssh_allow_reuse` in from a config file, an env var, or `merge_profile_with_cli` in either command. But the *effect* the key gates — "connect through a listener Harlequin did not open" — is reachable from a config file through `ssh_timeout`, which is not in `CLI_ONLY_SSH_KEYS`.

`build_tunnel()` passes the same `timeout` to the `ssh -G` probe (`resolve_config(argv, timeout=timeout)`), and `resolve_config()` returns `None` on `TimeoutExpired`. `None` means no `forwards`, so `SshTunnel.endpoints` is empty, and `_wait_until_ready()`'s empty-endpoints branch sleeps `min(1.0, self.timeout)` and returns success without probing anything. With `ssh_timeout = 0.001` that sleep is 1 ms, so the child has not exited yet and the already-bound check never fires.

A/B against an unrelated process already listening on 127.0.0.1:15432:

```
ssh_timeout normal   -> error: 127.0.0.1:15432 was already bound ... pass --ssh-allow-reuse    exit 3
ssh_timeout = 0.001  -> note: ssh: tunnel to bastion                                            exit 0, query ran
```

The notice printed is the innocuous `describe()` line, not the reuse warning, so nothing tells the user.

The second way to the same state needs no race: `_split()` returns `None` for any field containing `/`, so a forward that is entirely unix sockets (`ssh_forward = ["/tmp/z.sock:x:1"]`) also yields empty `endpoints` and the same unverified grace path.

Root cause for both: **`_wait_until_ready()` treats "nothing to poll" as success, and a config file decides whether there is anything to poll.** Worth making the degraded case fail closed, or at least refusing `--ssh-allow-reuse`-equivalent outcomes when `resolve_config()` returned `None`.

---

src/harlequin/ssh.py
```diff
@@ -115,6 +114,104 @@ class SshConfig:
...
+    Neither signal runs an `atexit` handler, so a session that is killed, or
+    whose terminal goes away, would otherwise leave `ssh` holding a forward
+    open. The tunnels are stopped from the handler rather than by the exception
+    it raises, which something further up may catch.
+
+    A no-op off the main thread, and on a platform without the signal.
+    """
+    replaced: list[tuple[int, Any]] = []
+
+    def handle(number: int, frame: Any) -> None:
```
tconbeer, line 207: **Correctness — `raise SystemExit` here reintroduces the exit-134 abort that `hsql/timeout.py` exists to prevent.**

`timeout.py`'s own module docstring says it: "`sys.exit()` with a live thread in duckdb's `to_arrow_table()` exits 134, which is not a code hsql documents" — which is why `_halt()` uses `os._exit()`. This handler unwinds the main thread with `SystemExit` while the `--timeout` worker is still inside the driver, and gets exactly that.

Measured, fake `ssh` on PATH, duckdb, profile with `timeout = 120`:

- `ssh_host` set → SIGTERM → **exit 134**, stderr `terminate called without an active exception`
- same profile without `ssh_host` (no tunnel, no handler) → SIGTERM → **exit 143**

134 is not in `ExitCode`. The tunnel *is* stopped, so only the exit code and the abort message are wrong — but a script wrapping `hsql` now sees an undocumented code on any SIGTERM/SIGHUP during a tunnelled, timed run. The same handler is installed around `tui.run()`, where Textual's workers are also in the driver.

Since `stop_all()` has already done the work by that point, the handler may not need to raise at all — or could route through the same `os._exit()` path `_halt()` uses.

Separately, at the restore on line 210: `signal.signal()` returns `None` when the previous handler was not installed from Python, and `signal.signal(number, None)` raises `TypeError`, which `suppress(ValueError, OSError)` does not catch:

```
>>> signal.signal(signal.SIGTERM, None)
TypeError: signal handler must be signal.SIG_IGN, signal.SIG_DFL, or a callable object
```

Rare trigger, but it escapes the `finally`.

---

src/harlequin/hsql/cli.py
```diff
@@ -615,6 +669,16 @@ def inner_cli(
...
 
+        tunnel = _open_tunnel(ctx, ssh_config)
+        if tunnel is not None:
+            # the run's, not the process's: click closes the context however
+            # this ends, and `harlequin.ssh` keeps an atexit backstop under that
+            from harlequin.ssh import stopping_on_signal
+
+            ctx.with_resource(stopping_on_signal())
+            ctx.call_on_close(tunnel.stop)
+            diagnostics.report_tunnel(tunnel.notice())
```
tconbeer: **Correctness — the signal handler is installed *after* the tunnel is started, so a signal during startup orphans the `ssh` child.**

`_open_tunnel()` spawns `ssh` and then blocks in `_wait_until_ready()` for up to `--ssh-timeout` — and in practice for the whole passphrase / 2FA / host-key wait. `stopping_on_signal()` is only entered once that returns, and `atexit` does not run on SIGTERM/SIGHUP.

Measured with a fake `ssh` that delays binding and `ssh_timeout = 60`, SIGTERM 4s in: hsql exits 143 and `ssh -N -o ExitOnForwardFailure=yes bastion` is **still running afterwards**. That is the exact leak this feature is for, in the window where a user is most likely to give up and hit Ctrl-C or kill the process.

`SshTunnel.start()` already adds to `_LIVE` before it waits, so entering `stopping_on_signal()` *before* `_open_tunnel()` closes the window with no other change.

`src/harlequin/cli.py:668` has the same shape.

While here: the IDE's `ExitStack` unwinds in the opposite order from this one. `stack.callback(tunnel.stop)` is registered first and so runs **last**, after `stopping_on_signal()` has already restored `SIG_DFL` — so the up-to-2.5s `tunnel.stop()` runs unprotected. `ctx.with_resource(...)` then `ctx.call_on_close(tunnel.stop)` here gets it right; small, but the two should probably match.

---

src/harlequin/ssh.py
```diff
@@ -115,6 +114,104 @@ class SshConfig:
...
+        Raises: HarlequinConfigError for a value its key does not take. A
+        profile can put anything under one, and click has already vetted
+        whatever was typed on the command line.
+        """
+        forwards = config.get("ssh_forward") or ()
+        if isinstance(forwards, str):
+            forwards = (forwards,)
+        return cls(
+            host=_text(config.get("ssh_host"), key="ssh_host"),
+            forwards=tuple(
```
tconbeer: **Correctness — a non-string, non-list `ssh_forward` in a profile raises `TypeError`, not `HarlequinConfigError`.**

Only `str` is wrapped, so any other non-iterable is iterated directly. `Profile = Dict[str, Any]` and `ssh_forward` is a command option, so `parse_profile_options()` never type-checks it, and nothing upstream catches this:

```toml
[profiles.prod]
adapter = "sqlite"
ssh_host = "bastion"
ssh_forward = 5432          # or true, or 1.5
```

`hsql -P prod -c "select 1"` → full traceback, `TypeError: 'int' object is not iterable`, uncaught out of `main()` (`hsql/__init__.py` catches only `ClickException` / `Abort` / `KeyboardInterrupt`). The docstring above promises "Raises: HarlequinConfigError for a value its key does not take", and `_open_tunnel()` only catches `HarlequinConfigError` / `HarlequinSshError`.

A TOML *table* is worse — it iterates silently and uses the keys as forward specs: `{"5432:db:5432" = 1}` → `('5432:db:5432',)`.

`test_a_destination_that_is_not_text_is_refused` covers `ssh_host` only; the same parametrization over `ssh_forward` would have caught this.

Also on line 138: `_text(forward, key="ssh_forward") or ""` turns the `None` that `_text()` returns for `""` straight back into `""`, so an empty entry survives into the argv as `-L ""`:

```python
build_argv('bastion', ('', '15432:db:5432'))
# ['ssh','-N','-o','ExitOnForwardFailure=yes','-L','','-L','15432:db:5432','bastion']
```

`ssh` fatals on that, so `resolve_config()` returns non-zero → degrades to `None` → the "configures no local forward" check is skipped too, and the run dies at `start()` quoting ssh instead of naming the bad entry.

---

src/harlequin/schemas/config-v1.json
```diff
@@ -184,6 +184,38 @@
...
+        "ssh_batch_mode": {
+          "type": "boolean",
+          "description": "Fail rather than prompt for a passphrase, a password or a host key. ssh's own BatchMode; set it in scripts, CI and cron.",
+          "default": false
+        },
+        "ssh_allow_reuse": {
+          "type": "boolean",
+          "description": "When the local port is already bound, warn and connect through the listener that has it instead of failing.",
+          "default": false
+        },
```
tconbeer: **Correctness — the schema advertises a profile key the runtime refuses, and `--config init` writes it.**

`ssh_allow_reuse` lands under `$defs/profile/properties` here, with a default, so a schema-aware editor autocompletes it and validates it as fine. `take_ssh_keys()` then exits 2 on it in both commands.

`--config init` completes the loop — it writes a file it will not read back:

```
hsql --config init --config-path cfg.toml -P prod -a sqlite \
     --ssh-host bastion --ssh-forward 15432:db:5432 --ssh-allow-reuse ':memory:'
# note: wrote the profile named prod ...        exit 0, file contains ssh_allow_reuse = true

hsql --config-path cfg.toml -P prod -c "select 1"
# hsql: error: ssh_allow_reuse turns off the check ... Pass --ssh-allow-reuse.    exit 2
harlequin --config-path cfg.toml -P prod
# same error, exit 2
```

`hsql --config validate --config-path cfg.toml` reports **0 problems** for that file, so nothing points at it either.

Three places to line up with `CLI_ONLY_SSH_KEYS`: `_profile()` in `config_schema.py` should drop the key from `properties` (`TUI_ONLY_KEYS` is already handled a few lines above, so there's a place for it), `_typed_profile_keys()` should not write it, and ideally `--config validate` should flag it.

---

src/harlequin/cli.py
```diff
@@ -426,6 +464,47 @@ def _adapter_class(name: str) -> type[HarlequinAdapter]:
...
+    )
+    @click.option(
+        "--ssh-forward",
+        multiple=True,
+        help=(
+            "A local forward, spelled as ssh -L takes one: LOCAL:HOST:REMOTE. "
+            "Repeat this option for more than one. Omit it when your ssh config "
+            "already has the LocalForward."
+        ),
+    )
```
tconbeer, line 484: **Security — this is the diff that makes a CWD-discovered config file able to hand a forward spec to `ssh`, and `_UNSAFE_IN_VALUE` does not refuse `*` or `/`.**

`_refuse_unsafe_value()` landed in #1112 with no caller; this PR is what connects `pyproject.toml` → `build_argv()` → `Popen`. Confirmed: a `pyproject.toml` in a cloned repo with `ssh_host` set spawns `ssh` to that destination — authenticating with the user's agent and keys, prompting on their terminal — before any query runs. (Good: `--info`, `--spec`, `--config show` and `--version` open nothing; I measured 0 ssh invocations for each. And leading `-` *is* refused, so no option injection.)

Two characters that survive the filter and change what `ssh` binds:

```python
>>> build_argv("h", ["*:15432:internal:5432"])       # accepted
>>> build_argv("h", ["/home/user/pwned.sock:internal:5432"])   # accepted
```

- `*` — per `ssh(1)`, "an empty address or `*` indicates that the port should be available from all interfaces", and an explicit bind address overrides `GatewayPorts`. So a hostile repo config turns the user's laptop into an open proxy into whatever network the bastion fronts. Partly mitigated by `_pretty()` reporting the bind address (`*:15432 -> db:5432 via bastion` renders correctly) — but in the IDE that is a transient toast at `app.py:359` with `severity="information"`.
- `/` — picks an arbitrary absolute path for `ssh` to create a unix socket at. Reading OpenSSH, `-L` local sockets are not unlinked first, so an existing file should hit `EADDRINUSE` rather than being clobbered; I could not verify that here, so treat "cannot destroy existing data" as the unverified half.

`_ANY_INTERFACE` already exists for the polling side, so the concept is in the module; it just isn't a refusal.

---

src/harlequin/cli.py
```diff
@@ -264,6 +281,27 @@ def _adapter_option_group(
...
+        # nothing to open, and so nothing to import
+        return None
+    from harlequin.ssh import open_tunnel
+
+    try:
+        return open_tunnel(ssh_config)
+    except HarlequinConfigError as e:
+        pretty_print_error(e)
+        ctx.exit(2)
+    except HarlequinSshError as e:
```
tconbeer, line 301: **Security — the IDE does not redact, and this path now carries up to 8 KB of third-party-controlled text into it.**

`_child_failed()` / `_timed_out()` embed `_Stderr.text()` — 8192 bytes, unsanitised — in the error. An SSH server controls its pre-auth banner, `ssh` writes it to stderr verbatim, and (per the comment above) a config file gets to pick the server. Escapes survive in both front ends:

```
hsql: ^[]52;c;cHduZWQ=^G^[[2J^[[31mBANNER FROM SERVER^[[0m     (intact, incl. OSC 52 + BEL)
IDE : ^[]52;c;cHduZWQ=^[[2J^[[31mBANNER FROM SERVER^[[0m       (rich drops BEL, keeps ESC/CSI/OSC)
```

On the redaction half: **`hsql` is correct.** `hide_secrets_in()` runs at `hsql/cli.py:494`, before `take_ssh_keys()` at 515 and before the tunnel opens at 672, and `diagnostics._write()` redacts every line — verified that a password in `conn_str` echoed back by ssh comes out as `postgres://u:********@h/db`.

The IDE has no equivalent: `hide_secrets_in()` is called nowhere in `harlequin/cli.py`, and `pretty_print_error()` does not redact, so the same run prints `postgres://u:hunter2@h/db` verbatim in the error panel. That gap is pre-existing — what this PR adds is a new, large, attacker-influenced string flowing into it. Might be worth a `hide_secrets_in()` call on the IDE side, or at least tracking it separately.

---

src/harlequin/ssh.py
```diff
@@ -274,6 +374,30 @@ def describe(self) -> str:
...
+        Harlequin opened.
+        """
+        if self.reused:
+            listed = ", ".join(f"{host}:{port}" for host, port in self.endpoints)
+            return (
+                f"{listed} is already bound; connecting through the existing "
+                "listener (--ssh-allow-reuse)"
+            )
+        return self.describe()
+
```
tconbeer, line 445: The flattening itself is sound: `parse_config()` only accepts a `localforward` line with exactly two whitespace-separated fields, so neither half can contain a space and `f"{listen} {destination}"` cannot be ambiguous. Two bastions, two forwards, two keys — verified. The re-hash at `cli.py:676` is stable, and an untunnelled connection hashes byte-identically to the old implementation (`old(cs, cfg) == new(cs, cfg) == new(cs, cfg, through=())`).

The gap is the degraded path from the `CLI_ONLY_SSH_KEYS` comment: when `resolve_config()` returns `None`, `self.forwards` is empty and `cache_material()` collapses to `(self.host,)`. Two tunnels to the same host with different `--ssh-forward` then hash to the same key — I get `08d69cb8541fc3498bb8e887e74ae398` for both. One database's pickled catalog **and query history** served under the other's identity, which is the thing this method exists to prevent.

Falls out if the degraded case fails closed instead.

---

src/harlequin/config.py
```diff
@@ -638,6 +669,40 @@ def refuse() -> HarlequinConfigError:
...
+    for key in CLI_ONLY_SSH_KEYS:
+        if taken.get(key) and key not in typed:
+            spelled = key.replace("_", "-")
+            raise HarlequinConfigError(
+                f"{key} turns off the check that the local port is not already "
+                "in use by something else, so it is read from the command line "
+                f"and not from a config file. Pass --{spelled}.",
+                title=CONFIG_ERROR_TITLE,
+            )
+    if not taken.get("ssh_host") and any(taken.values()):
```
tconbeer: `any(taken.values())` misses every falsy tunnel key, so with no `ssh_host` these are dropped silently rather than reported: `ssh_timeout = 0`, `ssh_batch_mode = false`, `ssh_forward = []`, `ssh_forward = ""`. All end in "no tunnel", which is arguably the right outcome, so this is a missing diagnostic rather than a wrong one — `key in taken` would be the stricter read of "any ssh key set with no destination".

The inverse is more confusing: `ssh_timeout = -5` alone *does* raise, but with "`ssh_timeout` describes an SSH tunnel, but ssh_host says nothing to open one to" rather than the "not a positive number of seconds" that `parse_seconds()` would have said — the destination check runs before anything types the value.

---

src/harlequin/hsql/timeout.py
```diff
@@ -118,6 +118,12 @@ def _cancel(self, connection: HarlequinConnection) -> None:
 
 def _halt(code: ExitCode) -> NoReturn:
     """End the process now, around a worker thread that would abort it."""
+    with contextlib.suppress(Exception):
+        # `os._exit()` runs no atexit handler, and an ssh child that outlives
+        # the run is the thing the tunnel feature exists to remove
+        from harlequin.ssh import stop_all
+
+        stop_all()
```
tconbeer: Right call to reach the tunnel from here, but note what it costs the function whose docstring is "End the process now": `stop_all()` → `stop()` can spend `terminate()` + `wait(2.0)` + `kill()` + `wait(2.0)`, plus `_Stderr.close()`'s 0.5s join — up to ~4.5s per tunnel, all *before* the stdout/stderr flush.

Also `contextlib.suppress(Exception)` will not contain a `KeyboardInterrupt` raised inside `stop_all()`, which on this path would skip the flush entirely.

---

src/harlequin/app.py
```diff
@@ -347,6 +352,15 @@ def append_to_history(
     async def on_mount(self) -> None:
         self.run_query_bar.apply_configured_limit()
 
+        if self.ssh_tunnel is not None:
+            # which database this session is actually looking at, which the
+            # connection details alone no longer say
```
tconbeer: **Comments** — "no longer say" is a live distinction that stops being visible the moment this merges; `AGENTS.md` calls out exactly this ("Never reference code that was removed or refactored"). The reader after merge has no "previously" to compare against. Just:

```python
# which database this session is actually looking at
```

Same shape at `catalog_cache.py:67`: "so an untunneled connection keys the same as it always has" states a compatibility fact relative to the old implementation. The invariant is worth keeping, but phrased forward — "an untunneled connection keys on its details alone."

---

src/harlequin/cli.py
```diff
@@ -264,6 +281,27 @@ def _adapter_option_group(
     }
 
 
+def _open_tunnel(ctx: click.Context, ssh_config: dict[str, Any]) -> "SshTunnel | None":
+    """The tunnel the IDE runs through, or exit having said why there is none.
+
+    Here rather than in the app: once Textual owns the terminal, `ssh` cannot
+    ask for a passphrase and a 2FA push has nowhere to print "check your phone".
+    """
```
tconbeer, line 290: **Comments** — "Here rather than in the app" names the alternative that was rejected, which `AGENTS.md` puts in the PR description rather than the source. The reason on its own is the whole of what the next reader needs:

```
Once Textual owns the terminal, `ssh` cannot ask for a passphrase and a 2FA
push has nowhere to print "check your phone".
```

Then line 666 repeats it a third time ("before Textual owns the terminal, which is the only place a passphrase prompt or a 2FA push has anywhere to go") twelve lines from the call it annotates — that one can go entirely, since the function it calls now says it.

Two more duplications worth collapsing to one statement each:

- "an adapter is never told a tunnel exists" appears in `take_ssh_keys()`'s docstring, at `cli.py:643`, and at `hsql/cli.py:513`. The docstring is the right home; the two call sites can just say what they take off.
- "an ssh child that outlives the run is the thing this feature exists to remove" is in `stop_all()`'s docstring and again at `timeout.py:123`. At the call site the load-bearing half is only `os._exit()` runs no atexit handler.

And `ssh.py:257` — "a usage error rather than a failure to connect: nothing has been run yet, and what is wrong is what was asked for" — records the reasoning for picking the exception class, which the `Raises:` line three lines above already states as fact. That is PR-description material.

---

src/harlequin/ssh.py
```diff
@@ -115,6 +114,104 @@ class SshConfig:
...
+        whatever was typed on the command line.
+        """
+        forwards = config.get("ssh_forward") or ()
+        if isinstance(forwards, str):
+            forwards = (forwards,)
+        return cls(
+            host=_text(config.get("ssh_host"), key="ssh_host"),
+            forwards=tuple(
+                _text(forward, key="ssh_forward") or "" for forward in forwards
+            ),
```
tconbeer: **Reuse / consistency** — the three key types disagree about how much they trust a profile. `ssh_timeout` is type-checked through `parse_seconds()`; `ssh_host` and `ssh_forward` through `_text()`; but the two flags go through bare `bool()`, so a profile written as `ssh_batch_mode = "false"` silently means **true**:

```python
>>> SshOptions.from_config({'ssh_host': 'h', 'ssh_batch_mode': 'false'}).batch_mode
True
```

`ssh_allow_reuse = "false"` likewise trips the CLI-only refusal. `read_only` elsewhere in `hsql/cli.py` is handled the same way, so this is an existing convention rather than a new problem — but this is the one struct where a strict path sits next to a permissive one, and the docstring promises "HarlequinConfigError for a value its key does not take".

---

tconbeer (commented):
## Scope

Reviewed the diff of `claude/ssh-tunnels-implementation-wxjvhs-2-options` against its base `...-1-tunnel` only — `harlequin/ssh.py` as it landed in #1112 was read for context but not reviewed. Four lenses, run as separate passes: **correctness/logic**, **security** (unexpected access to or destruction of user data), **reuse & simplification**, and **comments** against `AGENTS.md`. CI is green on all 32 checks; `tests/unit_tests/test_ssh.py` at head is 70 passed, 1 skipped.

The design is sound and the interesting decisions hold up under test: the adapter really is never told a tunnel exists, informational modes (`--info`, `--spec`, `--config show`, `--version`) open nothing, the untunnelled cache key is byte-identical to before, `hsql`'s stderr redaction ordering is correct, and `--ssh-allow-reuse` genuinely cannot be set from a config file, an env var, or `merge_profile_with_cli`. Twelve inline comments; the four that would change my merge answer:

1. **`ssh_timeout` defeats the `--ssh-allow-reuse` guard** (`config.py:114`). It isn't in `CLI_ONLY_SSH_KEYS`, and it's the same timeout as the `ssh -G` probe — `ssh_timeout = 0.001` degrades `resolve_config()` to `None`, empties `endpoints`, and `_wait_until_ready()` declares success without probing. A/B measured: normal timeout refuses the bound port and exits 3; tiny timeout connects through the foreign listener and exits 0, printing the innocuous notice. An all-unix-socket forward reaches the same state with no race. Root cause is shared: **"nothing to poll" is treated as success, and a config file decides whether there is anything to poll.**
2. **`raise SystemExit` in the signal handler reintroduces exit 134** (`ssh.py:197`) — the exact abort `hsql/timeout.py`'s docstring exists to document. Measured 134 with a tunnel, 143 without.
3. **The handler is installed after the tunnel starts** (`hsql/cli.py:672`, `cli.py:668`), so SIGTERM during the passphrase wait orphans `ssh` — the leak the feature is for, in its likeliest window. Measured.
4. **A non-string `ssh_forward` in a profile is an uncaught `TypeError`** (`ssh.py:133`), and the JSON schema plus `--config init` between them lead a user to write an `ssh_allow_reuse` profile that every later run exits 2 on.

The rest are lower: `*` and `/` unrefused in a forward spec, unredacted ssh stderr reaching the IDE's error panel, `cache_material()` collapsing in the degraded path, `bool()` coercion, `_halt()`'s new latency, and the comment trims.

## Not checked

- **Behavior against a real `ssh` client.** Every run used a fake `ssh` on `PATH` (no ssh binary in the container). Not observed: whether `-L /path:h:p` unlinks an existing file, whether `*:port` actually binds all interfaces, whether a real server's banner reaches stderr with escapes intact, and the exact fatal for `-L ""`. Mechanisms are confirmed; the OpenSSH side is read from `ssh(1)` and source.
- **The IDE end to end.** Confirmed `harlequin` reads the config keys, refuses `ssh_allow_reuse`, and spawns `ssh` before Textual starts; did not drive the TUI to a successful connection, and `SystemExit` from the handler inside Textual's event loop is inference from the `hsql` result, not measured. `Harlequin.on_mount`'s notification was not seen rendered.
- **Windows and macOS.** `SIGHUP`'s absence, `_client_path()`'s `PATHEXT` branch, and `resolve_config()` timeouts were not exercised; CI green on those platforms is what I'm relying on.
- **`harlequin/ssh.py` as it stands in #1112** — `_refuse_unsafe_value`, `_UNSAFE_IN_VALUE`, `parse_config`, `_wait_until_ready` and `_Stderr` were read only where this diff makes them newly reachable. Review #1112 for those on their own terms.
- **The full test suite and the snapshot tests.** Only `tests/unit_tests/test_ssh.py` was run; the generator-vs-artifact tests for `config-v1.json` and `hsql-reference.md` were not, so I checked the schema's contents by hand rather than by regenerating.
- **`get_connection_hash()`'s other callers.** Verified the two in `cli.py` and the hash's own properties; did not audit every consumer of a connection hash for the assumption that it is derived from connection details alone.
- **Adapters beyond duckdb and sqlite.** The "no `ssh_*` key reaches the adapter" claim was checked against a spy adapter and the two in-tree ones, not against `harlequin-postgres` or any out-of-tree adapter.
- **The `--ssh-timeout`/`ssh_timeout` interaction with `--timeout`,** and whether a tunnel that dies mid-session is noticed — 3/5 of the stack, out of scope here.

---

tconbeer:
Thanks — the `ssh_timeout` bypass is real and is fixed, along with everything else here bar one, which I've left open below. Answering in one comment rather than thirteen.

## The bypass — `ssh_timeout` reaching the `ssh -G` probe

You found the right root cause: `_wait_until_ready()` treated "nothing to poll" as success, and a config file decided whether there was anything to poll. Two changes, and the second is the one that closes the class rather than the instance.

- **`ssh -G` has its own bound now** (`_PROBE_SECONDS`), not the caller's. `-G` connects to nothing, so a wait configured for a network or for a person answering a prompt never said anything about it. `ssh_timeout = 0.001` no longer reaches the probe at all.
- **A probe that ran and ssh rejected is fatal**, quoting ssh, instead of degrading to the same unverified path. This was the other way in — `ssh_forward = ["!"]` makes `-G` exit non-zero just as surely as a small timeout made it expire. It is the same argv `start()` is about to run, so the run ended either way; now it ends with ssh naming the value that is wrong, before anything is spawned. `None` still means "no client, or output this cannot read", which is the tolerance the degrade exists for.

The unix-socket route stays, because a socket-only forward genuinely has nothing to poll — but it is no longer silent: see the notice below.

## Correctness

- **`raise SystemExit` → exit 134.** Fixed. The handler stops the tunnels, flushes, restores `SIG_DFL` and re-raises the signal to itself, so a tunnelled run killed with SIGTERM exits 143 like an untunnelled one, with no unwinding past a worker inside a driver.
- **The restore that raises `TypeError`.** Fixed — a `previous` of `None` is skipped rather than handed back to `signal.signal()`, and `TypeError` joins the suppressed set as a backstop.
- **The handler installed after the tunnel starts.** Fixed in both commands: `stopping_on_signal()` is entered before `_open_tunnel()`, keyed off `ssh_host` being set so a run without a tunnel still changes no disposition. The IDE's `ExitStack` now matches hsql's order, so `tunnel.stop()` runs while the handler is still installed.
- **`ssh_forward` that is not a list of strings.** Fixed — `_forwards()` refuses a non-list, a non-string entry, and an entry with nothing in it, all as `HarlequinConfigError`. A table no longer iterates silently into its keys, and `-L ""` never reaches the argv.
- **`bool()` on the two flags.** Fixed: `_flag()` takes a boolean or nothing. `ssh_batch_mode = "false"` is refused rather than read as true.
- **`any(taken.values())`** is `key in taken`, so a key that is set but falsy is named; and `ssh_timeout` is typed before the destination check, so `ssh_timeout = -5` says "not a positive number of seconds" rather than "ssh_host says nothing to open one to".
- **`_halt()`** flushes stdout and stderr *before* stopping tunnels, which can spend seconds, and suppresses `BaseException` rather than `Exception` so a `KeyboardInterrupt` inside `stop_all()` cannot skip the exit.
- **`cache_material()`** falls back to the specs a caller asked for where `ssh -G` did not say what was resolved, so two tunnels to the same host with different forwards no longer key one database's catalog and history under the other's name.

## Security

- **The IDE's redaction, and the escapes.** Both fixed, in the module rather than at the front ends. `_Output.text()` runs ssh's bytes through a new `_readable()` — which drops every control character a terminal acts on, OSC 52 and the CSI included — and then through `redact_text()`, so both front ends get the same string. The IDE also calls `hide_secrets_in()` now, right after the config is merged, so `_HIDDEN` is populated there rather than only under hsql.
- **The schema and `--config init`.** Fixed: `CLI_ONLY_SSH_KEYS` comes out of `$defs/profile/properties` and out of what `init` writes, so the document no longer advertises a key the next run exits 2 on. I did not add a `--config validate` check — `Profile` is `Dict[str, Any]` and validate is a msgspec shape pass, so flagging it means teaching that pass about a rule only the commands know. Worth doing, but as its own change.

## What I did not close

**`*` and `/` in a forward spec** — thread left open, because I think the call is yours rather than mine.

Refusing `*` contradicts §8, which rejects a configurable bind address on the grounds that "`ssh_config` and the forward spec already do it; `0.0.0.0` should take deliberate effort" — the forward spec being able to say it is the design's answer, and `_UNSAFE_IN_VALUE` refusing `*` while permitting `0.0.0.0` would be an arbitrary half of it. Refusing `/` bans unix-socket forwards outright, which the postgres and mysql adapters can legitimately connect over.

So instead the exposure is now impossible to miss rather than impossible: `SshTunnel.exposed` reads the bind address off `ssh -G`'s own parse — catching `*`, `0.0.0.0`, `::` and any non-loopback literal, from a `Host` block as well as from a flag — and `warnings()` says "`*:15439` is bound to every interface, so anything that can reach this machine can reach the database through it." hsql prints it on stderr beside the notice; the IDE's toast goes to `severity="warning"`. The degraded case says its own line there too: "ssh did not say what it forwards, so Harlequin did not wait for a local port to answer before connecting."

If you would rather refuse the wildcard outright, that is a two-line change and I am happy to make it — it just needs §8 to change with it.

## Comments

All applied: the `app.py` and `catalog_cache.py` "no longer say" / "as it always has" phrasings, "Here rather than in the app", the third copy of the passphrase reason at `cli.py:666`, the two remaining copies of "an adapter is never told a tunnel exists", the second copy of "an ssh child that outlives the run", and `ssh.py:257`'s exception-class rationale.

---

Validated on each branch: 1476 tests pass on the tip, plus mypy, ruff and the four import-linter contracts. Merged forward through 3/6–6/6, and the stack now sits on `main` at 8187f0f.

---

PR #1114, opened 2026-09-01
SSH tunnels 3/5: keep the tunnel open, and say when it is not

tconbeer:
Part 3 of 5 of `docs/plans/ssh-tunnels-design.md`. **Based on #1113**; review the stack in order.

Stack: 1/5 tunnel → 2/5 options → **3/5 (this)** → 4/5 reconnect → 5/5 changelog.

Part of #545

Three additions, all §5.4 and §4 of the design.

- **A keepalive, where nothing set one.** The `ssh -G` probe already reports the resolved `ServerAliveInterval`; when it is `0`, `build_tunnel()` rebuilds the argv with `-o ServerAliveInterval=30`, and otherwise leaves it alone.
- **A death notification.** `SshTunnel.watch(on_close)` starts a thread that waits on the child and reports `tunnel closed: <ssh's last line>` if it exits on its own. The IDE arms it on mount and posts a `TunnelClosed` message; a handler notifies.
- **Two places report the tunnel.** `hsql --info` gains an `"ssh"` field naming the client and its version, and the IDE's debug screen gains an "SSH Tunnel" line with the forwards this session is reached through.

- **A keepalive with no flag.** `ssh` sends none by default, so an idle forward behind a NAT or a corporate firewall is reaped silently — and a tunnel that drops after lunch is the difference between a feature people trust and one people work around. It is not configurable because an interval is not a thing anyone should have to tune, and the one person who wants to has a `Host` block. A command-line `-o` beats the config file, so imposing it unconditionally would be Harlequin retuning someone else's connection; hence the probe rather than a blanket default. `--ssh-keepalive SECONDS` is deferred in §8 with the case that brings it back.
- **A dropped forward would otherwise be an unexplained wall of query errors**, and the front end is the only thing that can say what happened. `stop()` sets a flag first, so a tunnel taken down deliberately is not reported — that is not news.
- **The notification is posted, not called.** The watcher runs on its own thread, so it does the one thing a thread may do here: `post_message`. The handler notifies.
- **`hsql --info` is the diagnostic an agent runs first**, and the tunnel options are the one part of hsql that depends on a program outside the wheel. `ssh -V` connects to nothing, which is `--info`'s whole promise.
- **The debug-screen widget is only added when there is a tunnel**, so the existing snapshot is unchanged.
- **Only the IDE watches.** `hsql` is short-lived; the adapter's error is the whole story there. Only the keepalive applies to both commands.

- [x] Yes.

A resolved `serveraliveinterval 0` adds the option and a non-zero one does not; a probe that says nothing readable imposes nothing; the fake client grows a `FAKE_SSH_LIFETIME` so a child can drop on its own, and the watcher reports ssh's last line while a deliberate `stop()` reports nothing.

---

### DIFF
--- a/src/harlequin/app.py
+++ b/src/harlequin/app.py
@@ -163,6 +163,14 @@ class ResultsFetched(Message):
         self.elapsed = elapsed
 
 
+class TunnelClosed(Message):
+    """The SSH tunnel's child exited on its own, and took the forward with it."""
+
+    def __init__(self, notice: str) -> None:
+        super().__init__()
+        self.notice = notice
+
+
 class TransactionModeChanged(Message):
     def __init__(self, new_mode: HarlequinTransactionMode | None) -> None:
         super().__init__()
@@ -360,6 +368,8 @@ class Harlequin(AppBase):
                 title="SSH tunnel",
                 severity="warning" if self.ssh_tunnel.reused else "information",
             )
+            # posted from the watcher's thread, which is why it is a message
+            self.ssh_tunnel.watch(self._report_tunnel_closed)
 
         self._connect()
         self._load_catalog_cache()
@@ -807,6 +817,16 @@ class Harlequin(AppBase):
                 self.editor.focus()
         self.data_catalog.disabled = sidebar_hidden
 
+    def _report_tunnel_closed(self, notice: str) -> None:
+        """Called on the tunnel's watcher thread, so it only posts a message."""
+        self.post_message(TunnelClosed(notice))
+
+    @on(TunnelClosed)
+    def report_tunnel_closed(self, message: TunnelClosed) -> None:
+        """Say the forward is gone, rather than let queries fail unexplained."""
+        message.stop()
+        self.notify(message.notice, title="SSH tunnel", severity="error")
+
     @on(TransactionModeChanged)
     def update_transaction_button_label(self, message: TransactionModeChanged) -> None:
         message.stop()
@@ -1018,6 +1038,9 @@ class Harlequin(AppBase):
             config_path=config_path,
             keymap_names=self.keymap_names,
             theme=self.theme,
+            ssh_tunnel=(
+                self.ssh_tunnel.describe() if self.ssh_tunnel is not None else None
+            ),
         )
         adapter_info = AdapterDebugInfo(
             adapter_options=adapter_options,
--- a/src/harlequin/components/debug_info.py
+++ b/src/harlequin/components/debug_info.py
@@ -50,6 +50,7 @@ class HarlequinDebugInfo:
         active_profile_name: str | None = None,
         active_profile_config: Profile | None = None,
         adapter_options: Sequence[AbstractOption] | None = None,
+        ssh_tunnel: str | None = None,
     ) -> None:
         self.all_keymaps = all_keymaps
         self.config = config
@@ -61,6 +62,8 @@ class HarlequinDebugInfo:
         self.adapter_options = adapter_options
         """What the connected adapter declares, so that this screen knows
         which of the profile's values it must not print."""
+        self.ssh_tunnel = ssh_tunnel
+        """The forwards this session is reached through, where it is tunneled."""
 
     def parse_info(self) -> List[DebugWidget]:
         redacted_config = {
@@ -96,6 +99,17 @@ class HarlequinDebugInfo:
                 widget_type=WidgetType.MARKDOWN,
                 title="Theme",
                 content=f"`{self.theme}`",
+            ),
+            *(
+                [
+                    DebugWidget(
+                        widget_type=WidgetType.MARKDOWN,
+                        title="SSH Tunnel",
+                        content=f"`{self.ssh_tunnel}`",
+                    )
+                ]
+                if self.ssh_tunnel
+                else []
             ),
             DebugWidget(
                 widget_type=WidgetType.MARKDOWN,
--- a/src/harlequin/hsql/modes/info.py
+++ b/src/harlequin/hsql/modes/info.py
@@ -30,6 +30,8 @@ from __future__ import annotations
 
 import json
 import platform
+import shutil
+import subprocess
 import sys
 from importlib.metadata import version
 from typing import TYPE_CHECKING, Any, BinaryIO
@@ -93,6 +95,7 @@ def report(
             "release": platform.release(),
             "machine": platform.machine(),
         },
+        "ssh": _ssh(),
         "config": _config_files(config_path),
         "profile": profile,
         "adapter": adapter_in_use,
@@ -100,6 +103,28 @@ def report(
     }
     out.write((json.dumps(document, indent=2, default=str) + "\n").encode("utf-8"))
     return ExitCode.OK
+
+
+def _ssh() -> dict[str, Any]:
+    """The client `--ssh-host` would run, if there is one on this machine.
+
+    The tunnel options are the one part of hsql that depends on a program
+    outside the wheel, so "is there an ssh here, and which" is a fact about the
+    installation exactly like the adapters below. It runs `ssh -V`, which
+    connects to nothing.
+    """
+    path = shutil.which("ssh")
+    if path is None:
+        return {"client": None, "version": None}
+    try:
+        completed = subprocess.run([path, "-V"], capture_output=True, timeout=10)
+    except (OSError, subprocess.SubprocessError):
+        return {"client": path, "version": None}
+    # OpenSSH writes its version to stderr
+    reported = (
+        (completed.stderr or completed.stdout).decode("utf-8", errors="replace").strip()
+    )
+    return {"client": path, "version": reported or None}
 
 
 def _config_files(config_path: Path | None) -> dict[str, Any]:
--- a/src/harlequin/ssh.py
+++ b/src/harlequin/ssh.py
@@ -26,7 +26,7 @@ import threading
 import time
 from collections import deque
 from dataclasses import dataclass
-from typing import Any, Mapping, Sequence
+from typing import Any, Callable, Mapping, Sequence
 
 from harlequin.config import CONFIG_ERROR_TITLE, DEFAULT_SSH_TIMEOUT, parse_seconds
 from harlequin.exception import HarlequinConfigError, HarlequinSshError
@@ -38,6 +38,16 @@ DEFAULT_TIMEOUT = DEFAULT_SSH_TIMEOUT
 """Seconds to wait for the forwards, when nothing says otherwise."""
 
 ERROR_TITLE = "Harlequin could not open the SSH tunnel."
+
+KEEPALIVE_SECONDS = 30
+"""What Harlequin sets `ServerAliveInterval` to when the user's config sets none.
+
+`ssh` sends no keepalives by default, so an idle forward behind a NAT or a
+corporate firewall is reaped silently -- and a tunnel that drops after lunch is
+the difference between a feature people trust and one they work around. Not
+configurable: an interval is not a thing anyone should have to tune, and the one
+person who wants to has a `Host` block, which is left alone.
+"""
 
 _POLL_INTERVAL = 0.05
 _CONNECT_TIMEOUT = 0.5
@@ -98,6 +108,9 @@ class SshConfig:
     """
 
     forwards: tuple[Forward, ...]
+
+    server_alive_interval: int | None = None
+    """What the resolved config sets, or None where `-G` did not say."""
 
 
 @dataclass(frozen=True)
@@ -172,13 +185,16 @@ def build_argv(
     forwards: Sequence[str] = (),
     *,
     batch_mode: bool = False,
+    keepalive: int | None = None,
 ) -> list[str]:
     """The `ssh` command that holds `forwards` open, and runs nothing.
 
-    `ExitOnForwardFailure` is the only option Harlequin imposes on its own: a
-    forward that silently did not happen is the one failure a user cannot
-    diagnose. Notably not imposed is `ServerAliveInterval`, which a command-line
-    `-o` would take away from a `Host` block that already sets it.
+    `ExitOnForwardFailure` is the only option Harlequin imposes unconditionally:
+    a forward that silently did not happen is the one failure a user cannot
+    diagnose. `keepalive` is the other, and is only ever passed where the
+    resolved config asked for none -- a command-line `-o` beats the config file,
+    and overriding a `Host` block's own interval would be Harlequin retuning
+    someone else's connection.
 
     Raises: HarlequinSshError if a value would reach `ssh` as a flag.
     """
@@ -188,6 +204,8 @@ def build_argv(
     argv = [SSH, "-N", "-o", "ExitOnForwardFailure=yes"]
     if batch_mode:
         argv += ["-o", "BatchMode=yes"]
+    if keepalive is not None:
+        argv += ["-o", f"ServerAliveInterval={keepalive}"]
     for forward in forwards:
         argv += ["-L", forward]
     argv.append(host)
@@ -210,6 +228,13 @@ def build_tunnel(
     """
     argv = build_argv(host, forwards, batch_mode=batch_mode)
     config = resolve_config(argv, timeout=timeout)
+    if config is not None and config.server_alive_interval == 0:
+        # the probe answered, and what it said is that nothing is keeping this
+        # connection alive. Built again rather than probed again: the keepalive
+        # changes nothing else `-G` would report.
+        argv = build_argv(
+            host, forwards, batch_mode=batch_mode, keepalive=KEEPALIVE_SECONDS
+        )
     if config is not None and not config.forwards:
         # a usage error rather than a failure to connect: nothing has been run
         # yet, and what is wrong is what was asked for
@@ -254,12 +279,14 @@ def resolve_config(argv: Sequence[str], *, timeout: float) -> SshConfig | None:
 def parse_config(text: str) -> SshConfig | None:
     """`ssh -G`'s output, or None if this is not `ssh -G`'s output.
 
-    Each line is a lowercase keyword and its resolved value, and the only one
-    read is `localforward`. A `hostname` line is what tells the two apart: every
+    Each line is a lowercase keyword and its resolved value; the two read are
+    `localforward` and `serveraliveinterval`. A `hostname` line is what tells
+    `-G`'s output from anything else that reached this: every
     `-G` prints one, so text without it is something else, and the caller
     degrades rather than concluding there is nothing to forward.
     """
     forwards: list[Forward] = []
+    server_alive_interval: int | None = None
     is_ssh_output = False
     for line in text.splitlines():
         match = _KEYWORD.match(line.strip())
@@ -272,9 +299,14 @@ def parse_config(text: str) -> SshConfig | None:
             fields = value.split()
             if len(fields) == 2:
                 forwards.append(Forward(listen=fields[0], destination=fields[1]))
+        elif keyword == "serveraliveinterval":
+            with contextlib.suppress(ValueError):
+                server_alive_interval = int(value)
     if not is_ssh_output:
         return None
-    return SshConfig(forwards=tuple(forwards))
+    return SshConfig(
+        forwards=tuple(forwards), server_alive_interval=server_alive_interval
+    )
 
 
 class SshTunnel:
@@ -304,6 +336,8 @@ class SshTunnel:
         """Whether a listener that was already there is what answers now."""
         self._process: subprocess.Popen[bytes] | None = None
         self._stderr: _Stderr | None = None
+        self._on_close: Callable[[str], None] | None = None
+        self._stopping = False
 
     @property
     def endpoints(self) -> tuple[tuple[str, int], ...]:
@@ -349,6 +383,17 @@ class SshTunnel:
         """
         return (self.host, *(f"{f.listen} {f.destination}" for f in self.forwards))
 
+    def watch(self, on_close: Callable[[str], None]) -> None:
+        """Call `on_close` from a thread when the child exits on its own.
+
+        A dropped forward is otherwise an unexplained wall of query errors, and
+        the front end is the only thing that can say so. Not called for a tunnel
+        `stop()` took down, which is not news. Set once; every `start()` after
+        this re-arms it.
+        """
+        self._on_close = on_close
+        self._arm_watcher()
+
     def start(self) -> None:
         """Start it and block until the forwarded ports accept connections.
 
@@ -357,6 +402,7 @@ class SshTunnel:
         if self.running:
             return
         self.reused = False
+        self._stopping = False
         already_bound = any(_accepts(endpoint) for endpoint in self.endpoints)
         try:
             process = subprocess.Popen(
@@ -380,11 +426,13 @@ class SshTunnel:
         except BaseException:
             self.stop()
             raise
+        self._arm_watcher()
 
     def stop(self) -> None:
         """Kill the child, if there is one. Safe to call more than once."""
         atexit.unregister(self.stop)
         _LIVE.discard(self)
+        self._stopping = True
         process, self._process = self._process, None
         if process is None:
             return
@@ -405,6 +453,23 @@ class SshTunnel:
 
     def __exit__(self, *exc: Any) -> None:
         self.stop()
+
+    def _arm_watcher(self) -> None:
+        """Start the thread that waits on the child, if anything is listening."""
+        if self._on_close is None or not self.running:
+            return
+        threading.Thread(target=self._watch, daemon=True).start()
+
+    def _watch(self) -> None:
+        process, on_close = self._process, self._on_close
+        if process is None or on_close is None:
+            return
+        process.wait()
+        if self._stopping or self._process is not process:
+            # taken down deliberately, or replaced by a restart
+            return
+        last_line = self._stderr.last_line() if self._stderr is not None else ""
+        on_close(f"tunnel closed: {last_line}" if last_line else "tunnel closed")
 
     def _wait_until_ready(self, *, already_bound: bool) -> None:
         """Block until every forwarded port answers, or say why it never did."""
@@ -469,6 +534,11 @@ class _Stderr:
 
     def text(self) -> str:
         return b"".join(self._chunks).decode("utf-8", errors="replace").strip()
+
+    def last_line(self) -> str:
+        """ssh's own last word, which is what a reader wants out of a paragraph."""
+        lines = [line for line in self.text().splitlines() if line.strip()]
+        return lines[-1].strip() if lines else ""
 
     def close(self) -> None:
         self._thread.join(timeout=_TERMINATE_SECONDS)
--- a/tests/unit_tests/test_ssh.py
+++ b/tests/unit_tests/test_ssh.py
@@ -15,7 +15,7 @@ import subprocess
 import sys
 import time
 from pathlib import Path
-from typing import Iterator, Sequence
+from typing import Callable, Iterator, Sequence
 from unittest.mock import MagicMock
 
 import click
@@ -610,46 +610,46 @@ def test_the_keepalive_is_parsed_off_the_resolved_config() -> None:
     assert bare is not None and bare.server_alive_interval is None
 
 
-def wait_for(notices: list[str], *, seconds: float = 5.0) -> None:
+def wait_until(predicate: Callable[[], bool], *, seconds: float = 10) -> bool:
+    """Poll until something a thread does becomes true, or give up."""
     deadline = time.monotonic() + seconds
-    while not notices and time.monotonic() < deadline:
+    while time.monotonic() < deadline:
+        if predicate():
+            return True
         time.sleep(0.02)
+    return predicate()
 
 
-def test_a_child_that_dies_says_so_in_ssh_s_last_words(
-    monkeypatch: pytest.MonkeyPatch,
-) -> None:
-    monkeypatch.setenv("FAKE_SSH_STDERR", "Timeout, server web-1 not responding.")
-    monkeypatch.setenv("FAKE_SSH_DIE_AFTER", "0.2")
+def test_a_child_that_dies_says_so_in_ssh_s_last_words(drop_trigger: Path) -> None:
     notices: list[str] = []
     tunnel = child_tunnel(free_port())
     tunnel.watch(notices.append)
     tunnel.start()
     try:
-        wait_for(notices)
+        drop_trigger.touch()
+        assert wait_until(lambda: bool(notices))
     finally:
         tunnel.stop()
     assert notices == ["tunnel closed: Timeout, server web-1 not responding."]
 
 
 def test_a_tunnel_watched_after_it_started_is_still_reported(
-    monkeypatch: pytest.MonkeyPatch,
+    drop_trigger: Path,
 ) -> None:
     """The IDE's own order: `cli.py` starts the tunnel, `on_mount` watches it.
 
     Everything in between -- building the app, mounting the widget tree -- is
     time the child can die in.
     """
-    monkeypatch.setenv("FAKE_SSH_STDERR", "Timeout, server web-1 not responding.")
     notices: list[str] = []
     tunnel = child_tunnel(free_port())
     tunnel.start()
     try:
+        drop_trigger.touch()
         assert tunnel._process is not None
-        tunnel._process.kill()
         tunnel._process.wait()
         tunnel.watch(notices.append)
-        wait_for(notices)
+        assert wait_until(lambda: bool(notices))
     finally:
         tunnel.stop()
     assert notices == ["tunnel closed: Timeout, server web-1 not responding."]
@@ -672,6 +672,21 @@ def test_a_reused_listener_is_never_called_closed(
     assert notices == []
 
 
+def test_watching_twice_does_not_double_the_notice(drop_trigger: Path) -> None:
+    notices: list[str] = []
+    tunnel = child_tunnel(free_port())
+    tunnel.watch(notices.append)
+    tunnel.watch(notices.append)
+    tunnel.start()
+    try:
+        drop_trigger.touch()
+        assert wait_until(lambda: bool(notices))
+        time.sleep(0.3)
+    finally:
+        tunnel.stop()
+    assert len(notices) == 1
+
+
 def test_a_tunnel_taken_down_on_purpose_is_not_news() -> None:
     """The control is the test above: the same harness does fire, on a death."""
     notices: list[str] = []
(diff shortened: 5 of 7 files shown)


src/harlequin/hsql/modes/info.py
```diff
@@ -102,6 +105,28 @@ def report(
...
 
+def _ssh() -> dict[str, Any]:
+    """The client `--ssh-host` would run, if there is one on this machine.
+
+    The tunnel options are the one part of hsql that depends on a program
+    outside the wheel, so "is there an ssh here, and which" is a fact about the
+    installation exactly like the adapters below. It runs `ssh -V`, which
+    connects to nothing.
+    """
+    path = shutil.which("ssh")
```
tconbeer: **`shutil.which` reintroduces the working-directory lookup `_client_path()` exists to close.**

CPython's `shutil.which` on `win32` does this:

```python
if sys.platform == "win32":
    # The current directory takes precedence on Windows.
    curdir = os.curdir
    ...
    path.insert(0, curdir)
```

(verified in this container's interpreter). It also expands `PATHEXT`, so `.\ssh.bat` and `.\ssh.cmd` count. Line 120 then executes whatever it found.

That is exactly the hazard `_client_path()` (`ssh.py:730`) was written for — *"a repository that ships an `ssh.exe` would otherwise run in place of the user's client"* — and which `test_the_client_is_resolved_against_path_not_the_working_directory` pins with a decoy that literally writes `pwned`. `resolve_config()` and `start()` both route through it; this doesn't.

The exposure is new and lands in the worst mode for it: `--info` is what a human or an agent runs *first*, its docstring promises it connects to nothing, and the test at line 689 previously asserted `--info` ran `ssh` **not at all**. On Windows, `cd` into a cloned repo and run `hsql --info` → repo-supplied code executes as the user.

It also makes this function's own docstring false: it claims to report *"the client `--ssh-host` would run"*, but `start()` would run a different one.

Fix: resolve through `harlequin.ssh`. Nothing blocks the import — `harlequin.ssh` is not in the `hsql does not reach the TUI` forbidden list, `test_import_hygiene.py:36` already proves it drags in no TUI, and `hsql/cli.py` and `hsql/timeout.py` already deferred-import it. Suggest splitting the "find it" half out:

```python
def find_client(program: str = SSH) -> str | None:
    """`program` as a path, resolved against PATH and nothing else, or None."""
    ...  # current body, returning None instead of `program`

def _client_path(program: str) -> str:
    return find_client(program) or program
```

`real_ssh()` in `test_ssh.py:495` open-codes the same pair a third time, so a shared helper has three callers.

---

src/harlequin/ssh.py
```diff
@@ -457,6 +504,28 @@ def __enter__(self) -> SshTunnel:
     def __exit__(self, *exc: Any) -> None:
         self.stop()
 
+    def _arm_watcher(self) -> None:
+        """Start the thread that waits on the child, if anything is listening."""
+        if self._on_close is None or not self.running:
```
tconbeer: **The `running` gate drops the notification in exactly the case the feature exists for — and the IDE's real call order walks through it.**

`running` is `_process is not None and _process.poll() is None`. If the child is already dead when arming happens, no thread starts and `on_close` is never called.

The IDE hits this window every start-up: `cli.py:668` calls `_open_tunnel()` → `start()` **before** the app object exists, so `_on_close` is still `None` and `start()`'s own `_arm_watcher()` (line 479) is a no-op. The only arming is `app.py:372`, inside `on_mount`. Everything between — constructing `Harlequin`, Textual start-up, mounting the widget tree — is unwatched, and a child that drops in there (server-side idle disconnect, a bastion closing right after auth, a laptop suspending during start-up) produces no notice at all. The user gets precisely the "unexplained wall of query errors" this docstring says it prevents.

Reproduced on this branch, against the fake client, in the IDE's order (start → child dies → watch):

```
running at watch time: False
notices after death-then-watch: []
```

There is a second, narrower instance of the same gate: the child can exit between `_wait_until_ready()` returning and `_arm_watcher()` at line 479 — `ExitOnForwardFailure` guarantees ssh exits on a forward it cannot take.

Suggested fix — gate on the process existing rather than on it still living, and let `_watch()` fall through, since `process.wait()` returns immediately for a dead child and the notice is still correct:

```python
if self._on_close is None or self._process is None or self.reused:
    return
```

The `self.reused` clause is load-bearing: under `--ssh-allow-reuse` the child has already exited by design and `running` is what currently (accidentally) suppresses a bogus "tunnel closed" for a listener Harlequin never opened. Make that explicit here or the fix regresses it. Worth a test for the reuse case at the same time — it's the one that will break.

---

src/harlequin/app.py
```diff
@@ -813,6 +823,16 @@ def watch_sidebar_hidden(self, sidebar_hidden: bool) -> None:
...
 
+    def _report_tunnel_closed(self, notice: str) -> None:
+        """Called on the tunnel's watcher thread, so it only posts a message."""
+        self.post_message(TunnelClosed(notice))
+
+    @on(TunnelClosed)
+    def report_tunnel_closed(self, message: TunnelClosed) -> None:
+        """Say the forward is gone, rather than let queries fail unexplained."""
+        message.stop()
+        self.notify(message.notice, title="SSH tunnel", severity="error")
```
tconbeer: **ssh's last line reaches `notify()` as Textual markup, and it's the one ssh-stderr path with no `redact_text()`.**

`App.notify()` defaults to `markup=True` (verified against the pinned textual), and `Toast.render()` does `Content.from_markup(notification.message)`. The message here is built in `ssh.py:527` straight from `_Stderr.last_line()` → `_Stderr.text()`, i.e. the raw decoded pipe.

Two things follow. Verified locally:

```python
Content.from_markup('tunnel closed: [@click=app.quit()]Click to reconnect[/]')
# plain: 'tunnel closed: Click to reconnect'
# spans: [Span(15, 33, style='@click=app.quit()')]

Content.from_markup('tunnel closed: [red on white]FAKE: enter your password[/]')
# plain: 'tunnel closed: FAKE: enter your password'  spans: [Span(15, 40, style='red on white')]

Content.from_markup('Received disconnect from 10.0.0.5 port 22:2: [/b] bye')
# MarkupError: closing tag '[/b]' does not match any open tag
```

Routine ssh output is fine — I checked `bind [127.0.0.1]:15439: Address already in use` and `ssh: connect to host [::1] port 22: ...`, which parse cleanly, since the parser tolerates unmatched *opening* brackets. The reachable input is the text ssh does **not** author: OpenSSH prints the server's own disconnect description (`Received disconnect from …: <server text>`) and `ProxyCommand`/`Match exec` helper output verbatim. So a hostile or compromised bastion picks the string, and it becomes styled text or a live action span rather than the literal line, with a closing tag raising in the toast's render path.

`markup=False` here fixes it, and is worth applying to `app.py:366` (`notice()`) for the same reason.

Separately, on the redaction side: every sibling path in `ssh.py` handles these bytes through `redact_text()` — `_Stderr._show()` does, with a class docstring saying *"so `redact_text()` sees whole lines: ssh is third-party output, and a driver's DSN can reach it"* — and `_child_failed()`/`_timed_out()` quote only `_unseen()`, which returns `""` whenever `echo_stderr` is true, i.e. always in both shipped front ends. This notice is the only one that carries raw stderr to a surface with no redactor in the path.

It leaks nothing *today* (`hide_secrets_in()` has one caller, in `hsql/cli.py`, so `_HIDDEN` is empty under the IDE). But that's the latency, not a defence: the day the IDE registers its secrets — which the debug screen's profile redaction suggests it should — `_show()` starts masking and this line silently won't. `on_close(redact_text(...))` in `_watch()` is one line and puts it on the same footing as its siblings.

---

src/harlequin/ssh.py
```diff
@@ -43,6 +43,16 @@
...
 
+KEEPALIVE_SECONDS = 30
+"""What Harlequin sets `ServerAliveInterval` to when the user's config sets none.
+
+`ssh` sends no keepalives by default, so an idle forward behind a NAT or a
+corporate firewall is reaped silently -- and a tunnel that drops after lunch is
+the difference between a feature people trust and one they work around. Not
+configurable: an interval is not a thing anyone should have to tune, and the one
+person who wants to has a `Host` block, which is left alone.
+"""
```
tconbeer, line 65: **"which is left alone" is not true, and `-G` cannot make it true.**

`ssh -G` prints every *resolved* option, and `ServerAliveInterval`'s built-in default is `0`. `fill_default_options()` normalizes the unset sentinel to `0` before `dump_client_config()` prints it, so `serveraliveinterval 0` is emitted identically for "nothing set it" and for an explicit `ServerAliveInterval 0` in a `Host` block. `parse_config()` cannot answer the question `build_tunnel` asks it — the information is not in the output.

So the person this sentence promises to leave alone — the one who deliberately turned keepalives off in a `Host` block — is the one person who gets overridden, and since a command-line `-o` beats the config file they have no way to get what they configured short of writing an absurd interval.

That makes three claims in the diff wrong as written:

- here, "the one person who wants to has a `Host` block, which is left alone";
- line 127, `"""What the resolved config sets, or None where `-G` did not say."""` — `-G` always says; `None` only ever means this wasn't `-G` output at all;
- line 239, "`keepalive` is passed only where the resolved config asked for none" — it is also passed where the config asked for exactly zero.

The behaviour is defensible (someone with `ServerAliveInterval 0` is rare, and a reaped forward is worse); the documented contract is what's off. Either state plainly that a resolved `0` always becomes 30, or add the escape hatch §8 defers.

While here — this docstring also runs afoul of the AGENTS.md comment rules on two counts. *"a tunnel that drops after lunch is the difference between a feature people trust and one they work around"* is a product argument for the feature existing, and *"Not configurable: an interval is not a thing anyone should have to tune, and the one person who wants to has a `Host` block"* is a decision recorded with its alternative rebutted — both belong in the PR description or `docs/plans/`, which already carry them (§5.4, §8). Suggest:

```python
KEEPALIVE_SECONDS = 30
"""What Harlequin sets `ServerAliveInterval` to where `ssh -G` resolves it to 0.

`ssh` sends no keepalives by default, so an idle forward behind a NAT or a
firewall is reaped silently.
"""
```

`test_the_keepalive_is_parsed_off_the_resolved_config` pins `parse_config("hostname web-1\n").server_alive_interval is None` — output real `ssh -G` never produces — so nothing here catches the divergence.

---

src/harlequin/ssh.py
```diff
@@ -255,6 +272,13 @@ def build_tunnel(
     """
     argv = build_argv(host, forwards, batch_mode=batch_mode)
     config = resolve_config(argv, timeout=timeout)
+    if config is not None and config.server_alive_interval == 0:
+        # the probe answered, and what it said is that nothing is keeping this
+        # connection alive. Built again rather than probed again: the keepalive
+        # changes nothing else `-G` would report.
+        argv = build_argv(
+            host, forwards, batch_mode=batch_mode, keepalive=KEEPALIVE_SECONDS
+        )
```
tconbeer: Two things on this block.

**The comment.** *"Built again rather than probed again"* is the "X rather than the old Y" shape AGENTS.md bans — it presents a road not taken as a live distinction, and the reader has to reconstruct what "probed again" would have meant. The first sentence also narrates the `if` immediately above it. Only the last clause carries information:

```python
        # nothing in the resolved config keeps this connection alive; the added
        # option changes nothing else `-G` would report
```

**The keepalive's other half.** `ServerAliveCountMax` is left at whatever the config resolves to (default 3), so the net effect is a hard ~90s unresponsiveness budget where there was none. Before this, a tunnel with no keepalive survived a network stall or a brief suspend as long as the TCP connection held; now ssh actively tears it down. The PR ships both halves, so for a flaky-link user the visible result of "keepalive" is *more* tunnel deaths, announced. Probably still the right trade — a silently NAT-reaped forward is worse than a loudly dead one — but worth stating, and worth considering pinning `ServerAliveCountMax` alongside the interval so the budget is Harlequin's and predictable rather than inherited.

One concrete regression that falls out of it: `ServerAliveCountMax 0` is a coherent "keepalives off" setting *today*, because the interval is 0 and countmax is inert. OpenSSH's `server_alive_check()` is `if (ssh_packet_inc_alive_timeouts(ssh) > options.server_alive_count_max)` and the increment is pre-, so the first check is `1 > 0` → `Timeout, server %s not responding.` and exit 255. With this PR that user's tunnel dies 30 seconds in against a healthy server. (Read off the OpenSSH source, not executed — no ssh client in the review container — so worth a sanity check before acting on it.)

---

src/harlequin/ssh.py
```diff
@@ -457,6 +504,28 @@ def __enter__(self) -> SshTunnel:
...
+        if self._stopping or self._process is not process:
+            # taken down deliberately, or replaced by a restart
+            return
+        last_line = ""
+        if self._stderr is not None:
+            # what the child wrote on its way out is still in the pipe when
+            # `wait()` returns, and it is the whole of what this notice says
+            self._stderr.settle()
+            last_line = self._stderr.last_line()
+        on_close(f"tunnel closed: {last_line}" if last_line else "tunnel closed")
```
tconbeer, line 636: Three smaller things in `_watch()`.

**`_stderr` is read live, after the identity check that guards it.** `process` is captured at the top but `_stderr` is not, so between `self._process is not process` passing and the `self._stderr` read, a `stop()` + `start()` can replace both — and the watcher then settles and quotes the *new* child's `_Stderr` while attributing it to the old child's death. Narrow (no in-tree caller restarts a tunnel yet; 4/5 will), and the fix is to capture it alongside `process`:

```python
process, stderr, on_close = self._process, self._stderr, self._on_close
```

**Two guards for one question.** `stop()` already does `process, self._process = self._process, None`, so `self._process is not process` is True for a deliberate stop before the watcher ever looks — the same check that covers a restart. `_stopping` (added at lines 388/453/485) buys only the sliver between the flag being set and the attribute being swapped. Worth deciding which one is load-bearing and dropping the other; as written the code doesn't say.

**The comment.** *"what the child wrote on its way out is still in the pipe when `wait()` returns"* is already the whole content of `_Stderr.settle()`'s own docstring twelve lines down, which is its right home, and the second clause narrates line 527. The neighbouring one-liner at 519 (`# taken down deliberately, or replaced by a restart`) is the model — it names the two conditions the `if` tests. Suggest dropping both lines here.

---

src/harlequin/ssh.py
```diff
@@ -398,6 +431,17 @@ def cache_material(self) -> tuple[str, ...]:
...
+    def watch(self, on_close: Callable[[str], None]) -> None:
+        """Call `on_close` from a thread when the child exits on its own.
+
+        A dropped forward is otherwise an unexplained wall of query errors, and
+        the front end is the only thing that can say so. Not called for a tunnel
+        `stop()` took down, which is not news. Set once; every `start()` after
+        this re-arms it.
+        """
+        self._on_close = on_close
+        self._arm_watcher()
```
tconbeer, line 545: **"Set once" is documented but not enforced.** `watch()` unconditionally calls `_arm_watcher()`, and `start()` arms too, so two `watch()` calls on a running tunnel give two threads and two notices — reproduced: `['tunnel closed', 'tunnel closed']`. Latent with one in-tree caller, but this is a public method on a class out-of-tree front ends can reach, and duplicate error toasts are invisible until they aren't. An `if self._on_close is not None` check before arming closes it.

On the docstring: sentences three and four are real contract a caller can't infer from the signature — keep those. Sentence two (*"an unexplained wall of query errors, and the front end is the only thing that can say so"*) is the feature's justification, and it's repeated a third time in `app.py:832`'s handler docstring. *"which is not news"* is editorial. Suggest:

```python
"""Call `on_close` from a thread when the child exits on its own.

Not called for a tunnel `stop()` took down. Set once; every `start()`
after this re-arms it.
"""
```

---

src/harlequin/hsql/modes/info.py
```diff
@@ -102,6 +105,28 @@ def report(
...
+        return {"client": None, "version": None}
+    try:
+        completed = subprocess.run([path, "-V"], capture_output=True, timeout=10)
+    except (OSError, subprocess.SubprocessError):
+        return {"client": path, "version": None}
+    # OpenSSH writes its version to stderr
+    reported = (
+        (completed.stderr or completed.stdout).decode("utf-8", errors="replace").strip()
+    )
+    return {"client": path, "version": reported or None}
```
tconbeer: Four smaller things on this function (the `shutil.which` one is separate, above).

**`returncode` is never checked.** `ssh` on PATH is not necessarily OpenSSH — a wrapper script, a renamed `plink`, a corporate shim. Anything that rejects `-V` and prints usage to stderr lands verbatim in `info["ssh"]["version"]`, and `.strip()` only trims the ends, so embedded newlines survive into a document that's a machine-readable contract (`hsql/skill/references/troubleshooting.md:90` tells users to capture it). Gate on `returncode == 0`, or take the first line, or both.

**No `stdin=subprocess.DEVNULL`.** `resolve_config()` passes it with a comment explaining why. `ssh -V` parses no config so `Match exec` doesn't apply — this is consistency rather than a bug — but it's the one place in the feature that spawns ssh without it, and it matters more given the decoy-binary path above.

**It runs unconditionally**, whether or not any `--ssh-*` option was given, so a hanging client stalls `--info` for the full hardcoded `timeout=10` (`DEFAULT_TIMEOUT` is right there), and the ~25 existing `--info` tests in `test_hsql.py` now spawn a real subprocess on any dev machine with ssh — making the document machine-dependent in a mode whose other tests pin it exactly.

**The docstring.** The middle sentence argues why this belongs in `--info` at all — a decision, for the PR description — and *"exactly like the adapters below"* is a positional reference into `report()`'s dict that goes stale the moment it's reordered. The last clause is the keeper: that a `--info` run shells out but connects to nothing is genuinely non-obvious.

```python
    """The client `--ssh-host` would run, if there is one on this machine.

    `ssh -V` connects to nothing.
    """
```

(`# OpenSSH writes its version to stderr` at line 123 is exactly right — that's the model.)

Naming, per AGENTS.md: `_ssh` → `_ssh_client` (it answers "which client", matching `_config_files`/`_adapter_in_use`), and `reported` → `version_text`.

---

src/harlequin/components/debug_info.py
```diff
@@ -97,6 +100,17 @@ def parse_info(self) -> List[DebugWidget]:
...
+            *(
+                [
+                    DebugWidget(
+                        widget_type=WidgetType.MARKDOWN,
+                        title="SSH Tunnel",
+                        content=f"`{self.ssh_tunnel}`",
+                    )
+                ]
+                if self.ssh_tunnel
+                else []
```
tconbeer: `grep -rn '\*(\[' src/harlequin/` returns only this — it's the one splat-conditional in the package, twelve lines and three nesting levels inside a flat list literal for one widget. The file's own idiom for state-dependent content is to compute it above the literal (see `AdapterDebugInfo.parse_info`'s `options_markdown`), which also gives the value a name:

```python
tunnel_details = (
    [DebugWidget(WidgetType.MARKDOWN, "SSH Tunnel", f"`{self.ssh_tunnel}`")]
    if self.ssh_tunnel
    else []
)
```

Simpler still if a `None` row is acceptable: `Active Profile`, `Config File Location` and `Theme` are all emitted unconditionally and render `` `None` `` when unset, so an unconditional entry needs no branch at all. Your call — it does cost the "existing snapshot unchanged" property the PR description claims.

Passing `describe()`'s **string** rather than the `SshTunnel` (`app.py:1047`) is the right call, though — it keeps `components/` free of an `ssh` import and reuses the formatter instead of forking one.

Two notes on what this prints. `describe()` includes the raw `--ssh-host`, which may be `user@bastion.example.com`. That's consistent with the repo's stated position — `_URI_PASSWORD` is documented as matching *"the password ... and not the user beside it"*, and `redact_conn_str()` deliberately keeps host, port, database and user — so a username and an internal hostname aren't secrets here, and the screen already shows the DSN's host and the profile's `ssh_host`. Not a finding. The gap is only that this is the one value on that screen passed through no redactor at all: if a user typed a credential into `--ssh-host` (`ssh://user:pw@bastion` — ssh ignores it, `_UNSAFE_IN_VALUE` permits `:` and `@`, nothing strips it), it prints verbatim into text meant for bug reports. `redact_conn_str([...])[0]` costs nothing.

The attribute docstring at line 66 is slightly off in two ways: the value is `describe()`'s single line, not "the forwards", and *"where it is tunneled"* restates what `str | None` and the guard here already say. `"""The tunnel this session's connection is reached through."""`

Also worth a line in `tests/functional_tests/test_debug_info_screen.py` — it builds `HarlequinDebugInfo` directly, so covering the new field is a one-line addition and nothing currently exercises it.

---

tests/unit_tests/test_ssh.py
```diff
@@ -520,6 +522,72 @@ def test_the_lifecycle_helpers_agree_about_ports(
...
+def test_a_tunnel_taken_down_on_purpose_is_not_news() -> None:
+    said: list[str] = []
+    tunnel = child_tunnel(free_port())
+    tunnel.watch(said.append)
+    tunnel.start()
+    tunnel.stop()
+    time.sleep(0.3)
+    assert said == []
+
+
```
tconbeer, line 689: The positive test is good — it pins the notice text, not just that something fired. Four gaps around it.

**Both watcher tests call `watch()` before `start()`, which is the opposite of the only production caller.** `cli.py:668` starts the tunnel and `app.py:372` watches it later, in `on_mount`. A test in that order — start, kill, *then* watch — fails today; see the `_arm_watcher` comment above. That ordering is worth being the tested one, since it's the shipped one.

**`test_a_tunnel_taken_down_on_purpose_is_not_news` can hardly fail.** `time.sleep(0.3)` then `assert said == []` passes if `watch()` never armed anything, if `_watch` returned early for the wrong reason, or if a real notification just took longer than 300ms under xdist. It can't fail *spuriously*, but it also isn't pinning much. A `threading.Event` that must not be set, with a generous timeout and a control case showing the same harness does fire, would discriminate.

**Untested:** the `reused` path with `watch()` armed (the case that regresses under the obvious fix for `_arm_watcher`); restart re-arming (I checked by hand that it delivers exactly one notice and leaks no thread, but `_stopping` + the identity check are subtle enough to deserve a test); `--info` with no ssh on PATH, so the `{"client": None, "version": None}` branch never runs.

Nits: `assert not [arg for arg in ... if "ServerAlive" in arg]` (lines 544, 553) builds a list to negate it — `assert not any("ServerAlive" in arg for arg in ...)` is the same assertion as the sentence it is. `said` is the "you have one" name AGENTS.md warns about; `notices`. And `tunnel._process.kill()` is the first reach into `SshTunnel`'s privates in this module — defensible, since there's no public way to make a child die on its own, but a `FAKE_SSH_DIE_AFTER` env var would join the other seven `FAKE_SSH_*` knobs and let the test drive the fake the way every other test here does.

On line 691: weakening `calls(ssh_on_path) == []` to `all("-L" not in argv ...)` still catches a regression that opens a forward, so it's adequate — but `calls(ssh_on_path) == [["-V"]]` restores the original strength for free.

Ran the file on this branch: **79 passed, 1 skipped**. Worth noting the fake's new `FAKE_SSH_ALIVE_INTERVAL` default of `'0'` means every existing `build_tunnel` test now gets `-o ServerAliveInterval=30` injected into its argv — nothing asserts exact argv equality, so it breaks nothing, but the default is load-bearing and the env-var list is where it should say so: `FAKE_SSH_ALIVE_INTERVAL  what `-G` reports for serveraliveinterval (default 0)`.

---

src/harlequin/app.py
```diff
@@ -360,6 +368,8 @@ async def on_mount(self) -> None:
                 title="SSH tunnel",
                 severity="warning" if self.ssh_tunnel.reused else "information",
             )
+            # posted from the watcher's thread, which is why it is a message
+            self.ssh_tunnel.watch(self._report_tunnel_closed)
```
tconbeer: The comment duplicates the "why" that's already stated on the thing it points at (`_report_tunnel_closed`'s docstring, line 827) — and it sits on the `watch(...)` line rather than the line it explains. The docstring is the correctly scoped copy; suggest dropping the comment.

Two related nits nearby:

- `_report_tunnel_closed` / `report_tunnel_closed` differ only by a leading underscore while doing different things on different threads. The file's convention is to name the handler for what it does (`update_transaction_button_label` for `TransactionModeChanged`) — `notify_tunnel_closed` here. The private trampoline can go entirely, matching the `partial(self.post_message, ...)` callback style already used at lines 459/467/498/511/681: `self.ssh_tunnel.watch(lambda notice: self.post_message(TunnelClosed(notice)))`.
- Line 832's docstring (*"rather than let queries fail unexplained"*) is the feature's justification, and the third copy of that argument in the diff. Of ~34 `@on` handlers in this file exactly one carries a docstring; a two-line notify handler needn't be the second.

Threading is right, for the record — I checked `MessagePump.post_message` detects a foreign thread and routes through `loop.call_soon_threadsafe`, returning `False` early if `_closing or _closed`, and `notify()` then runs on the message-pump thread in the handler. The one residual: if the child dies during shutdown, after Textual's loop closes but before `cli.py`'s ExitStack calls `tunnel.stop()` (so `_stopping` is still False), `call_soon_threadsafe` on a closed loop raises `RuntimeError` inside the daemon thread and prints a traceback after the app exits. The `_closed` guard covers the normal ordering; `contextlib.suppress(RuntimeError)` around the `post_message` closes the sliver.

---

src/harlequin/ssh.py
```diff
@@ -569,6 +638,11 @@ def __init__(self, stream: Any, *, echo: bool = True) -> None:
     def text(self) -> str:
         return b"".join(self._chunks).decode("utf-8", errors="replace").strip()
 
+    def last_line(self) -> str:
+        """ssh's own last word, which is what a reader wants out of a paragraph."""
+        lines = [line for line in self.text().splitlines() if line.strip()]
+        return lines[-1].strip() if lines else ""
```
tconbeer: `text()` ends in `.strip()`, so the last element of `splitlines()` can never be blank or whitespace-padded — both the filter and the trailing `.strip()` are dead:

```python
def last_line(self) -> str:
    """The last non-blank line ssh wrote."""
    lines = self.text().splitlines()
    return lines[-1] if lines else ""
```

It does belong on `_Stderr` — same class owns the buffer, sits next to `text()`, doesn't duplicate `flush()`/`settle()`.

The docstring's *"which is what a reader wants out of a paragraph"* defends the choice of last-line-over-whole-text; that's the caller's design question, and a reader of this method needs to know what it returns.

---

tconbeer (commented):
Reviewed as four narrow passes — **correctness/logic**, **reuse & simplification**, **security & secret leakage**, and **comments/docstrings against AGENTS.md** — plus my own read. Reviewed against the stack base (`...-2-options`), not `main`. CI is green, 32/32.

The shape is right and matches §5.4: the probe-then-impose keepalive, the watcher that only ever `post_message`s, and reporting in both front ends. `describe()` reused for the debug screen rather than a second formatter, and the `TunnelClosed` message/handler split, are both the correct patterns. Three things I'd want addressed before merge, then a tail of smaller ones.

**1. `hsql --info` executes a repo-local `ssh.exe` on Windows** (`info.py:116`). `shutil.which` inserts `os.curdir` at the front of the search path on `win32` — verified in the container's interpreter — which is exactly the hazard `_client_path()` was written to close and that `test_the_client_is_resolved_against_path_not_the_working_directory` pins with a decoy that writes `pwned`. New exposure in the mode that promises it connects to nothing, and the one an agent runs first. Nothing blocks importing the existing resolver.

**2. The death notification is silently dropped in the IDE's real call order** (`ssh.py:509`). `cli.py:668` starts the tunnel before the app exists; `app.py:372` arms the watcher in `on_mount`. `_arm_watcher()` gates on `running`, so a child that dies anywhere in that window — Textual start-up, widget mount — is never reported. Reproduced against the fake client: `notices after death-then-watch: []`. The fix has to keep the `reused` path suppressed explicitly, since `running` is currently what does that by accident.

**3. ssh's last line reaches `notify()` as Textual markup** (`app.py:834`). `notify()` defaults `markup=True`. Routine ssh lines parse cleanly (including `[::1]` — the parser tolerates unmatched opening brackets), but the reachable input is the text ssh doesn't author: the server's own disconnect description and `ProxyCommand` helper output. Verified `[@click=app.quit()]…[/]` becomes a live action span and `[/b]` raises `MarkupError`. `markup=False` fixes it. Separately it's the only ssh-stderr path with no `redact_text()` — harmless today only because `_HIDDEN` is empty under the IDE.

**Then:** a resolved `ServerAliveInterval 0` is indistinguishable from unset, so the docstring's "the one person who wants to has a `Host` block, which is left alone" describes the one person who is *overridden* — the behaviour is defensible, the documented contract isn't (two more claims wrong for the same reason, at lines 127 and 239). The keepalive also imports a ~90s unresponsiveness budget via the inherited `ServerAliveCountMax`, worth pinning. `_ssh()` ignores `returncode`. A handful of comments record decisions and rebut alternatives that `docs/plans/` already carries — the `KEEPALIVE_SECONDS` docstring and "Built again rather than probed again" most of all. Test-side, both watcher tests use the opposite ordering from production, and the negative one is a 300ms sleep that can hardly fail.

Nothing here touches the design; #2 is a gate placement, #1 and #3 are one-liners.

## Not checked

- **I did not review 1/5 or 2/5.** This read the diff against `...-2-options` and took the tunnel lifecycle, `_Stderr`, `_client_path`, `_refuse_unsafe_value` and the options plumbing as given. Findings that name them (e.g. `_client_path` as the fix for #1) assume they're already sound.
- **I did not run the app, or any test outside `tests/unit_tests/test_ssh.py`** (79 passed, 1 skipped, on 3.10 in a Linux container). No functional/snapshot run, no `make check`, no mypy, no `lint-imports` — so the claim that `hsql` may import `harlequin.ssh` is read off the contracts in `pyproject.toml` and `test_import_hygiene.py`, not executed.
- **No SSH client and no Windows in the container.** Every claim about `ssh -G`'s output for an unset vs. explicit `ServerAliveInterval 0`, about `ServerAliveCountMax 0` being fatal at the first check, and about `CreateProcess`/`PATHEXT` behaviour is from OpenSSH and CPython source, not observed. The `shutil.which` curdir insert I did verify in-interpreter; the Windows *consequence* I did not.
- **I did not verify the `MarkupError` reaches a user-visible crash.** I confirmed the parse raises and that `Toast.render()` calls `Content.from_markup` when `notification.markup`; a headless `run_test()` didn't surface a toast for me to render, so whether Textual swallows it at render time is open. The action-span reinterpretation is demonstrated either way, and the fix is the same keyword.
- **I did not exercise the reconnect path**, which is 4/5 — several findings here (the `_stderr` re-read race, restart re-arming, duplicate `watch()`) are latent precisely because nothing restarts a tunnel yet, and 4/5 is what makes them live. They should be re-checked there rather than treated as closed by this PR.
- **I did not check the `harlequin-web` docs** or the 5/5 changelog entry beyond confirming `[Unreleased]` carries no SSH entry anywhere in the stack — consistent with the plan. Two things there worth separate clauses: the imposed keepalive, and `--info` gaining an `ssh` key (an output change for anyone parsing that JSON).
- **`redact` coverage was spot-checked, not swept.** I traced `hide_secrets_in`'s single caller and `redact_text`'s behaviour under the IDE; I did not enumerate every path that could register a secret.

---

tconbeer:
All twelve addressed. The `_arm_watcher` finding is the one that mattered most — the notice was unreachable in the only order that ships.

## Correctness

- **`shutil.which` in `--info`.** Fixed by splitting `find_client()` out of `harlequin.ssh` exactly as you suggested; `_client_path()` is now `find_client(program) or program`, and `info.py` calls the shared one. `--info` reports the client `start()` would actually run, on a machine where the working directory shadows it or not, and its docstring is true again. `real_ssh()` in the tests is the third caller.
- **The `running` gate.** Fixed: `_arm_watcher()` returns early on `self._process is None` rather than on `not self.running`, so a child that died between `cli.py`'s `start()` and `on_mount`'s `watch()` is still reported — `process.wait()` returns immediately for a dead child and the notice is still correct. `self.reused` is explicit in the guard, as you said it had to be, since it was `running` that was accidentally suppressing a bogus "tunnel closed" for a listener Harlequin never opened. There is a test for the IDE's real order now, and one for the reuse case.
- **`watch()` set once.** Enforced: a second call returns rather than arming a second thread. Test included; it fails without the guard.
- **Textual markup on ssh's last line.** Fixed with `markup=False`, on both the drop notice and the mount notice. The redaction half fell out of a change in 2/6: `_Output.text()` now runs the bytes through `redact_text()` and a new `_readable()` that strips every control character, so `last_line()` is on the same footing as its siblings whether or not `_HIDDEN` is populated — and the IDE does register its secrets now.
- **`_watch()` reading `_stderr` live.** Fixed — captured beside `process`.
- **Two guards for one question.** `_stopping` is gone. The identity check is load-bearing and covers both conditions, and the comment now says so: `stop()` clears `_process` and `start()` replaces it, so either way the child this thread waited on is not the tunnel's any more.
- **`last_line()`'s dead filter.** Fixed, to your version.
- **`_post_tunnel_closed`** wraps the `post_message` in `contextlib.suppress(RuntimeError)` for the shutdown sliver, the handler is `notify_tunnel_closed`, and the trampoline stays only because it is what carries that suppression.

## The keepalive

You are right that `-G` always prints `serveraliveinterval`, and that all three claims were wrong as written. Two changes:

- **`KEEPALIVE_SECONDS` says what happens**: a resolved `0` always becomes 30, *including* where a `Host` block set it to zero deliberately, because the output cannot tell the two apart. No promise to leave anyone alone.
- **`ServerAliveCountMax` is pinned alongside the interval** (`KEEPALIVE_COUNT = 3`). Your reading of `server_alive_check()` matches mine — the increment is pre-, so `ServerAliveCountMax 0` makes the first check `1 > 0` and the connection dies at 30s against a healthy server. That user's countmax is inert today precisely because their interval is 0, so imposing ours alongside ours changes nothing they had working, and it makes the budget a known `30 × 3` rather than one inherited from a config that was never consulted about it.

`build_argv`'s docstring and the comment at the call site are rewritten; "Built again rather than probed again" is gone.

## `--info`

`returncode` is checked, only the first line is taken, and `stdin=subprocess.DEVNULL` is passed. `_ssh` → `_ssh_client`, `reported` → `version_text`, and the docstring is down to your two lines.

I kept it **unconditional**, with a `_VERSION_SECONDS = 5.0` of its own rather than the hardcoded 10. `--info` is the document the troubleshooting reference tells people to capture, and "is there an ssh here, and which" is a fact about the installation whether or not this particular invocation names a tunnel — making it conditional means `--info` answers differently depending on flags that change nothing about the machine. The bounded timeout is the part that was actually wrong.

## Debug screen

Computed above the literal as `tunnel_details`, matching `AdapterDebugInfo`'s idiom. It also goes through `redact_conn_str()` now — you are right that a username and an internal hostname are not secrets here, but `--ssh-host` takes an `ssh://user:pw@bastion` that nothing else strips, and this screen's text exists to be pasted into an issue. The attribute docstring is your version, and there is a test in `test_debug_info_screen.py` for the field.

## Tests

- The two watcher tests are joined by one in the **shipped order** (start → die → watch), which is the one that failed before the `_arm_watcher` fix.
- `test_a_tunnel_taken_down_on_purpose_is_not_news` now names its control in the docstring — the test above it proves the same harness fires on a real death — and waits longer.
- Added: the reuse path with a watcher armed, and a second `watch()`.
- `--info` with no client on PATH.
- `FAKE_SSH_DIE_AFTER` went in and then came back out during the 4/6 merge: 4/6 already has `FAKE_SSH_DROP_WHEN`, which is the better mechanism for the same job, so the tests here drive that instead of reaching into `_process.kill()`. The env-var list documents `FAKE_SSH_ALIVE_INTERVAL`'s load-bearing default of `0`.
- `assert not any(...)`, and `said` → `notices`.
- `calls(ssh_on_path) == [["-V"]]`, restored to full strength.

---

1476 tests pass on the stack tip, with mypy, ruff and the import-linter contracts. Merged forward through 4/6–6/6, and the stack now sits on `main` at 8187f0f.

---

PR #1115, opened 2026-09-01
SSH tunnels 4/5: reopen a dropped tunnel, and the connection through it

tconbeer:
Part 4 of 5 of `docs/plans/ssh-tunnels-design.md`, and the one step that touches `app.py`. **Based on #1114**; review the stack in order.

Stack: 1/5 tunnel → 2/5 options → 3/5 keepalive & reporting → **4/5 (this)** → 5/5 changelog.

Part of #545

§5.4's third layer: the IDE recovers from a tunnel that went down, lazily.

1. The watcher from 3/5 marks the tunnel `dropped` when the child exits.
2. The next thing that needs the database — `_execute_query`, `update_schema_data` — calls `_recover_tunnel()` on its own worker thread, ahead of reading the connection.
3. That restarts `ssh`, re-runs the readiness poll, and re-runs `adapter.connect()`, replacing `self.connection`.
4. Then the worker does what the user asked.

`SshTunnel` gains `dropped`, `restart_failed`, `needs_restart` and `restart()`. The app gains two messages: `TunnelReconnected` (a warning notification, and a `TransactionModeChanged` for the button) and `TunnelUnrecoverable` (an error modal quoting ssh).

- **Both halves, or neither.** Restarting `ssh` alone is not enough: the adapter's TCP connection ran *through* the old forward and died with it. Reconnecting alone cannot work either, because there is nothing to connect to.
- **Lazily, not on a timer.** Recovery happens before the next thing that needs the database, so a session left open overnight does not hammer a bastion, and nothing reconnects for a user who is not asking for anything.
- **A restart is always `BatchMode=yes`**, whatever the flag said at start-up — the option is inserted ahead of the destination, which `ssh` takes last. Textual owns the terminal by now, so a passphrase prompt has nowhere to go; failing with the credential ssh wanted is the only outcome a user can act on.
- **One attempt per action, and none once one has failed.** `needs_restart` goes false after a failed restart: the user restarts Harlequin, having been told why. A background retry storm against a bastion that is down is how an account gets locked.
- **The notification says what was lost.** A reconnect is a new session, so an open transaction, a temp table and anything set with `SET` are gone with the old one, and a user who believes they are still inside a `BEGIN` is the failure mode worth spending a sentence on.
- **The worker assigns `self.connection` itself**, rather than posting and waiting. It is a plain attribute rather than a widget, and the worker that called `_recover_tunnel()` is about to run a query on it and cannot wait for a message to land. The message it posts is for the notification and the transaction button.
- **`hsql` does none of this**, deliberately: it is short-lived, its statements are not idempotent, and an unattended caller that loses its connection should fail loudly.

Reconnecting a connection that dropped with the tunnel *up* — a server restart, an idle timeout — is deferred in §8; v1 triggers only from the case it can detect without guessing.

- [x] Yes.

Unit tests for `restart()` — it forces `BatchMode=yes`, brings the port back, and marks itself not-to-be-retried when it fails. And `tests/functional_tests/test_ssh_recovery.py` drives the real app: a child that drops mid-session is reopened before the next catalog refresh and the connection is replaced, and one that will not come back raises the modal once and is not tried again.

---

### DIFF
--- a/src/harlequin/app.py
+++ b/src/harlequin/app.py
@@ -204,6 +204,18 @@ class CompletersReady(Message):
         self.member_completer = member_completer
 
 
+_PARTIAL_FAILURE_WORKER_NOTIFICATIONS: dict[str, str] = {
+    "_load_catalog_cache": (
+        "Harlequin could not load its cache; your query history may be missing."
+    ),
+    "_extend_and_merge_completers": "Harlequin could not update completions.",
+    "_build_completers": "Harlequin could not build completions.",
+}
+"""Toast text for the workers whose failure is partial: the app stays usable,
+so their errors surface as a notification rather than an error modal.
+"""
+
+
 class Harlequin(AppBase):
     """
     The SQL IDE for your Terminal.
@@ -627,30 +639,60 @@ class Harlequin(AppBase):
             await self._handle_worker_error(message)
 
     async def _handle_worker_error(self, message: Worker.StateChanged) -> None:
-        if (
-            message.worker.name == "update_schema_data"
-            and message.worker.error is not None
-        ):
+        worker_name = message.worker.name
+        worker_error = message.worker.error
+        if self._exit or worker_error is None:
+            # an error that lands while the app is exiting is noise, not news.
+            return
+        if worker_name == "update_schema_data":
             self._push_error_modal(
                 title="Catalog Error",
                 header="Could not update data catalog",
-                error=message.worker.error,
+                error=worker_error,
             )
             self.data_catalog.database_tree.loading = False
-        elif message.worker.name == "_connect" and message.worker.error is not None:
+        elif worker_name == "_connect":
             title = getattr(
-                message.worker.error,
+                worker_error,
                 "title",
                 "Harlequin could not connect to your database.",
             )
             error = (
-                message.worker.error
-                if isinstance(message.worker.error, HarlequinError)
-                else HarlequinConnectionError(
-                    msg=str(message.worker.error), title=title
-                )
+                worker_error
+                if isinstance(worker_error, HarlequinError)
+                else HarlequinConnectionError(msg=str(worker_error), title=title)
             )
             self.exit(return_code=2, message=pretty_error_message(error))
+        elif worker_name in ("_execute_query", "_fetch_data"):
+            # the worker died before posting QueriesExecuted or ResultsFetched,
+            # which is what would have restored these.
+            self.run_query_bar.set_responsive()
+            self.results_viewer.show_table()
+            header = getattr(worker_error, "title", worker_error.__class__.__name__)
+            self._push_error_modal(
+                title="Query Error",
+                header=header,
+                error=worker_error,
+            )
+        elif worker_name == "toggle_transaction_mode":
+            self._push_error_modal(
+                title="Transaction Error",
+                header="Harlequin could not change the transaction mode.",
+                error=worker_error,
+            )
+        elif worker_name in _PARTIAL_FAILURE_WORKER_NOTIFICATIONS:
+            self.notify(
+                _PARTIAL_FAILURE_WORKER_NOTIFICATIONS[worker_name],
+                severity="warning",
+            )
+        else:
+            # loud by default: a worker added later is an error modal until
+            # someone decides its failures are benign.
+            self._push_error_modal(
+                title="Unexpected Error",
+                header="A background task failed. Harlequin is still running.",
+                error=worker_error,
+            )
 
     @on(HarlequinTree.CatalogError)
     def handle_catalog_error(self, message: HarlequinTree.CatalogError) -> None:
--- a/src/harlequin/app.py
+++ b/src/harlequin/app.py
@@ -697,6 +697,12 @@ class Harlequin(AppBase):
             callback = partial(self.post_message, message)
             self.set_timer(delay=0.5, callback=callback)
 
+    @on(CodeEditor.SymbolsFound)
+    def load_catalog_items_named_by_buffer(
+        self, message: CodeEditor.SymbolsFound
+    ) -> None:
+        self.data_catalog.database_tree.load_items_named(message.symbols.names)
+
     @on(QueriesExecuted)
     def fetch_data_or_reset_table(self, message: QueriesExecuted) -> None:
         if message.cursors:  # select query
--- a/src/harlequin/app.py
+++ b/src/harlequin/app.py
@@ -171,6 +171,22 @@ class TunnelClosed(Message):
         self.notice = notice
 
 
+class TunnelReconnected(Message):
+    """The dropped tunnel is back, and what runs through it is a new session."""
+
+    def __init__(self, connection: HarlequinConnection) -> None:
+        super().__init__()
+        self.connection = connection
+
+
+class TunnelUnrecoverable(Message):
+    """The tunnel dropped and would not come back, in ssh's own words."""
+
+    def __init__(self, error: BaseException) -> None:
+        super().__init__()
+        self.error = error
+
+
 class TransactionModeChanged(Message):
     def __init__(self, new_mode: HarlequinTransactionMode | None) -> None:
         super().__init__()
@@ -827,6 +843,29 @@ class Harlequin(AppBase):
         message.stop()
         self.notify(message.notice, title="SSH tunnel", severity="error")
 
+    @on(TunnelReconnected)
+    def report_tunnel_reconnected(self, message: TunnelReconnected) -> None:
+        message.stop()
+        self.post_message(
+            TransactionModeChanged(new_mode=message.connection.transaction_mode)
+        )
+        self.notify(
+            "The tunnel dropped and has been reopened, so this is a new "
+            "session: an open transaction, a temp table, and anything set with "
+            "SET went with the old one.",
+            title="SSH tunnel",
+            severity="warning",
+        )
+
+    @on(TunnelUnrecoverable)
+    def report_tunnel_unrecoverable(self, message: TunnelUnrecoverable) -> None:
+        message.stop()
+        self._push_error_modal(
+            title="SSH Tunnel Error",
+            header="Harlequin could not reopen the SSH tunnel.",
+            error=message.error,
+        )
+
     @on(TransactionModeChanged)
     def update_transaction_button_label(self, message: TransactionModeChanged) -> None:
         message.stop()
@@ -1127,6 +1166,8 @@ class Harlequin(AppBase):
         description="Executing queries.",
     )
     def _execute_query(self, message: QuerySubmitted) -> None:
+        # ahead of reading the connection: a tunnel that dropped took it too
+        self._recover_tunnel()
         if self.connection is None:
             return
         cursors: Dict[str, ExecutedStatement] = {}
@@ -1183,6 +1224,28 @@ class Harlequin(AppBase):
                 error=e,
             )
         self.post_message(QueriesCanceled())
+
+    def _recover_tunnel(self) -> None:
+        """Reopen a tunnel that dropped, and the connection that died with it.
+
+        Called on a worker's thread, ahead of the database work it was about to
+        do: restarting `ssh` is not enough on its own, because the adapter's TCP
+        connection ran *through* the old forward. One attempt, and none at all
+        once one has failed -- the user retries by running their query again.
+        """
+        tunnel = self.ssh_tunnel
+        if tunnel is None or not tunnel.needs_restart or self.connection is None:
+            return
+        try:
+            tunnel.restart()
+            connection = self.adapter.connect()
+        except BaseException as e:  # adapters are third-party code
+            self.post_message(TunnelUnrecoverable(e))
+            return
+        # assigned here rather than by the handler: the worker that called this
+        # is about to run a query on it, and cannot wait for a message to land.
+        self.connection = connection
+        self.post_message(TunnelReconnected(connection=connection))
 
     def _get_selected_queries(self) -> list[str]:
         if self.editor is None:
@@ -1316,6 +1379,7 @@ class Harlequin(AppBase):
 
     @work(thread=True, exclusive=True, exit_on_error=False, group="schema_updaters")
     def update_schema_data(self) -> None:
+        self._recover_tunnel()
         if self.connection is None:
             return
         catalog = self.connection.get_catalog()
--- a/src/harlequin/app.py
+++ b/src/harlequin/app.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import os
 import sys
+import threading
 import time
 from functools import partial
 from pathlib import Path
@@ -286,6 +288,15 @@ class Harlequin(AppBase):
             )
         self.query_timer: Union[float, None] = None
         self.connection: HarlequinConnection | None = None
+        self._recovery_lock = threading.Lock()
+        """Held across reopening the tunnel and the connection through it.
+
+        Two workers reaching recovery at once would otherwise each open a
+        connection, and one of them would be closed out from under the worker
+        still running on it.
+        """
+        self._recovered_connection: HarlequinConnection | None = None
+        """The connection a recovery opened, for a worker that waited on it."""
         self.harlequin_driver = HarlequinDriver(app=self)
         self._completer_merge_timer: Timer | None = None
         self._pending_completer_items: list[tuple[CatalogItem, list[CatalogItem]]] = []
@@ -377,15 +388,17 @@ class Harlequin(AppBase):
         self.run_query_bar.apply_configured_limit()
 
         if self.ssh_tunnel is not None:
-            # which database this session is actually looking at, which the
-            # connection details alone no longer say
+            # which database this session is actually looking at
+            warnings = self.ssh_tunnel.warnings()
             self.notify(
-                self.ssh_tunnel.notice(),
+                "\n\n".join((self.ssh_tunnel.notice(), *warnings)),
                 title="SSH tunnel",
-                severity="warning" if self.ssh_tunnel.reused else "information",
+                severity=(
+                    "warning" if self.ssh_tunnel.reused or warnings else "information"
+                ),
+                markup=False,
             )
-            # posted from the watcher's thread, which is why it is a message
-            self.ssh_tunnel.watch(self._report_tunnel_closed)
+            self.ssh_tunnel.watch(self._post_tunnel_closed)
 
         self._connect()
         self._load_catalog_cache()
@@ -839,19 +852,31 @@ class Harlequin(AppBase):
                 self.editor.focus()
         self.data_catalog.disabled = sidebar_hidden
 
-    def _report_tunnel_closed(self, notice: str) -> None:
-        """Called on the tunnel's watcher thread, so it only posts a message."""
-        self.post_message(TunnelClosed(notice))
+    def _post_tunnel_closed(self, notice: str) -> None:
+        """Called on the tunnel's watcher thread, so it only posts a message.
+
+        A child that dies during shutdown reaches a loop that is already
+        closing, which raises rather than delivering.
+        """
+        with contextlib.suppress(RuntimeError):
+            self.post_message(TunnelClosed(notice))
 
     @on(TunnelClosed)
-    def report_tunnel_closed(self, message: TunnelClosed) -> None:
-        """Say the forward is gone, rather than let queries fail unexplained."""
+    def notify_tunnel_closed(self, message: TunnelClosed) -> None:
         message.stop()
-        self.notify(message.notice, title="SSH tunnel", severity="error")
+        # ssh quotes a server's own disconnect message and a helper's output,
+        # so the notice is text rather than markup
+        self.notify(message.notice, title="SSH tunnel", severity="error", markup=False)
 
     @on(TunnelReconnected)
     def report_tunnel_reconnected(self, message: TunnelReconnected) -> None:
         message.stop()
+        replaced, self.connection = self.connection, message.connection
+        if replaced is not None and replaced is not message.connection:
+            # its socket died with the forward, but an adapter does real work
+            # here -- thread pools, temp files -- and may raise on the way out
+            with contextlib.suppress(Exception):
+                replaced.close()
         self.post_message(
             TransactionModeChanged(new_mode=message.connection.transaction_mode)
         )
@@ -1173,8 +1198,8 @@ class Harlequin(AppBase):
     )
     def _execute_query(self, message: QuerySubmitted) -> None:
         # ahead of reading the connection: a tunnel that dropped took it too
-        self._recover_tunnel()
-        if self.connection is None:
+        connection = self._connection_for_worker()
+        if connection is None:
             return
         cursors: Dict[str, ExecutedStatement] = {}
         ddl_queries: list[str] = []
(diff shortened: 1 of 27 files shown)


src/harlequin/app.py
```diff
@@ -1190,6 +1231,28 @@ def _cancel_query(self) -> None:
...
+        """
+        tunnel = self.ssh_tunnel
+        if tunnel is None or not tunnel.needs_restart or self.connection is None:
+            return
+        try:
+            tunnel.restart()
+            connection = self.adapter.connect()
+        except BaseException as e:  # adapters are third-party code
+            self.post_message(TunnelUnrecoverable(e))
+            return
```
tconbeer: **A failed recovery returns, and the worker then runs on the dead connection.**

`_recover_tunnel()` returns `None` on both paths, so the caller can't tell "nothing to do" from "recovery failed." After the `except` posts `TunnelUnrecoverable` and returns, `self.connection` is not `None` — it's the *stale* connection that died with the forward — so `_execute_query`'s `if self.connection is None: return` doesn't fire and it goes on to `execute(connection=self.connection, ...)`.

Failure scenario: tunnel drops, bastion is down, user hits Run Query. They get the "SSH Tunnel Error" modal, and then every statement fails against the dead connection, raising a second `QueryError` modal on top of it. Same via `update_schema_data` → `CatalogError`.

The functional test doesn't catch this because the `app` fixture is DuckDB `:memory:` (`tests/conftest.py:155`) — nothing actually runs through the tunnel, so the "stale" connection still works and `get_catalog()` succeeds. That's why `assert len(app.screen_stack) == 2` holds there but wouldn't in production.

Suggest returning the connection so the callers can't get this wrong:

```python
def _connection_for_worker(self) -> HarlequinConnection | None:
    """The connection to run on, reopening a dropped tunnel first."""
    tunnel = self.ssh_tunnel
    if tunnel is None or not tunnel.needs_restart or self.connection is None:
        return self.connection
    try:
        tunnel.restart()
        connection = self.adapter.connect()
    except BaseException as e:  # adapters are third-party code
        self.post_message(TunnelUnrecoverable(e))
        return None
    self.post_message(TunnelReconnected(connection=connection))
    return connection
```

Each caller then becomes `connection = self._connection_for_worker()` / `if connection is None: return`, and uses the value it was handed. That also lets `TunnelReconnected`'s handler own `self.connection = message.connection` the way `initialize_app` (`app.py:457-458`) already does, which restores the "handlers own all state changes" invariant AGENTS.md states — the comment on 1251-1252 is only true because the helper throws the connection away instead of returning it.

---

src/harlequin/ssh.py
```diff
@@ -478,6 +494,32 @@ def start(self) -> None:
...
+        Raises: HarlequinSshError, having marked this tunnel as not to be tried
+        again.
+        """
+        self.echo_stderr = False
+        if "BatchMode=yes" not in self.argv and len(self.argv) > 1:
+            # ahead of the destination, which ssh takes last
+            self.argv = [*self.argv[:-1], "-o", "BatchMode=yes", self.argv[-1]]
+        self.stop()
+        try:
+            self.start()
```
tconbeer: **Two workers can enter recovery at once, and the loser proceeds on the dead connection.**

`@work(exclusive=True, group=...)` is exclusive *per group*: `_execute_query` is `query_runners` and `update_schema_data` is `schema_updaters`, so they run concurrently. Neither `restart()` nor `_recover_tunnel()` takes a lock (there's no `Lock` anywhere in `ssh.py`).

The wide window is the benign-looking one. `start()` clears `self.dropped` at its *top* (`ssh.py:468`), before `Popen` and before the readiness poll. So for the whole duration of thread A's restart — up to `self.timeout`, 60s by default, plus `adapter.connect()` — thread B sees `needs_restart == False`, skips recovery, and runs `self.connection.get_catalog()` on the connection that died with the old forward. B neither waits nor recovers; it just fails.

Narrower but worse: if B enters between A's `needs_restart` check and A's `start()` clearing `dropped`, both call `restart()`. B's `stop()` sets `self._process = None` while A is inside `_wait_until_ready`, which does `assert self._process is not None` and `self._process.poll()` on shared state — an `AssertionError`/`AttributeError` that surfaces as an "SSH Tunnel Error" modal quoting a Python internal rather than ssh. Two `adapter.connect()` calls also mean one connection is created and immediately orphaned.

A `threading.Lock` held across restart-and-reconnect, with the `needs_restart` check re-read inside it, makes B either wait for A's new connection or skip cleanly. Worth deciding deliberately either way — if concurrent recovery is meant to be impossible, it'd help to say what enforces that.

---

src/harlequin/app.py
```diff
@@ -1190,6 +1231,28 @@ def _cancel_query(self) -> None:
...
+        # assigned here rather than by the handler: the worker that called this
+        # is about to run a query on it, and cannot wait for a message to land.
+        self.connection = connection
```
tconbeer: The connection being replaced is never closed. Every other teardown goes through `self.connection.close()` (`app.py:1056-1057`); here the old object is dropped on the floor.

Its socket is already dead, so this isn't a leak of anything the database can see — but adapters do non-trivial work in `close()` (thread pools, temp files, driver-side cleanup), and after a few reconnects over a long session that adds up. Worth a `contextlib.suppress`-wrapped `close()` on the old one before rebinding, since a dead connection's `close()` may itself raise.

If you take the `_adopt_connection()` extraction suggested elsewhere — shared with `initialize_app`, which posts the same `TransactionModeChanged` — this becomes one line in one place rather than a second thing to remember.

---

src/harlequin/ssh.py
```diff
@@ -478,6 +494,32 @@ def start(self) -> None:
...
+        except HarlequinSshError:
+            # still down, and now not to be tried again
+            self.dropped = True
+            self.restart_failed = True
```
tconbeer: `restart_failed` is redundant with what `start()` already does. `start()` sets `self.dropped = False` on entry (`ssh.py:468`) and the failure path (`except BaseException: self.stop(); raise`) never sets it back — so when `start()` raises here, `dropped` is **already** `False`, and `needs_restart` is already `False`. This block sets `dropped = True` and then adds a second flag whose only job is to cancel the first.

Deleting lines 518-520 gives identical `needs_restart` behavior. Going further, a single `needs_restart` attribute — set `True` in `_watch()`, cleared in `start()` — removes both flags and the property, and the "not tried twice" behavior falls out of the ordering instead of being two flags that have to agree.

Keep `dropped` distinct from `not running` if you keep it, though: after a reuse (`self.reused`) the child has also exited, so deriving it from `running` would report a working reused tunnel as dropped.

One caveat on `except HarlequinSshError` specifically: if `start()` ever raises anything else, the tunnel stays in a state where `needs_restart` is `True` and every subsequent database action retries — the retry storm this is designed to prevent. `except BaseException` (re-raising) would be the safer guard.

---

src/harlequin/ssh.py
```diff
@@ -478,6 +494,32 @@ def start(self) -> None:
...
+        Whatever `--ssh-batch-mode` said at start-up, a restart runs under
+        `BatchMode=yes`: by now something owns the terminal, so a passphrase
+        prompt has nowhere to go, and failing with the credential ssh wanted is
+        the only outcome a user can act on. For the same reason ssh's stderr
+        stops passing through to ours -- there is no longer a stream anyone is
+        reading -- and goes back into the error instead.
+
+        Raises: HarlequinSshError, having marked this tunnel as not to be tried
+        again.
+        """
```
tconbeer: Eight lines of prose over a ten-line method. Two of the whys are real and should survive (`BatchMode` because the TUI owns the terminal; stderr stops echoing for the same reason). The rest is narration that AGENTS.md puts in the PR description rather than the source — "and failing with the credential ssh wanted is the only outcome a user can act on" records why that outcome was chosen over another.

```python
"""Open it again under `BatchMode=yes`, with ssh's stderr in the error.

The TUI owns the terminal by now, so a passphrase prompt has nowhere to
go and nobody is reading ssh's stderr.

Raises: HarlequinSshError, after marking this tunnel as not to be retried.
"""
```

Same note on `needs_restart` above (`ssh.py:405-411`): "a background retry storm against a bastion that is down is how an account gets locked" earns its place — it's the non-obvious reason a property latches false, and without it someone will "fix" it. The tail, "so it fails once with ssh's message and the user restarts Harlequin, having been told why," is the decision-recording half.

Also `# still down, and now not to be tried again` (line 518) restates `self.dropped = True; self.restart_failed = True` — the names already say it.

Separately on line 511: `len(self.argv) > 1` is dead weight (argv always ends with a host, by `build_argv`), and the splice re-implements placement knowledge `build_argv` already owns at `ssh.py:249-250`. Consider making batch-mode a `start(*, batch_mode=False, echo_stderr=True)` parameter instead of permanently mutating `self.argv` and `self.echo_stderr` — the object currently ends up misreporting what it was configured with, and `test_a_restart_never_asks_for_a_credential` asserts on that mutation rather than on behavior.

---

tests/unit_tests/test_ssh.py
```diff
@@ -71,6 +71,19 @@ def listener() -> Iterator[int]:
         yield int(sock.getsockname()[1])
 
 
+@pytest.fixture
+def drop_trigger(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> Path:
+    """Touch the path this returns, and the fake client drops its forwards.
+
+    A file rather than a clock: a child that dropped on a timer could beat the
+    readiness poll on a loaded machine, and the test would fail as a start-up
+    error rather than exercising the drop.
```
tconbeer: "A file rather than a clock" names an approach that was replaced earlier in this stack (`8b685cb test(ssh): drop the fake tunnel on a file, not on a clock`). Once this merges the timer isn't visible anywhere, and AGENTS.md's rule is explicit that this reads to the next person as a live distinction — they'll go looking for the timer.

It's also a duplicate that has diverged from its own twin: the identical fixture at `test_ssh_recovery.py:45-46` carries only the one-line version, which is the right one. Suggest dropping the second paragraph so the two copies match:

```python
"""Touch the path this returns, and the fake client drops its forwards."""
```

Same "rather than [the replaced approach]" phrasing in `tests/data/unit_tests/ssh/ssh:137-138`; the second clause there is load-bearing, the first isn't:

```python
# the test controls when the forward drops, so nothing races the readiness poll
```

While here — `wait_until`, `free_port` and `accepts` are now byte-identical in this file and in `test_ssh_recovery.py`, and `dropping_tunnel` re-implements `child_tunnel()`. No test module in this repo imports from another; the convention is `tests/conftest.py`, which already serves both directories. Worth hoisting these (and building the fake-ssh path off the existing `data_dir` fixture — the two files currently hardcode it with a different number of `.parent` hops).

---

tests/functional_tests/test_ssh_recovery.py
```diff
@@ -0,0 +1,145 @@
...
+            monkeypatch.setenv("FAKE_SSH_EXIT", "255")
+
+            app.action_refresh_catalog()
+            await wait_for_workers(app)
+            await pilot.pause()
+            assert not dropping_tunnel.needs_restart
+            # said once, in ssh's own words
+            assert len(app.screen_stack) == 2
+            await pilot.press("escape")
+            await pilot.pause()
```
tconbeer, line 135: This is the assertion that doesn't pin what it claims. The comment says "said once, in ssh's own words," but `len(app.screen_stack) == 2` is a screen count — it doesn't inspect the modal's text at all, so nothing here would notice if the error stopped quoting ssh.

More importantly it passes for a reason that won't hold in production: the `app` fixture is DuckDB `:memory:`, so no connection actually runs through the tunnel. After the failed restart, `update_schema_data` continues on the "stale" connection, which still works, and `get_catalog()` succeeds — hence exactly one modal. With a real tunnelled connection the stale one is dead, `get_catalog()` raises, and a second `CatalogError` modal lands on the stack. See the note on `_recover_tunnel` for the underlying issue.

If `_recover_tunnel` grows a failure signal, this test can assert the stronger thing: that the worker *aborted* rather than that only one screen was pushed. Asserting on the modal's rendered text would also make the "in ssh's own words" claim real.

Minor: `# said once, in ssh's own words` → `# the error modal, once`, and the docstring on line 120 restates the source's rationale that the test name already covers.

---

src/harlequin/app.py
```diff
@@ -1322,6 +1385,7 @@ def _build_completers(self, catalog: Catalog) -> None:
 
     @work(thread=True, exclusive=True, exit_on_error=False, group="schema_updaters")
     def update_schema_data(self) -> None:
+        self._recover_tunnel()
```
tconbeer: Two of roughly nine workers that touch the database call `_recover_tunnel()`, and the mechanism is a pasted first line — which is the shape that gets forgotten when the next worker is added. Named-accessor form (`connection = self._connection_for_worker()`) fixes that by construction, since it'd be the only way to get a connection on a thread.

Omitting `_cancel_query`, `commit`/`rollback` and `toggle_transaction_mode` reads as deliberate and right — cancelling or committing a session that no longer exists is meaningless.

The one real gap is the Data Catalog's lazy loader: `DatabaseTree._load_children` (`components/data_catalog/database_tree.py:467`) reaches the database through `item.fetch_children()` on connections the adapter captured when the catalog was built (`harlequin_duckdb/catalog.py:45`, etc.), which `_recover_tunnel` can't reach at all. Expanding an unloaded catalog node after a drop fails with a Catalog Error and no recovery, and because those items hold the *old* connection object they stay broken until the next full `update_schema_data` replaces the tree.

Not necessarily this PR's to fix — but since §5.4 says recovery happens "before the next thing that needs the database (a query, a catalog refresh)," it's worth either handling or naming as a known limit alongside the §8 deferral.

---

tconbeer (commented):
## Review: SSH tunnels 4/5

Reviewed through four explicit lenses: **correctness/concurrency/lifecycle**, **reuse & simplification**, **security**, and **comments/docstrings** (against AGENTS.md's conventions). Reviewed as a diff against 3/5, not against `main`.

The design is sound and matches §5.4 of the plan point for point — both halves or neither, lazy rather than timed, batch-mode on restart, one attempt per action. The watcher/restart handoff is genuinely race-safe: `_watch()` guards on `self._process is not process` and `stop()` nulls `_process` before `start()` reassigns it, so a stale watcher can't fire a spurious "tunnel closed" after a restart. `cache_material()` is unchanged by a restart, so `connection_hash` stays stable and the catalog/history caches don't split across a reconnect. All 84 tests in `test_ssh.py` + `test_ssh_recovery.py` pass locally on 3.10 (12.6s), and CI is green.

Two things I'd want resolved before merge, both about what happens *around* a recovery rather than the recovery itself:

1. **A failed recovery doesn't stop the worker.** `_recover_tunnel()` returns `None` on both the "nothing to do" and "recovery failed" paths, so after `TunnelUnrecoverable` the caller's `if self.connection is None` guard doesn't fire and it runs the query on the stale connection — a second error modal on top of the first. Returning the connection fixes this and, as a bonus, lets the handler own the `self.connection` assignment the way `initialize_app` already does.

2. **Recovery has no mutual exclusion, and `dropped` clears too early.** `_execute_query` and `update_schema_data` are in different `@work` groups, so `exclusive=True` doesn't serialize them. `start()` clears `dropped` at its top, so for the whole duration of one thread's restart (up to 60s + connect) the other sees `needs_restart == False`, skips recovery, and fails on the dead connection. A narrower interleaving lets both call `restart()` and trips an assert on shared `_process` state.

Also worth a look: the replaced connection is never `close()`d; `restart_failed` is redundant with what `start()` already does to `dropped`; and the Data Catalog's lazy loader holds adapter-captured connections that recovery can't reach, so expanding an unloaded node after a drop still fails.

On prose — the diff is carrying one design rationale written out three or four times across `ssh.py`, `app.py`, and both test files. AGENTS.md gives the *why* one home. One comment ("a file rather than a clock") points at the timer approach that `8b685cb` removed earlier in this stack, which is the "reads as a live distinction" rule specifically.

**Security: nothing found.** `restart()` introduces no new user-controlled argv — `_refuse_unsafe_value` still gates every value at build time, and the splice inserts only a literal. The unredacted `adapter.connect()` exception in the error modal is the same behavior the existing `_connect` failure path already has, so it's consistent rather than new.

**Changelog:** correctly deferred to 5/5. Parts 1-4 are slices of one unshipped feature; four entries narrating the stack would be the implementation detail the rule forbids. When 5/5 writes it, the session-loss consequence (open transactions, temp tables, `SET` values) is the user-visible part worth a clause.

## Not checked

- **I did not review 1/5, 2/5 or 3/5.** This is a diff against 3/5's head; anything inherited from the earlier parts (argv construction, `_Stderr`, the readiness poll, `build_tunnel`) I read only as context for this change, not as a review of it.
- **I did not test against a real SSH server or a real tunnelled database.** Every test here — and every test I ran — uses the fake client and a DuckDB `:memory:` connection, so "the adapter's TCP connection died with the forward," the premise the whole feature rests on, is asserted by no test in this repo. That's the gap I'd most want closed, and it's why the failed-restart test passes for the wrong reason.
- **I did not verify behavior on Windows.** Both new test modules skip there (`sys.platform == "win32"`), so the recovery path is untested on the one platform where `SshTunnel`'s process handling differs most.
- **I did not run the full suite or check snapshots** — only the two SSH modules. I relied on CI for the rest.
- **I did not reason about `--ssh-allow-reuse` interacting with recovery.** A reused listener's child exits immediately, so `_arm_watcher` never arms and `dropped` should stay false, but I did not trace what happens if a reused tunnel's *external* owner goes away.
- **I did not verify the `os._exit`/atexit interaction** between a restarted tunnel and `hsql`'s timeout machinery; `hsql` doesn't use recovery, but `atexit.register(self.stop)` is re-registered per `start()` and I didn't check for duplicate registrations across restarts.

Findings are inline. A clean-ish review isn't the merge gate here — the intent question (is lazy, one-shot recovery the behavior you want users to get?) is yours.

---

src/harlequin/ssh.py
tconbeer: Reproduced this, and it's worse than I described above — upgrading it from "reasoned from the code" to confirmed, with a second failure mode I'd missed.

Two threads calling `restart()` on a dropped tunnel behind a `threading.Barrier`, against the fake client:

```
[HarlequinSshError('ssh exited with code 255 without opening the forward.\n\n
                    bind [127.0.0.1]:57625: Address already in use'),
 AttributeError("'NoneType' object has no attribute 'poll'")]
```

The `AttributeError` is the interleaving I predicted — B's `stop()` nulls `self._process` while A is inside `_wait_until_ready`.

**The `bind: Address already in use` is the one I want to flag.** A's child takes the forwarded port successfully; B's child then can't bind it and exits 255. `ExitOnForwardFailure=yes` is doing exactly its job. But that surfaces in B as a `HarlequinSshError`, which means B's `except HarlequinSshError` sets `restart_failed = True` — so the tunnel ends up **permanently marked never-retry-again while the forward is actually open and working**.

User-visible: a spurious "SSH Tunnel Error" modal saying the tunnel could not be reopened, on a tunnel that is up; and from then on `needs_restart` is `False` forever, so the *next* genuine drop in that session is never recovered. The user is told to restart Harlequin over a tunnel that had already come back.

That makes the ordering point sharper than "add a lock, either way is fine": the `needs_restart` check and the restart-plus-reconnect need to be atomic, or a losing thread has to distinguish "I lost a race" from "ssh won't come back." Serializing on a lock and re-reading `needs_restart` inside it gives you both — the loser sees `needs_restart == False` with a live tunnel and simply proceeds.

Repro is a scratch file, not proposed for the branch — happy to clean it up into a real test if you want one pinning this.

---

tconbeer:
Both of the big ones were real, and your follow-up repro changed the shape of the fix. All eight addressed.

## The failed recovery that let the worker run on

Fixed as you suggested — `_recover_tunnel()` is now `_connection_for_worker()`, returning the connection to run on, and each caller uses the value it was handed rather than reading `self.connection` back. `TunnelReconnected`'s handler owns the assignment, which restores the "handlers own all state changes" invariant.

One thing that only showed up once the test could see it: returning `self.connection` whenever `needs_restart` is false is *also* wrong, because `needs_restart` latches false after a restart fails. So the helper distinguishes the two:

```python
if not tunnel.needs_restart:
    if tunnel.dropped:
        # a restart already failed and is not tried again, so what
        # is left is a connection whose socket went with the forward
        return None
    return self._recovered_connection or self.connection
```

Your point about the functional test was exactly right, and it is why I found this: the `app` fixture is DuckDB `:memory:`, so the stale connection kept working and the screen count kept passing. The test now asserts `app._connection_for_worker() is None` after the failed restart, which is the thing that would actually be wrong in production, and asserts the modal's text quotes ssh rather than counting screens.

## The concurrency

Your follow-up is what settled the design — the spurious `restart_failed` on a *working* tunnel is much worse than the `AttributeError`, because it silently disarms recovery for the rest of the session.

Two locks, and they have distinct jobs, which I have said in the code so the next reader does not have to guess which is load-bearing:

- **`SshTunnel._restart_lock`**, held across `restart()`, with `needs_restart` re-read inside it. The loser wakes to a tunnel that is already up, sees `needs_restart == False`, and returns — no bind race, no latch. This is the one your repro hits, and it makes the class safe for an out-of-tree front end rather than only for this one.
- **`Harlequin._recovery_lock`**, held across restart *and* `adapter.connect()`. Without it the loser would still open a second connection, and the handler would close one out from under the worker still running on it.

`tests/unit_tests/test_ssh.py::test_two_workers_reopening_at_once_do_not_bury_a_working_tunnel` is your repro, cleaned up onto the branch: two threads behind a `threading.Barrier`. I checked it fails without the lock (`assert tunnel.running` → False).

## The rest

- **The replaced connection is closed**, from the handler, under `contextlib.suppress(Exception)` — you are right that a dead connection's `close()` may itself raise, and that adapters do real work there.
- **`except HarlequinSshError` → `except BaseException`** (re-raising). Anything else `start()` raised would have left `needs_restart` true and every database action retrying, which is the storm the property exists to prevent.
- **`restart()`'s eight lines of prose** are down to your four, and `needs_restart`'s docstring keeps the account-lockout half and drops the decision-recording tail. `# still down, and now not to be tried again` is gone.
- **`len(self.argv) > 1`** is gone, and — taking your last suggestion — **batch mode and the stderr echo are arguments to `start()`** rather than a permanent mutation. `self.argv` keeps reporting what the tunnel was configured with, and `test_a_restart_never_asks_for_a_credential` now asserts on what ssh was actually run with (via `FAKE_SSH_ARGV`) rather than on the mutation.

I kept `dropped` and `restart_failed` as two attributes rather than collapsing to one. You are right that deleting the two assignments gives identical `needs_restart` behaviour, since `start()` clears `dropped` on entry and the failure path never sets it back — but then `dropped` reads False on a tunnel that is demonstrably down, and it is a public signal, not just an input to the property. Two attributes that answer two questions ("is the forward gone" and "should the next action try") seemed better than one that answers the second while being named for the first.

- **The catalog's lazy loader.** Confirmed and not fixed here — `fetch_children()` runs on connections the adapter captured when the tree was built, which recovery cannot reach without going through each adapter's items. It is named as a known limit in §8's deferred table, with what would bring it back.

## Tests

- `drop_trigger` is in `tests/conftest.py`, which already serves both directories, so the two copies cannot diverge again; its docstring is the one-line version, and the fake client's comment keeps only the load-bearing clause.
- `_modal_text()` asserts the error quotes ssh.

I did **not** hoist `free_port`, `accepts` and `wait_until`. They are genuinely duplicated, but conftest's only sharing mechanism is fixtures, and turning three one-line helpers into fixtures means threading a parameter through ~25 call sites for tests that read worse afterwards. Happy to do it if you would rather; I did add a `fake_ssh_client` fixture off `data_dir` so at least the path is not hardcoded with two different `.parent` counts.

---

1476 tests pass on the stack tip, with mypy, ruff and the import-linter contracts. Merged forward through 5/6 and 6/6, and the stack now sits on `main` at 8187f0f.

---

PR #1122, opened 2026-09-02
fix: surface errors from unrecognized background workers (#1117)

### LINKED ISSUE #1117: `_handle_worker_error` silently swallows errors from every worker it doesn't name
Found while planning #687.

`Harlequin._handle_worker_error` (`src/harlequin/app.py:576`) branches on two worker names and has no `else`:

```python
async def _handle_worker_error(self, message: Worker.StateChanged) -> None:
    if message.worker.name == "update_schema_data" and message.worker.error is not None:
        ...   # error modal
    elif message.worker.name == "_connect" and message.worker.error is not None:
        ...   # exit(2)
```

Every worker is declared `exit_on_error=False`, so a failure in any *other* worker doesn't crash the app — it produces no modal, no notification, and no log line. The user sees the UI simply not do the thing they asked for.

Two shapes this can take today:

- Workers that route here and vanish: `_execute_query` guards the fetch loop but not the code around it, and `_fetch_data`, `_load_catalog_cache`, the completer merge and the export worker all have paths that can raise past their own handling.
- Workers that go the *other* way: `components/data_catalog/database_tree.py:498` (`_loader`, the background loading queue processor) and `components/code_editor.py:88` (`read_symbols`, tree-sitter over the buffer) are the only two `@work` decorators in `src/` that **omit** `exit_on_error=False`, so they take the whole app down through `_handle_exception`, wrapped in `WorkerFailed`. (#687's crash handler unwraps that so the report names the real error, but these two should probably be deciding this deliberately rather than by omission.)

### What this needs

Not just an `else` branch — an audit. The reason this is filed separately from #687 rather than folded into it is that "route every unrecognized worker error to an `ErrorModal`" would turn ordinary shutdown cancellations into modal noise, so it needs a pass over which workers can fail benignly:

- Which workers can fail as part of normal operation (cancellation at shutdown, an exclusive worker superseded mid-flight) and should stay quiet.
- Which should surface as an `ErrorModal` via the existing `_push_error_modal`.
- Which should surface as a notification instead, because the failure is partial and the app is still usable.
- Whether `_loader` and `read_symbols` want `exit_on_error=False` — a tree-sitter parse failing over a weird buffer taking down the whole app looks like an oversight.

Then add the `else` that catches whatever is left over, so the next worker someone adds is loud by default rather than silent by default.

Related: #687 (never crash from an uncaught exception).

visheshgubrani:
`_handle_worker_error` handled only `update_schema_data` and `_connect`. Errors from every other app-owned worker disappeared without a modal, notification, or log. Query worker failures could also leave the run bar stuck in its non-responsive state.

`DatabaseTree._loader` and `CodeEditor.read_symbols` also inherited Textual's default `exit_on_error=True`, allowing failures in these background tasks to crash the entire app.

Closes #1117

* Route query worker failures to a "Query Error" modal and restore the run bar and results viewer.
* Route transaction-mode failures to a "Transaction Error" modal.
* Show warning notifications for cache and completion failures because these are partial failures and the app remains usable.
* Show a generic error modal for unknown worker failures, making future workers loud by default.
* Ignore errors delivered while the app is already exiting.
* Set `exit_on_error=False` for symbol scanning and the database-tree loader.
* Handle component-owned worker failures within their owning widgets because their `Worker.StateChanged` messages do not bubble to the app.
* Suppress repeated symbol-scan warnings until a scan succeeds.
* Preserve the database loader's existing per-item recovery while stopping and reporting unexpected failures in the outer loader.

The response depends on how much functionality failed. Query execution, transaction changes, and unexpected workers use modals because the requested operation failed completely. Cache, completion, and symbol-scan failures use warnings because Harlequin remains usable.

A single generic modal for every worker was considered, but it would make optional background failures unnecessarily disruptive. Allowing unknown failures to remain silent was also rejected because newly added workers would repeat the original bug.

The database loader deliberately does not broadly catch and continue after an unexpected infrastructure failure. Expected per-item failures already have their own recovery path. Continuing after an unexpected queue or tree-state failure could produce repeated modals or operate on inconsistent state.

* [x] No.
* [ ] Yes; I haven't opened a PR, but the gist of the change is: ...

* [x] Yes.
* [ ] No, I believe tests aren't necessary.

Added functional coverage for:

* Existing `_connect` and `update_schema_data` behavior.
* Query UI recovery for both query workers, including a real worker failure.
* Transaction and partial-failure routing.
* Unknown-worker fallback and shutdown guards.
* Symbol-scan survival and once-per-failure-streak warnings.
* Child catalog-load recovery and unexpected outer-loader failure handling.

* [x] I have added an entry to `CHANGELOG.md`, under the `[Unreleased]` section heading. That entry references the issue closed by this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
 
 - The config wizard no longer pre-fills the unset query limit with `-1` or unset adapter options with `None` (`****` for secrets). Leave the limit blank for app defaults, or enter `-1` for no limit ([#1105](https://github.com/tconbeer/harlequin/issues/1105)).
 - Harlequin and hsql no longer crash when trying to open a file that isn't text ([#1108](https://github.com/tconbeer/harlequin/issues/1108)).
+- Background task failures no longer disappear silently or crash Harlequin; they now show an error dialog or warning as appropriate ([#1117](https://github.com/tconbeer/harlequin/issues/1117)).
 
 ## [2.12.2] - 2026-08-31
 
--- a/src/harlequin/app.py
+++ b/src/harlequin/app.py
@@ -177,6 +177,18 @@ class CompletersReady(Message):
         self.member_completer = member_completer
 
 
+_PARTIAL_FAILURE_WORKER_NOTIFICATIONS: dict[str, str] = {
+    "_load_catalog_cache": (
+        "Harlequin could not load its cache; your query history may be missing."
+    ),
+    "_extend_and_merge_completers": "Harlequin could not update completions.",
+    "_build_completers": "Harlequin could not build completions.",
+}
+"""Toast text for the workers whose failure is partial: the app stays usable,
+so their errors surface as a notification rather than an error modal.
+"""
+
+
 class Harlequin(AppBase):
     """
     The SQL IDE for your Terminal.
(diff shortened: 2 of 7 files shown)


tconbeer (approved):
thanks, this looks great!

PR #1128, opened 2026-09-03
fix(ssh): refresh the Data Catalog after a tunnel reconnects

### LINKED ISSUE #1127: Refresh the Data Catalog after an SSH tunnel reconnects
When an SSH tunnel drops, the next database action restarts `ssh`, calls `adapter.connect()` again, and replaces `Harlequin.connection` (`_connection_for_worker()` in `app.py`). The Data Catalog's tree is not rebuilt, and its `InteractiveCatalogItem`s still hold the connection that died with the old forward — so expanding a node that has not loaded yet fails with a Catalog Error until the user refreshes the catalog by hand with `ctrl+r`.

Harlequin should refresh the Data Catalog itself after a reconnect, so the tree is built on the new connection and the user has nothing to remember.

This was noted as a deferred item in `docs/plans/ssh-tunnels-design.md` §8 ("recovering the Data Catalog's lazy loader"), which suggests the fix belongs in handing catalog items a new connection rather than reaching into each adapter's items. Refreshing the catalog on `TunnelReconnected` may be enough on its own.

Documenting the workaround was cut from the SSH docs in tconbeer/harlequin-web in favor of fixing it here.

visheshgubrani:
Rebuild the catalog on the new connection when a dropped tunnel is reopened, so lazy catalog nodes keep loading without a manual refresh.

Closes #1127

- `Harlequin.report_tunnel_reconnected` now calls `update_schema_data()` after swapping in the recovered connection. That is the same worker `ctrl+r`, commit/rollback, and adapter-driven catalog refreshes use: it runs `get_catalog()` on the new connection and posts `NewCatalog`, so the DatabaseTree is rebuilt on the new session and the completers are reseeded. `DatabaseTree.watch_catalog` preserves expanded and highlighted nodes, so the rebuild is the same as a manual refresh.
- A functional regression test in `tests/functional_tests/test_ssh_recovery.py` drops the fake-ssh tunnel and recovers via a query (not a refresh, since a refresh-triggered recovery already rebuilt the tree). It asserts the catalog object was replaced with items bound to the new connection and that expanding an unloaded node still works with no Catalog Error modal. It uses the sqlite app because `HarlequinSqliteConnection.close()` really closes the driver connection, which makes a stale item's fetch fail deterministically.

- Root cause: `InteractiveCatalogItem.fetch_children()` runs on the connection the adapter captured when `get_catalog()` built the tree. Recovery reopens the tunnel and reconnects, but nothing re-pointed or replaced those items, so expanding an unloaded node after a reconnect failed until a manual refresh.
- Alternative considered: hand the existing items a new connection. Catalog items are adapter subclasses that capture connections in their own fields, so core cannot re-point them generically without a new public adapter API. The design doc defers that change (`docs/plans/ssh-tunnels-design.md` section 8); rebuilding from the new connection's `get_catalog()` uses the mechanism a refresh already has.
- Alternative considered: refresh lazily when the first `CatalogError` appears. Rejected: the user sees an error modal first and completions stay stale.
- Tradeoffs: when recovery is itself triggered by a manual refresh, the handler's refresh adds a second `get_catalog()` round trip that converges to the same state. Expanded nodes are re-fetched during the rebuild, exactly as with `ctrl+r` today. The refresh is scoped to the database tree; the file and S3 panes are not touched, since they do not ride the tunnel. The one-attempt-per-drop recovery policy is unchanged.

- [x] No.

- [x] Yes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- When an SSH tunnel reconnects, Harlequin now refreshes the Data Catalog on the new connection, so lazy catalog nodes keep loading without a manual refresh ([#1127](https://github.com/tconbeer/harlequin/issues/1127)).
+
 ## [2.13.0] - 2026-09-02
 
 ### Features
--- a/src/harlequin/app.py
+++ b/src/harlequin/app.py
@@ -922,6 +922,9 @@ class Harlequin(AppBase):
         self.post_message(
             TransactionModeChanged(new_mode=message.connection.transaction_mode)
         )
+        # the tree's items were built on the connection that died with the
+        # forward; rebuild the catalog on this one, or lazy nodes fail to load
+        self.update_schema_data()
         self.notify(
             "The tunnel dropped and has been reopened, so this is a new "
             "session: an open transaction, a temp table, and anything set with "
--- a/tests/functional_tests/test_ssh_recovery.py
+++ b/tests/functional_tests/test_ssh_recovery.py
@@ -14,8 +14,12 @@ from pathlib import Path
 from typing import Awaitable, Callable
 
 import pytest
+from textual.pilot import Pilot
+from textual.widgets._tree import TreeNode
 
 from harlequin import Harlequin
+from harlequin.catalog import CatalogItem, InteractiveCatalogItem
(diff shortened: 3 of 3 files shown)


tconbeer:
Great minds; already had this fix in flight, and #1129 handles a case this one doesn't (when you wake up a tunnel because of a manual ctrl+r). So this one is superseded by #1129

PR #1129, opened 2026-09-03
fix: refresh the Data Catalog after an SSH tunnel reconnects

### LINKED ISSUE #1127: Refresh the Data Catalog after an SSH tunnel reconnects

tconbeer:
Closes #1127

- `TunnelReconnected` carries a `rebuild_catalog` flag, and `report_tunnel_reconnected` rebuilds the Data Catalog when it is set — after it has swapped `self.connection` in, so the completers rebuild against the new connection too. It sets `database_tree.loading` first, matching what `action_refresh_catalog` already does.
- `_connection_for_worker()` takes `rebuilds_catalog`, and `update_schema_data` passes it. That worker is already about to build a tree on the connection it gets back, so a recovery that happens inside it does not ask for a second catalog.
- `update_schema_data` now clears the tree's spinner when it returns without a connection. That is a plain return rather than a worker error, so nothing else stopped it — pre-existing in `action_refresh_catalog`, and this fixes both callers.
- Two functional tests in `tests/functional_tests/test_ssh_recovery.py`: a query-triggered reconnect leaves a catalog whose unloaded nodes still expand, and a refresh-triggered reconnect rebuilds the tree exactly once. A third assertion pins the spinner on the existing unrecoverable-tunnel test.

The tree's `InteractiveCatalogItem`s call `fetch_children()` on the connection the adapter captured when it built them. After a drop, `_connection_for_worker()` restarts `ssh` and replaces `Harlequin.connection`, but the tree still holds items bound to the socket that died with the old forward — so expanding a node that had not loaded yet raised a Catalog Error. Rebuilding the tree on the new connection needs nothing from adapters.

The one subtlety is the flag. `_connection_for_worker()` has two callers, and one of them *is* the catalog rebuild. Refreshing unconditionally would start a second `schema_updaters` worker; since Textual cannot preempt a thread worker, the cancelled one keeps running, and the result is two concurrent `get_catalog()` calls inside one adapter. Passing the caller's intent down is what avoids that.

It does not cover every interleaving, and the docstring says so: a worker that loses `_recovery_lock` gets the recovered connection without posting, so a query and a refresh that hit the same drop still rebuild twice. Inspecting the `schema_updaters` group from the handler covers that case but is not a strict improvement — a schema worker that entered `get_catalog()` on the old connection before the drop was detected is still `RUNNING`, so the handler would stand down and leave a tree built on the dead connection behind the Catalog Error modal. A redundant `get_catalog()` is the milder failure of the two.

**Testing.** The regression test is lifted from #1128, which tested the symptom rather than the mechanism: it drives the recovery with a query, then expands a node that had not loaded and asserts no error modal. It uses the sqlite adapter because `HarlequinSqliteConnection.close()` really closes the driver connection, so an item left on the old one fails deterministically.

- [x] No.

- [x] Yes.

---

Verification: 1478 passed / 7 skipped on 3.10, 3 passed on 3.12, `mypy` clean, `ruff` clean, `lint-imports` 4/4 kept. Each of the three fixes is pinned by an assertion that fails without it.

https://claude.ai/code/session_01SDVnAxw5MNrpaK8W7dm9v4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Bug Fixes
 
-- Harlequin now refreshes the Data Catalog after an SSH tunnel drops and is reopened, so expanding a node no longer raises a Catalog Error until you press `ctrl+r` ([#1127](https://github.com/tconbeer/harlequin/issues/1127)).
+- Harlequin now refreshes the Data Catalog when an SSH tunnel reconnects, so expanding a node no longer raises a Catalog Error ([#1127](https://github.com/tconbeer/harlequin/issues/1127)).
 
 ## [2.13.0] - 2026-09-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- Harlequin now refreshes the Data Catalog after an SSH tunnel drops and is reopened, so expanding a node no longer raises a Catalog Error until you press `ctrl+r` ([#1127](https://github.com/tconbeer/harlequin/issues/1127)).
+
 ## [2.13.0] - 2026-09-02
 
 ### Features
--- a/src/harlequin/app.py
+++ b/src/harlequin/app.py
@@ -176,9 +176,11 @@ class TunnelClosed(Message):
 class TunnelReconnected(Message):
     """The dropped tunnel is back, and what runs through it is a new session."""
 
-    def __init__(self, connection: HarlequinConnection) -> None:
+    def __init__(self, connection: HarlequinConnection, stale_catalog: bool) -> None:
         super().__init__()
         self.connection = connection
+        self.stale_catalog = stale_catalog
+        """Whether the tree still holds items bound to the connection that died."""
 
 
 class TunnelUnrecoverable(Message):
@@ -922,6 +924,12 @@ class Harlequin(AppBase):
         self.post_message(
             TransactionModeChanged(new_mode=message.connection.transaction_mode)
         )
+        if message.stale_catalog:
+            # the tree's items load their children on the connection the adapter
+            # captured when it built them, so expanding an unloaded node would
+            # reach the socket that died with the old forward
+            self.data_catalog.database_tree.loading = True
+            self.update_schema_data()
         self.notify(
             "The tunnel dropped and has been reopened, so this is a new "
             "session: an open transaction, a temp table, and anything set with "
@@ -1298,7 +1306,9 @@ class Harlequin(AppBase):
             )
         self.post_message(QueriesCanceled())
 
-    def _connection_for_worker(self) -> HarlequinConnection | None:
+    def _connection_for_worker(
+        self, rebuilds_catalog: bool = False
+    ) -> HarlequinConnection | None:
         """The connection to run on, reopening a dropped tunnel first.
 
         Called on a worker's thread, ahead of the database work it was about to
@@ -1309,6 +1319,10 @@ class Harlequin(AppBase):
         None is a worker that must not run: a recovery that failed leaves a
         connection whose socket died with the forward, and running on it would
         raise a second modal behind the one that said why.
+
+        `rebuilds_catalog` is the caller about to build a tree on the connection
+        it gets back, so a recovery here does not ask for a second one -- two
+        `get_catalog()` calls at once are two threads inside one adapter.
         """
         tunnel = self.ssh_tunnel
         if tunnel is None or self.connection is None:
@@ -1332,7 +1346,9 @@ class Harlequin(AppBase):
             # what a worker still inside this method runs on; the handler is
             # what makes it the app's, once the message lands
             self._recovered_connection = connection
-        self.post_message(TunnelReconnected(connection=connection))
+        self.post_message(
+            TunnelReconnected(connection=connection, stale_catalog=not rebuilds_catalog)
+        )
         return connection
 
     def _get_selected_queries(self) -> list[str]:
@@ -1467,7 +1483,7 @@ class Harlequin(AppBase):
 
     @work(thread=True, exclusive=True, exit_on_error=False, group="schema_updaters")
     def update_schema_data(self) -> None:
-        connection = self._connection_for_worker()
+        connection = self._connection_for_worker(rebuilds_catalog=True)
         if connection is None:
             return
         catalog = connection.get_catalog()
--- a/src/harlequin/app.py
+++ b/src/harlequin/app.py
@@ -176,11 +176,11 @@ class TunnelClosed(Message):
 class TunnelReconnected(Message):
     """The dropped tunnel is back, and what runs through it is a new session."""
 
-    def __init__(self, connection: HarlequinConnection, stale_catalog: bool) -> None:
+    def __init__(self, connection: HarlequinConnection, rebuild_catalog: bool) -> None:
         super().__init__()
         self.connection = connection
-        self.stale_catalog = stale_catalog
-        """Whether the tree still holds items bound to the connection that died."""
+        self.rebuild_catalog = rebuild_catalog
+        """Whether this handler is the one that has to rebuild the tree."""
 
 
 class TunnelUnrecoverable(Message):
@@ -924,7 +924,7 @@ class Harlequin(AppBase):
         self.post_message(
             TransactionModeChanged(new_mode=message.connection.transaction_mode)
         )
-        if message.stale_catalog:
+        if message.rebuild_catalog:
             # the tree's items load their children on the connection the adapter
             # captured when it built them, so expanding an unloaded node would
             # reach the socket that died with the old forward
@@ -1322,7 +1322,9 @@ class Harlequin(AppBase):
 
         `rebuilds_catalog` is the caller about to build a tree on the connection
         it gets back, so a recovery here does not ask for a second one -- two
-        `get_catalog()` calls at once are two threads inside one adapter.
+        `get_catalog()` calls at once are two threads inside one adapter. It
+        reports the caller's own intent, which the worker that loses the lock
+        below cannot: a query and a refresh that hit the same drop rebuild twice.
         """
         tunnel = self.ssh_tunnel
         if tunnel is None or self.connection is None:
@@ -1347,7 +1349,9 @@ class Harlequin(AppBase):
             # what makes it the app's, once the message lands
             self._recovered_connection = connection
         self.post_message(
-            TunnelReconnected(connection=connection, stale_catalog=not rebuilds_catalog)
+            TunnelReconnected(
+                connection=connection, rebuild_catalog=not rebuilds_catalog
+            )
         )
         return connection
 
@@ -1485,6 +1489,11 @@ class Harlequin(AppBase):
     def update_schema_data(self) -> None:
         connection = self._connection_for_worker(rebuilds_catalog=True)
         if connection is None:
+            # no NewCatalog is coming, and a plain return is not a worker error,
+            # so nothing else stops the spinner a refresh started
+            self.call_from_thread(
+                setattr, self.data_catalog.database_tree, "loading", False
+            )
             return
         catalog = connection.get_catalog()
         self.post_message(NewCatalog(catalog=catalog))
--- a/src/harlequin/app.py
+++ b/src/harlequin/app.py
@@ -150,6 +150,10 @@ class QueriesCanceled(Message):
     pass
 
 
+class CatalogRefreshAborted(Message):
+    """The catalog worker stopped without a connection to build a tree on."""
+
+
 class ResultsFetched(Message):
     def __init__(
         self,
@@ -741,6 +745,10 @@ class Harlequin(AppBase):
     def handle_new_catalog(self, message: NewCatalog) -> None:
         self.data_catalog.update_database_tree(message.catalog)
         self.update_completers(message.catalog)
+
+    @on(CatalogRefreshAborted)
+    def stop_catalog_loading(self) -> None:
+        self.data_catalog.database_tree.loading = False
 
     @on(NewCatalogItems)
     def handle_new_catalog_item(self, message: NewCatalogItems) -> None:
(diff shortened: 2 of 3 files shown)


src/harlequin/app.py
```diff
@@ -176,9 +176,11 @@ def __init__(self, notice: str) -> None:
 class TunnelReconnected(Message):
     """The dropped tunnel is back, and what runs through it is a new session."""
 
-    def __init__(self, connection: HarlequinConnection) -> None:
+    def __init__(self, connection: HarlequinConnection, stale_catalog: bool) -> None:
         super().__init__()
         self.connection = connection
+        self.stale_catalog = stale_catalog
```
tconbeer: Naming nit: the flag isn't really "the catalog is stale" — after a reconnect the tree on screen is *always* stale, including when `update_schema_data` is the worker that recovered; the difference is only whether a rebuild is already in flight. So the name asserts something that isn't true in the `False` case, and a reader has to go find the one caller to learn what it actually gates.

`rebuild_catalog` (or `needs_catalog_rebuild`) says what the handler does with it, and pairs with `rebuilds_catalog` on the other side without the negation. Same for the attribute docstring: something like "Whether this handler is the one that has to rebuild the tree."

---

src/harlequin/app.py
```diff
@@ -1332,7 +1346,9 @@ def _connection_for_worker(self) -> HarlequinConnection | None:
             # what a worker still inside this method runs on; the handler is
             # what makes it the app's, once the message lands
             self._recovered_connection = connection
-        self.post_message(TunnelReconnected(connection=connection))
+        self.post_message(
+            TunnelReconnected(connection=connection, stale_catalog=not rebuilds_catalog)
```
tconbeer: The flag closes the single-worker case but not the one this method's own comment above warns about — two workers arriving at once.

Interleaving: `_execute_query` and `update_schema_data` both reach the drop. The query worker wins `_recovery_lock`, restarts, and posts `stale_catalog=True`. The schema worker was blocked on the lock; it wakes into the `if not tunnel.needs_restart` branch, returns the recovered connection, and goes on to call `get_catalog()` — it never posts, so it can't set the flag. Meanwhile the handler runs `update_schema_data()`, which cancels that worker (a thread worker keeps running) and starts a second one. That is exactly the "two `get_catalog()` calls inside one adapter" this argument is trying to prevent.

Narrow — it needs a query and a catalog refresh to hit the same drop — and I don't think it should block the PR. But it's worth deciding explicitly, because it's the case the rejected alternative (checking the `schema_updaters` group from the handler) *does* cover. Either say so in the docstring ("the caller's intent, which the worker that loses the lock cannot report"), or reconsider the group check on the strength of covering both. Passing the intent down can't reach the loser, since the winner holds the lock for the whole restart.

---

src/harlequin/app.py
```diff
@@ -922,6 +924,12 @@ def report_tunnel_reconnected(self, message: TunnelReconnected) -> None:
         self.post_message(
             TransactionModeChanged(new_mode=message.connection.transaction_mode)
         )
+        if message.stale_catalog:
+            # the tree's items load their children on the connection the adapter
+            # captured when it built them, so expanding an unloaded node would
+            # reach the socket that died with the old forward
+            self.data_catalog.database_tree.loading = True
```
tconbeer, line 939: Good call showing the rebuild — this matches `action_refresh_catalog` and is the thing #1128 leaves out, so a user there just watches a stale tree.

One low-priority hole, inherited from `action_refresh_catalog` rather than introduced here: nothing clears `loading` when the worker returns without posting `NewCatalog`. `_connection_for_worker()` returns `None` if the tunnel has dropped *again* by the time the worker runs, and that's a plain return, not a worker error, so `_handle_worker_error`'s `loading = False` never fires and the tree spins forever. It's slightly more reachable from here than from `ctrl+r`, since we're refreshing in the seconds right after a tunnel event. Worth an early `loading = False` on the `connection is None` path in `update_schema_data`, which would fix both callers at once.

---

tests/functional_tests/test_ssh_recovery.py
```diff
@@ -107,6 +108,84 @@ async def test_a_dropped_tunnel_is_reopened_before_the_next_thing_that_needs_it(
...
+
+            # the refresh builds the tree on the connection it recovered, so
+            # two `get_catalog()` calls would be two threads inside one adapter
+            schema_updates.clear()
+            app.action_refresh_catalog()
+            await wait_for_workers(app)
+            await pilot.pause()
+
+            assert len(schema_updates) == 1
+            assert app.data_catalog.database_tree.catalog is not None
```
tconbeer: This assertion is free: the tree already had a non-`None` catalog before the refresh, so it holds whether or not the one refresh did anything. Capture the catalog after the initial `wait_for_workers` and assert `is not first_catalog`, so the test pins "exactly one refresh, and it landed" rather than just the count.

I ran that change locally on this branch — it passes (3/3).

---

tests/functional_tests/test_ssh_recovery.py
```diff
@@ -107,6 +108,84 @@ async def test_a_dropped_tunnel_is_reopened_before_the_next_thing_that_needs_it(
...
+    def counted(app: Harlequin) -> None:
+        calls.append(None)
+        start_worker(app)
+
+    monkeypatch.setattr(Harlequin, "update_schema_data", counted)
+    return calls
+
+
+@pytest.mark.asyncio
+async def test_a_reconnect_rebuilds_the_catalog_on_the_new_connection(
```
tconbeer: The biggest gap in this PR: neither test exercises the symptom in #1127. Both assert on a refresh count and on catalog object identity — proxies for "the fix ran", not for "expanding an unloaded node works again". If a future change rebuilt the tree from something other than the new connection, these two would still pass.

The other implementation ([#1128](https://github.com/tconbeer/harlequin/pull/1128)) gets this right and is worth lifting: drive the recovery with a query, then

```python
await expand_catalog_node(pilot, db_node)
assert db_node.data.loaded
assert len(app.screen_stack) == 1  # no Catalog Error modal
```

Two details from it that matter. It uses `app_small_sqlite` rather than `app`, because `HarlequinSqliteConnection.close()` really closes the driver connection — so a stale item's `fetch_children()` fails deterministically instead of by luck. And it monkeypatches `DatabaseTree._schedule_prefetch_scan` to a no-op, so the node under test is still unloaded when the drop happens. (That second one is the brittle part of their test — reaching into a private method — but I don't see a cleaner way to guarantee an unloaded node, and it's worth the coupling to test the actual bug.)

I confirmed their assertion fails on `f19cdb0` and passes with the fix, so it's a real regression test for the symptom.

---

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Bug Fixes
+
+- Harlequin now refreshes the Data Catalog after an SSH tunnel drops and is reopened, so expanding a node no longer raises a Catalog Error until you press `ctrl+r` ([#1127](https://github.com/tconbeer/harlequin/issues/1127)).
```
tconbeer: The "until you press `ctrl+r`" tail reads backwards — as written it sounds like the error still happens up until you refresh. It's describing the old behavior inside a sentence about the new one. Suggest cutting it:

> - Harlequin now refreshes the Data Catalog when an SSH tunnel reconnects, so expanding a node no longer raises a Catalog Error ([#1127](https://github.com/tconbeer/harlequin/issues/1127)).

---

tconbeer (commented):
Reviewed this against [#1128](https://github.com/tconbeer/harlequin/pull/1128), the other implementation of #1127. **This is the one to take**, on the strength of one concrete difference, with the caveat that #1128's test is better than either of the two here.

## Why this one

Both PRs reach the same fix — rebuild the tree from the new connection's `get_catalog()` — and both are green across the matrix. The difference is the refresh-triggered recovery, which is a real path: `ctrl+r` is one of the two things that can wake a dropped tunnel, and it *is* the catalog rebuild.

#1128 calls `update_schema_data()` unconditionally from the handler, so on that path the worker that just recovered is cancelled mid-`get_catalog()` and a second one starts on the same connection. Textual can't preempt a thread worker, so the cancelled one runs to completion: two `get_catalog()` calls concurrently inside one adapter, two `NewCatalog` messages, two full tree reloads, and every expanded node re-fetched twice over a link that just proved itself slow. `get_catalog()` is cheap for the two bundled adapters, but the adapters that need tunnels most are out-of-tree.

That isn't theoretical. I ran this PR's `test_a_refresh_that_reconnects_does_not_ask_for_a_second_one` against #1128's `app.py`:

```
E       assert 2 == 1
E        +  where 2 = len([None, None])
```

Reproducible, 3/3. The flag is what buys that, and it's the smallest thing that could: one bool through one method with two callers.

Secondary: setting `database_tree.loading` matches what `action_refresh_catalog` already does, so the rebuild is visible instead of the user staring at a tree that is quietly wrong. #1128 doesn't.

I also checked the rest of the local suite behaves — `test_ssh_recovery.py` is 4/4 here, 3/3 on the base commit plus a failing new test, so the first new test is a genuine regression test for the mechanism.

## What I'd change before merging

Details are inline; the one that matters is the last:

1. `stale_catalog` names a fact that isn't true in its `False` case — it's really "this handler has to rebuild it".
2. The flag doesn't cover the case `_connection_for_worker`'s own comment warns about: a query worker and a schema worker hitting the same drop. Narrow, and not a blocker, but it's the case the rejected alternative *does* cover, so it's worth deciding out loud rather than by omission.
3. `loading = True` can stick if the follow-up refresh finds the tunnel gone again. Pre-existing in `action_refresh_catalog`; one line in `update_schema_data` fixes both.
4. The refresh test's `is not None` assertion is free — make it `is not first_catalog`.
5. **Take #1128's test.** Neither test here expands a node. They assert a refresh count and a catalog identity, which is the mechanism, not the bug — the user-visible failure in #1127 is a Catalog Error modal on expanding a node that hadn't loaded, and nothing here would notice if that came back. #1128 tests exactly that, and its choice of `app_small_sqlite` (because sqlite's `close()` really closes, so a stale item's fetch fails deterministically) is the reason it works. Credit where it's due.
6. Changelog wording.

## On the design doc

You asked whether to drop the §8 row. Leave it. The doc opens with "Nothing here is implemented yet" and §8's deferred table is written as "what brings each back" — it's a record of what the design decided at the time, not a status board, and editing one row makes it ambiguous whether the other rows are current. The row's own trigger ("bring it back when catalog items can be handed a new connection") is also still unmet: this rebuilds the tree instead of re-pointing items, which is the workaround the row describes, not the fix it asks for.

---

src/harlequin/app.py
tconbeer: Deciding out loud, and keeping the flag.

You're right that passing the intent down can't reach the loser — the winner holds the lock for the whole restart, so by the time the loser wakes there's nothing left to influence. I named that in the docstring rather than papering over it.

I didn't switch to the group check because it isn't a strict improvement — it trades this narrow double-call for a case where a needed rebuild is skipped. A schema worker that entered `get_catalog()` on the *old* connection, before the drop was detected, is still `RUNNING` when the query worker recovers. The handler would see the group busy and stand down; that worker then raises on the dead socket, so you get a Catalog Error modal and a tree that is still built on the dead connection — the #1127 symptom, behind the modal that hid it. The flag rebuilds correctly there.

So the two cover different halves, and the flag's failure mode (a redundant `get_catalog()`) is strictly milder than the group check's (a stale tree). Covering both means knowing which connection each in-flight rebuild is running against — worker-thread-mutated state, for a case that needs a query and a refresh to land on the same drop. Not worth it here; happy to revisit if the interleaving ever shows up in practice.

---

tests/functional_tests/test_ssh_recovery.py
tconbeer: Agreed, and taken. `test_a_reconnect_leaves_a_catalog_that_still_loads` is #1128's test — `app_small_sqlite`, the `_schedule_prefetch_scan` no-op, and the expand-and-no-modal assertions — with its polling loops folded into a `_first_database_node` helper (`wait_for_workers` returns before the message that reloads the tree has been handled, so both the initial and the rebuilt tree need the same wait). I kept the `schema_updates` count assertion in it, so the one test covers the symptom and the single-rebuild mechanism on the query path.

Checked it fails for the right reason: with the rebuild disabled and the count assertion removed, it stops at `db_node.data.connection is app.connection` — the tree's item still holding the connection that died — rather than passing on the count alone.

---

src/harlequin/app.py
tconbeer: Fixed in `update_schema_data`, so both callers get it. One wrinkle worth flagging: that worker runs on a thread, so it can't touch the widget directly — it goes through `call_from_thread`, and sets the `loading` reactive rather than calling `set_loading()`, which would uncover the tree while leaving the reactive `True` and make the next `loading = True` a silent no-op.

`test_a_tunnel_that_will_not_come_back_is_not_tried_twice` already drives that path (its second `ctrl+r` finds `tunnel.dropped`), so it now asserts the spinner stopped; the assertion fails without the fix.

---

tconbeer: Correction to my reply above: the `call_from_thread` is gone as of 95500af. It was the wrong tool — it works, but it's a worker reaching for a widget, which is the thing `app.py` otherwise never does.

The worker now posts `CatalogRefreshAborted` on the no-connection path, and a handler clears `loading` on the main thread, like every other state change here. Same assertion still pins it.

I kept the `loading = True` at the call sites rather than moving it into the worker too. Posting it from the worker *after* it has a connection would mean no spinner for the seconds `_connection_for_worker()` spends restarting ssh — exactly the wait worth showing — and posting it before would need a second message to undo on the abort path anyway.

---

PR #1132, opened 2026-09-03
Warm sessions 1/6: the entry-point split and the frames

tconbeer:
PR 1 of [`docs/plans/hsql-warm-sessions-design.md`](https://github.com/tconbeer/harlequin/blob/claude/hsql-warm-sessions-design-7ebvx7/docs/plans/hsql-warm-sessions-design.md), which this PR also revises for what has shipped since it was written. No open issue.

Two commits: the plan revision, then the code.

*The plan.* It was written against 2.9.0, when M2 was still a plan. M1–M3 have shipped and SSH tunnels have landed, so the flag partition changed and every number in it was measured on a different machine. Re-measured on this checkout; a new §10 lists the changes for a reader holding the first draft in their head.

*The code.* `main()` decides whether an invocation belongs to a warm session before it imports anything that is not stdlib, and hands off to a client that forwards argv over an `AF_UNIX` socket and copies back stdout bytes, stderr bytes and an exit code.

- `session.py` — which session an invocation names, and where its socket is. Stdlib only, and read before anything else.
- `client.py` — stdlib only: `socket`, `struct`, `sys`, `os`. Scans argv for the two things it cannot avoid knowing about (its own `--session`, and a `-f -` whose bytes only it can read) and forwards the rest opaquely.
- `protocol.py` — chunked frames, the handshake, the version literal.

There is no server yet, so the only reachable behaviors are the two the design specifies for a session that is not running: `--session` fails with exit 3, `HSQL_SESSION` warns on stderr and runs cold. The rest is pinned against a stub server built on `protocol.py`, so the frames these tests assert are the ones the real server will send.

The design's thesis is that this feature is decided by the first few lines of the console script, not by the protocol — a warm round trip is about a millisecond, so what a caller waits for is the client starting. Measuring it turned up **three** expensive imports on the entry point's path, not the one the plan named:

| | Cost | Why it was there |
| --- | --- | --- |
| `click` | 36ms | `main()` opened with it |
| `harlequin.hsql.diagnostics` | 68ms | module-scope `ExitCode` import, reaching `redact` → `config` → msgspec |
| `typing` | 10ms | `harlequin/__init__.py` imports it, and every `harlequin.*` import executes that file |

With all three deferred the client's whole graph is **25 modules and ~26ms over a bare interpreter, against 348ms cold** — 20 sequential invocations in 0.55s against 7.5s. The cold path is unchanged (342ms on this branch against 346ms on `main`, best of twenty): it imported all three anyway.

Two consequences worth flagging for review:

- **The client writes its own stderr** rather than going through `diagnostics._write()`, which is otherwise "the one line every message leaves through". That function exists to redact, and the client holds no secret to redact — it reads no profile, loads no adapter, never sees a connection string. The server's stderr arrives already redacted and is copied through untouched.
- **`harlequin/__init__.py` declares `TYPE_CHECKING = False` itself** instead of importing it. Every annotation there is a string under PEP 563, so nothing needs the name at run time, and mypy handles the literal. Slightly unusual, and it buys 10ms on a path where 10ms is a third of the budget.

Two decisions the design did not cover, both now recorded in its §9:

- **`--session` is deliberately not yet a click option.** A flag that `--help` and `--spec` advertise while no server can answer it is worse than one they do not mention. Declaring it also raises a question PR 2 has to answer anyway: a click parameter's name becomes a *profile* key, and `session` must not be one, because the decision is made before any config file is read. That is a `CLI_ONLY_SSH_KEYS`-shaped refusal and it belongs with the flag. Until then no adapter reserves the spelling either — none of duckdb, sqlite, postgres or mysql declares it.
- **The version literal lives in `protocol.py`.** The handshake compares releases, and `importlib.metadata.version()` costs ~43ms on a 27ms invocation, so it is a literal — the second version `uv version` cannot reach. `scripts/bump_plugin_version.py` therefore becomes `scripts/bump_versions.py` and writes both; a test pins each to the installed version, so a release that forgot fails in the release PR rather than in a user's shell.

Alternatives considered and not taken: platformdirs for the socket directory (~22ms, the client's whole budget — the derivation is a dozen lines of `os.path` shared by both halves instead); `json` for the request encoding (9ms, a third of the invocation — lengths and bytes instead); and a second import-linter contract for "`harlequin.hsql.__init__` may not import click at module scope", which is not expressible there because a contract's source module matches its descendants, so that half is a run-time test.

- [x] No. Nothing here is reachable as a feature yet; the docs land with PR 6, per the design's §9.

- [x] Yes. `tests/unit_tests/test_hsql_session.py` (66 tests: the argv scan, the frames, and the client against a stub server), plus four run-time guards in `test_import_hygiene.py` and a new import-linter contract. `make check` passes.

- [x] I have added an entry to `CHANGELOG.md`, under the `[Unreleased]` section heading. That entry references the issue closed by this PR. — **Not done, on purpose.** Nothing here is usable until the server lands, and `--session` is not a documented flag yet, so there is no notable user-facing change to announce. The entry belongs with PR 2.

https://claude.ai/code/session_019NghGdWpK6rSiuxGhhTP7r

---
_Generated by [Claude Code](https://claude.ai/code/session_019NghGdWpK6rSiuxGhhTP7r)_


src/harlequin/hsql/client.py
```diff
@@ -0,0 +1,244 @@
...
+            session,
+            f"{session.name!r} is not a session name",
+            remedy=(
+                " A name is letters, digits, underscores and dashes, up to "
+                f"{MAX_NAME_LENGTH} of them."
+            ),
+            code=USAGE,
+        )
+
+    connection = _connect(socket_path(session.name, environ))
```
tconbeer: **Bug: an `OSError` from `connect()` other than ENOENT/ECONNREFUSED is an uncaught traceback.** `_connect` re-raises everything else, and this call sits outside the `try` below, so the caller gets a stack trace and exit 1 instead of a diagnostic and exit 3. Two reproductions on this branch:

```
$ TMPDIR=/var/folders/2k/1v3p2b_92xz7f2k9h3n0mq1h0000gn/T XDG_RUNTIME_DIR= hsql --session $(printf 'x%.0s' {1..64}) -c 'select 1'
Traceback (most recent call last):
  ...
  File ".../harlequin/hsql/client.py", line 119, in _connect
    sock.connect(path)
OSError: AF_UNIX path too long

$ XDG_RUNTIME_DIR=/etc/passwd hsql --session nobody -c 'select 1'
NotADirectoryError: [Errno 20] Not a directory
```

EACCES on a directory someone else owns is the same path. Suggest moving the `_connect` call inside the `try`, or having `_connect` map any other `OSError` to `CONNECTION` with the errno text in the message.

---

src/harlequin/hsql/session.py
```diff
@@ -0,0 +1,138 @@
...
+    under `TMPDIR` where it does not -- which is macOS, WSL2 without systemd,
+    and a container, so the fallback is the common path rather than the exotic
+    one. platformdirs answers the same question and costs ~22ms to import, or
+    about the whole of what a warm invocation is supposed to cost, so the
+    derivation is stdlib and both halves of the session share this function.
+    """
+    xdg_runtime_dir = environ.get("XDG_RUNTIME_DIR")
+    if xdg_runtime_dir:
+        return os.path.join(xdg_runtime_dir, "hsql")
+    return os.path.join(environ.get("TMPDIR") or "/tmp", f"hsql-{os.getuid()}")
```
tconbeer, line 144: **Security: the client trusts whatever listens at this path, and what it sends is argv, cwd and stdin.** `argv` can carry `CONN_STR` with a password in it. Under `$XDG_RUNTIME_DIR` that is fine, since the system guarantees the directory is 0700 and the user's. Under the `/tmp/hsql-<uid>` fallback, which the design says is the common path, nothing guarantees it: on a shared host another user can create `/tmp/hsql-1000/` first, put a listener at `prod.sock`, echo the version in `HELLO`, and receive every warm invocation's arguments and piped SQL. The same directory is where `_unlink_stale` deletes files.

The fix is one `os.stat` before connecting: refuse unless the directory is owned by `os.getuid()` and has no group/other bits. It is stdlib and one syscall, so it fits the budget. PR 2's server needs the same check, since `os.mkdir(mode=0o700)` does nothing to a directory that already exists.

---

src/harlequin/hsql/session.py
```diff
@@ -0,0 +1,138 @@
...
+if TYPE_CHECKING:
+    from typing import Mapping, Sequence
+
+SESSION_ENV_VAR = "HSQL_SESSION"
+"""The ambient spelling: a preference, so an invocation still runs without one."""
+
+SESSION_OPTION = "--session"
+"""The typed spelling: an assertion, so an invocation fails without one."""
+
+MAX_NAME_LENGTH = 64
```
tconbeer, line 30: **A 64-character name cannot be served on macOS.** `sun_path` is 104 bytes there, and the fallback directory is under a `TMPDIR` like `/var/folders/2k/1v3p2b_92xz7f2k9h3n0mq1h0000gn/T/` (49 chars). With `hsql-501/` and `.sock` that leaves room for a name of about 40 characters; anything longer fails at `connect()` with `AF_UNIX path too long`, which today is the traceback in the comment on `client.py:94`. The test fixture already works around this with a short-pathed runtime dir, so the suite will not catch it.

Either lower the limit to something that fits the worst-case fallback path, or have `socket_path()` (or `run()`) check the assembled path against the platform limit and refuse with a message that names the cause.

---

src/harlequin/hsql/client.py
```diff
@@ -0,0 +1,244 @@
...
+    """
+    sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    try:
+        sock.connect(path)
+    except FileNotFoundError:
+        sock.close()
+        return None
+    except ConnectionRefusedError:
+        sock.close()
+        _unlink_stale(path)
```
tconbeer: **Unlinking on the first ECONNREFUSED can delete a live server's socket.** Two ways a running server refuses a connection:

- The window between its `bind()` and `listen()`. Narrow, but a client that lands in it unlinks the file the server just bound, and the server never notices.
- On macOS and the BSDs, a **full listen backlog** is reported as ECONNREFUSED (Linux blocks instead). That is exactly the load pattern this feature is for, an agent firing invocations in parallel, and one refused client takes the session away from every caller after it.

The design's §4.4 says the server owns the socket file, and I think that should hold for cleanup too: the server unlinks a stale file on startup (after failing to connect to it itself), and the client only reports. If the client has to clean up, it should at least not do so on a single refusal.

---

src/harlequin/hsql/client.py
```diff
@@ -0,0 +1,244 @@
...
+        if argument == "--":
+            break
+        # `-f-`, which click reads as a short option and its value
+        if argument == f"-f{STDIN_ARGUMENT}":
+            return True
+        if argument in FILE_OPTIONS and argv[index + 1 : index + 2] == [STDIN_ARGUMENT]:
+            return True
+        if any(argument == f"{option}={STDIN_ARGUMENT}" for option in FILE_OPTIONS):
+            return True
+    return False
```
tconbeer, line 282: **Bug: `-f` bundled with other short flags is missed.** click reads `-tAf -`, `-tf -`, `-Af -` and `-tf-` as `-t`/`-A` plus `-f -`, and the psql-style `-tAf -` is the idiom the skill teaches. Verified on this branch:

```
$ echo 'select 42' | hsql -tAf - --no-init :memory:
42
$ python -c "from harlequin.hsql import client; print(client._reads_stdin(['-tAf', '-']))"
False
```

So a warm `-tAf -` would hand the server `-f -` with no stdin. The inverse is also here: `-f=-` matches the `{option}=-` branch, but click gives short options no `=` form and reads a file named `=-` cold, so the client would drain stdin for an invocation that never uses it. Suggest matching a short cluster that ends in `f` (with the next argument, or a trailing `-`), keeping the `=` form for `--file` only, and adding the bundled cases to the parametrize in `test_which_invocations_need_the_callers_stdin`.

---

src/harlequin/hsql/protocol.py
```diff
@@ -0,0 +1,230 @@
...
+
+STDERR = 4
+"""Server to client, repeatable: bytes for the client's stderr."""
+
+EXIT = 5
+"""Server to client, last: the code the client exits with."""
+
+# Frame kinds are wire values: later ones append rather than renumbering.
+
+FORWARDED_ENV_VARS = ("TERM", "NO_COLOR")
```
tconbeer: **The request carries a variable nothing reads, and not the fact `--color auto` needs.** `_use_color()` in `cli.py` decides `auto` from `sys.stdout.isatty()` and `NO_COLOR`; it never reads `TERM`. On the server, stdout is a buffer, so `auto` would always be "no color" however the caller's terminal looks, and `TERM` travels for nothing.

Since the whole point of framing now is what "cannot be retrofitted without a version bump", the request should carry whether the client's stdout (and stderr) is a TTY, as a fifth section or a flags byte. Whether `TERM` stays is up to what PR 2 does with it.

---

src/harlequin/hsql/client.py
```diff
@@ -0,0 +1,244 @@
...
+    connection = _connect(socket_path(session.name, environ))
+    if connection is None:
+        return _no_session(
+            session,
+            f"no session named {session.name!r} is running",
+            remedy=f" Start one with `hsql --serve {session.name} ...`.",
+        )
+    try:
+        return _exchange(connection, session, argv, environ)
+    except (protocol.ProtocolError, OSError) as e:
```
tconbeer, line 111: **Ctrl-C on the warm path is a traceback.** `_run_cold` maps `KeyboardInterrupt` to `ExitCode.INTERRUPT`; nothing on the warm branch of `main()` does, so interrupting a client that is waiting in `_relay` prints a stack trace and exits 1. Copy `INTERRUPT` next to `USAGE`/`CONNECTION`, pin it in the same test, and catch it in `main()` around `run()`.

---

tests/unit_tests/test_import_hygiene.py
```diff
@@ -64,6 +66,72 @@ def test_headless_imports_do_not_load_the_tui(
...
+        "import harlequin.hsql\n"
+        "import sys\n"
+        "print(','.join(m for m in ('click', 'harlequin.hsql.cli', "
+        "'harlequin.hsql.diagnostics') if m in sys.modules))\n"
+    )
+    leaked = [m for m in proc.stdout.strip().split(",") if m]
+    assert not leaked, f"importing the entry point loaded {leaked}"
+
+
+def test_a_warm_invocation_never_reaches_the_cold_path(
```
tconbeer, line 111: **Test fragility: `run_python` (conftest) spreads `os.environ` without dropping `HSQL_SESSION`.** Every `main()`-driving test in this file now goes through the client first, and an ambient session puts `note: no session named ... running cold` on stderr line 1, which the `proc.stderr.split("\n")[:2]` parsing reads as the adapter list. On this branch:

```
$ HSQL_SESSION=devbox uv run pytest tests/unit_tests/test_import_hygiene.py -n0 -q
11 failed, 37 passed
```

`hsql_process` in `test_hsql_session.py` already strips the variable; `run_python` should too. Same for `XDG_RUNTIME_DIR`/`TMPDIR` pointing somewhere odd, which today turns this test into the traceback noted on `client.py:94` rather than an assertion that names the cause.

---

src/harlequin/hsql/session.py
```diff
@@ -0,0 +1,138 @@
...
+    """The session this invocation names, or None if it names none.
+
+    A scan rather than a parse: the client cannot afford click, and the server
+    is the thing that parses flags. The one ambiguity is an option *value* that
+    is the literal string `--session`, which this would read as the flag -- the
+    same ambiguity every argv pre-scan has, and one that fails visibly (a
+    session nobody started) rather than quietly.
+
+    `--session` with no value is left alone, so that click reports it with the
+    message it gives every other option missing an argument.
```
tconbeer: This is not what happens today. `--session` is not declared on the click command, so a bare `hsql --session` gets `Error: No such option '--session'. (Did you mean one of: '--extension', '--json', '--version'?)`, which is a worse message than the client's own would be. `test_a_session_flag_with_no_value_is_clicks_to_report` asserts only exit 2, which the client's `'' is not a session name` refusal also produces, so it pins neither who reports it nor what is said. Either have the client refuse a bare `--session` itself until PR 2 declares the option, or fix the docstring, and assert stderr in the test either way.

---

src/harlequin/hsql/protocol.py
```diff
@@ -0,0 +1,230 @@
+"""The frames a session's client and server exchange, and how they are framed.
+
+What crosses the wire is **stdout bytes, stderr bytes and an exit code**, not
+rows: the server runs the same execution core and the same writer the cold path
+does, into a buffer, and the client copies that buffer to its own streams. So
+the client cannot format anything, needs no pyarrow and no `--format`
+knowledge, and cannot disagree with a cold invocation about a timestamp.
+
+Output is **chunked** from the first version -- many `STDOUT` frames, then one
+`EXIT` -- because that is free now and impossible to retrofit without a
```
tconbeer: **Comments, against AGENTS.md's conventions.** The prose across the new modules is mostly what/why, but a set of it records decisions, alternatives, or numbers that will go stale. The ones I would cut, in order:

- Millisecond figures in about ten places (`~68ms`, `~22ms`, `~43ms`, `~10ms`, `about 20ms`), which already disagree: `client.py` says `diagnostics` costs ~68ms and `test_hsql_session.py` says ~63ms. The qualitative claim ("more than the round trip") is the durable one; `scripts/cold_start.py` is where the number lives.
- This sentence ("free now and impossible to retrofit without a protocol version bump") and `VERSION`'s docstring ("rather than a protocol number somebody has to remember to bump") are alternatives weighed. "Output is chunked: many `STDOUT` frames, then one `EXIT`" is the whole of what a reader needs.
- `client.py:71` "through 3.14" is a version bound that goes stale on the next CPython.
- "There is no server yet" / "until the server lands" in the test module and test docstrings describe the PR's staging, and will be false and unfixed after PR 2.
- `session.py`: "Spelled as a set rather than as a pattern because `re`...", and the platformdirs paragraph under `runtime_dir`, are "rather than X" notes; the module docstring and the import-linter contract already carry the rule.
- `bump_versions.py`'s module docstring restates `protocol.py`'s rationale with another number.

---

pyproject.toml
```diff
@@ -374,6 +374,61 @@ ignore_imports = [
...
+    "harlequin.hsql.timeout",
+    "harlequin.layout",
+    "harlequin.navigate",
+    "harlequin.options",
+    "harlequin.plugins",
+    "harlequin.query",
+    "harlequin.redact",
+    "harlequin.ssh",
+    "harlequin.statements",
+    "harlequin.transaction_mode",
```
tconbeer: Nit: this list has to be maintained by hand, and a future stdlib-only `harlequin.*` module would not be on it. `forbidden_modules = [..., "harlequin"]` with two `ignore_imports` entries (`harlequin.hsql.client -> harlequin.hsql.protocol` and `-> harlequin.hsql.session`) passes `lint-imports` on this branch and cannot go stale. I checked that both shapes catch a stray `import harlequin.keymap` in `protocol.py`, so this is only about upkeep.

---

tconbeer (commented):
## Review of #1132

**Scope.** Correctness of the argv scan and the client/protocol, security of the socket path, the test assertions, and the comments against AGENTS.md. Checked out at `1832855`; `ruff`, `mypy`, `lint-imports` and the two new test modules all pass here (117 tests). The design doc was skimmed for consistency with the code only.

**Verdict: not ready to merge as is.** The shape is right and the import discipline is real (25 modules over a bare interpreter, measured here too), but there are four things I would fix before this lands, since each is client behavior PR 2 will inherit:

1. **Uncaught `OSError` from `connect()`** is a traceback and exit 1 (over-long path, `ENOTDIR`, `EACCES`). Reproduced. `client.py:94`.
2. **The client sends argv, cwd and stdin to whatever listens** under `/tmp/hsql-<uid>` without checking the directory is the user's and 0700. `session.py:128`.
3. **`-tAf -` and other bundled short flags hide `-f -`** from the client, so the psql idiom the skill teaches would reach the server with no stdin. Reproduced. `client.py:201`.
4. **`run_python` inherits `HSQL_SESSION`**, so a developer with one set gets 11 failures in the import-hygiene suite. Reproduced. `test_import_hygiene.py:111`.

Also worth deciding now rather than in PR 2, because they are protocol or ownership questions: the stale-socket unlink on a single `ECONNREFUSED` (`client.py:125`), `--color auto` needing the client's `isatty()` while `TERM` travels unread (`protocol.py:58`), and `MAX_NAME_LENGTH = 64` being unreachable on macOS (`session.py:29`). Smaller: Ctrl-C on the warm path is a traceback, the bare `--session` docstring does not match what click prints today, a comment sweep for numbers and decision records, and a simpler contract shape.

**Minor, not commented inline:** a `-f -` larger than `MAX_PAYLOAD` (64 MiB) fails as "the session did not answer: a N-byte frame is too large to send", where the cold path would run it; `requested_session` takes the first of two `--session` values while click's convention is last-wins; several stub-server tests never assert `server.request is not None` or the returned exit code, so a malformed `REQUEST` frame would pass `test_a_response_reaches_the_caller_byte_for_byte`.

## Not checked
- I did not review the design doc's measurements or its §3–§8 content beyond checking that what it says PR 1 ships matches the code.
- I did not run the suite on macOS or Windows; the `sun_path` and `ECONNREFUSED`-on-full-backlog claims are from the platform semantics, not from a reproduction on those platforms.
- I did not verify the `~26ms` wall-clock figures; I relied on the module-count test.
- I did not exercise a hostile peer against `unpack_strings` beyond the truncation tests; the count loop is bounded by `MAX_PAYLOAD / 4` iterations and the directory is meant to be 0700, so I left it.
- I did not check `release.yml` end to end, only that `bump_versions.py` is what it calls and that the rewrite tests pin both literals.

---

src/harlequin/hsql/client.py
tconbeer: Fixed in 7639299. Both reproduced first. The connect now happens inside the `try`, so any other errno becomes `a session named 'x' is unreachable: <errno text>` and exit 3; a runtime directory that is simply absent is caught separately and means "no session", not an error.

```
$ XDG_RUNTIME_DIR=/etc/passwd hsql --session nobody -c 'select 1'
hsql: error: a session named 'nobody' is unreachable: [Errno 20] Not a directory: '/etc/passwd/hsql'.
$ echo $?
3
```

`test_a_runtime_dir_that_cannot_hold_a_socket_is_a_diagnostic` covers it. The over-long path is now refused earlier, by the check in the `session.py:29` thread.

---

src/harlequin/hsql/session.py
tconbeer: Agreed, and fixed in 7639299. `session.check_runtime_dir()` stats the directory and refuses unless it is a directory owned by `os.getuid()` with no group or other bits. It lives in `session.py` rather than `client.py` for the reason you gave — PR 2's server needs the same check, since its own `mkdir(mode=0o700)` is a no-op on a directory that already exists — and its docstring says so.

Reproduced the attack and confirmed the fix closes it: a listener in a 0777 `hsql/` directory, and a client invoked with `postgres://u:hunter2@h/db` in argv.

```
hsql: error: a session named 'prod' is unreachable: /tmp/hijack/hsql is mode 777, so other users can reach what it holds.
bytes the listener received: none
```

`test_a_directory_someone_else_could_reach_is_never_connected_to` is that reproduction, plus unit tests over 0777/0750/0701 and a non-directory. `_unlink_stale` is gone entirely (see the `client.py:125` thread), so there is no delete in that directory either.

---

src/harlequin/hsql/session.py
tconbeer: Took the second option, in 7639299: `MAX_SOCKET_PATH = 104` (the shorter of the two platforms, so a name works the same on both), and the client measures the assembled path before connecting.

```
hsql: error: the socket path for 'xxxx…' is 162 bytes, over the 104 a unix socket takes. Use a shorter name, or set XDG_RUNTIME_DIR to a shorter directory.
```

I kept `MAX_NAME_LENGTH` at 64 rather than lowering it, because the limit that actually binds depends on the directory: 64 is fine under a short `XDG_RUNTIME_DIR` and too long under macOS's `TMPDIR`, and a fixed lower number would refuse names that work. The path is the real constraint, so that is what is measured. `test_a_socket_path_too_long_to_bind_is_refused_by_name` exercises it with a long runtime dir, so it does not depend on the platform running the suite.

---

src/harlequin/hsql/client.py
tconbeer: Agreed — the backlog case makes this a real hazard under exactly the load the feature is for. `_unlink_stale` is gone in 7639299; the client reports a refusal and leaves the file alone, and `_connect`'s docstring records both refusal cases so nobody re-adds it.

The design's §4.9 said "client unlinks it", so I updated that row too: the server unlinks a stale socket when it starts, as §4.4 already implied about ownership.

`test_a_refused_socket_is_reported_and_left_alone` replaces the old test, and asserts the file is still there afterwards.

---

src/harlequin/hsql/client.py
tconbeer: Reproduced and fixed in 7639299. `_short_cluster_reads_stdin` now reads a cluster the way click does — left to right, stopping at the first option that takes a value, since that option consumes the rest of the cluster.

```
['-tAf', '-']   True     ['-f=-']      False
['-tf', '-']    True     ['-cf', '-']  False
['-Af', '-']    True     ['-f', '-']   True
['-tf-']        True
```

`-cf -` matters as much as the bundled cases: `f` there is `-c`'s value, and draining stdin for it would hang an invocation that never reads it — so the scan is conservative in that direction rather than liberal.

That means the client has to know which short options take no value, so `BOOLEAN_SHORT_FLAGS = "tArx"` is a copied set, pinned against `bare_command()` by `test_the_clients_boolean_short_flags_are_the_commands` — the same shape as `IDE_THEMES`. All your cases plus `-xrtAf -`, `-of -` and `-- -f -` are in the parametrize.

---

src/harlequin/hsql/protocol.py
tconbeer: Right on both halves, and this is the one I most wanted caught before the frames are fixed. In 7639299 the request grew a fifth section — a flags byte, `STDOUT_ISATTY | STDERR_ISATTY` — and `TERM` came off `FORWARDED_ENV_VARS`, leaving `NO_COLOR`, which `_use_color()` does read.

I dropped `TERM` rather than leaving it for PR 2 to decide: carrying a value nothing reads is the kind of thing that survives by inertia, and adding it back is a section that already exists.

`test_a_request_carries_which_streams_are_terminals` covers all four combinations, and the design's §4.2 now says what the request carries and why.

---

src/harlequin/hsql/client.py
tconbeer: Fixed in 7639299, as suggested: `INTERRUPT = 130` beside `USAGE`/`CONNECTION`, pinned in `test_the_clients_exit_codes_are_hsqls`, caught both in `run()` (around the exchange) and in `main()` around `run()` itself.

It exits silently, which is what the cold path does — `_run_cold` maps `KeyboardInterrupt` to `INTERRUPT` and prints nothing. When PR 4 adds the cancel frame this is where the "still running on the server" line goes, for an adapter that cannot cancel.

`test_an_interrupt_while_waiting_exits_the_way_the_cold_path_does` raises from `_relay` to cover it.

---

tests/unit_tests/test_import_hygiene.py
tconbeer: Reproduced (11 failed, 37 passed) and fixed in 7639299 — `run_python` now drops `HSQL_SESSION` the way `hsql_process` does, and its docstring says why, since the reason is new: every `main()` in a child now goes through the client first.

```
$ HSQL_SESSION=devbox uv run pytest tests/unit_tests/test_import_hygiene.py -n0 -q
48 passed
```

On `XDG_RUNTIME_DIR`/`TMPDIR`: I left those inherited. They no longer produce a traceback — a bad one is now the diagnostic from the `client.py:94` thread — and the tests that would notice are the ones that name a session, which all supply their own runtime dir. Stripping them would also make the child's environment differ from the parent's in a way that hides a real platform problem rather than surfacing it. Happy to strip them too if you would rather the child be hermetic on that axis.

---

src/harlequin/hsql/session.py
tconbeer: Took the first option — the client refuses it — in 7639299. `requested_session` now returns an empty name for a bare `--session`, and the client answers by the flag's name:

```
$ hsql --session
hsql: error: --session needs a name. Pass one: `hsql --session prod ...`.
$ echo $?
2
```

An empty name can only arrive that way: an empty `HSQL_SESSION` names no session at all, so there is no ambiguity with the `'' is not a session name` branch, which now cannot be reached.

The docstring is rewritten to describe that, and the test is renamed to `test_a_session_flag_with_no_value_names_itself` and asserts the message and empty stdout rather than only the code. When PR 2 declares the option, this becomes the question of whether click's message or the client's is better — the client's runs first either way.

---

src/harlequin/hsql/protocol.py
tconbeer: All cut in 7639299, in the order you listed them.

The two disagreeing figures were the tell — I had measured `diagnostics` twice on different days and written both down. Every `~Nms` in the new modules is gone, replaced by the qualitative claim ("more than the round trip", "a fraction of what `typing` costs"), with `scripts/cold_start.py` as the place the number lives. That includes `harlequin/__init__.py`'s `TYPE_CHECKING` docstring, which had the `~10ms`.

Also gone: the retrofit sentence and `VERSION`'s "rather than a protocol number" clause; "through 3.14"; "There is no server yet" and "until the server lands" in the test module and two docstrings; `session.py`'s "spelled as a set rather than as a pattern" and the platformdirs paragraph; and `bump_versions.py`'s restatement.

I kept the module-count figures in `test_import_hygiene.py`, since those are what the test asserts rather than prose about them.

---

pyproject.toml
tconbeer: Taken in 7639299 — `forbidden_modules` is now the third-party names plus `harlequin`, with the two `ignore_imports` entries, and a comment saying a stdlib-only module added later is covered without anyone remembering.

Confirmed it still bites: a stray `import harlequin.keymap` in `protocol.py` breaks the contract, and removing it passes 5/5.

---

tconbeer:
Thanks — all four blocking findings reproduced exactly as written, and every one is fixed in 7639299. Replies are on each thread; ten are resolved, and I left the `run_python` one open because I declined half of it and would like your call.

**The three minor findings**, which had no thread:

- **Oversized `-f -`.** Now refused by name rather than as "the session did not answer": `the SQL on stdin is N bytes, over the 67108864 a session request carries. Run it without a session.` I did not raise `MAX_PAYLOAD` — it is the guard that stops a bad length from becoming an allocation on the read side, and both halves share it — so the honest answer is to say what the limit is and where the invocation still works.
- **`--session` last-wins.** Fixed; `requested_session` now keeps the last match, as click parses every other option. Both spellings are in the parametrize.
- **Stub-server tests.** `test_a_response_reaches_the_caller_byte_for_byte` now asserts `server.request is not None` and the argv it received, and the two other stub tests assert the returned exit code, so a malformed `REQUEST` frame fails them.

Two things changed that the design doc asserted, so it moved with the code: §4.9's stale-socket row (the server owns cleanup) and §4.2's list of what the request carries. §4.4 and §4.7 gained the path limit and the directory check.

`make check` on the merge: 1595 passed, 7 skipped; py12 3 passed; mypy, ruff and 5/5 import-linter contracts clean. The client's graph is still 25 modules over a bare interpreter.

---

tests/unit_tests/test_hsql_session.py
```diff
@@ -0,0 +1,769 @@
...
+    assert client.run(typed(), ["-c", "select 1"], environ) == ExitCode.CONNECTION
+    assert "is unreachable" in capsys.readouterr().err
+
+
+@needs_unix_sockets
+def test_a_socket_path_too_long_to_bind_is_refused_by_name(
+    capsys: pytest.CaptureFixture[str],
+) -> None:
+    """`sun_path` is 104 bytes on macOS, and the TMPDIR it supplies is ~50 of
+    them, so a name well inside MAX_NAME_LENGTH can still be unreachable."""
```
tconbeer, line 575: **These two new tests will fail on macOS.** Both build the socket path under pytest's `tmp_path`, which is what the file's own `runtime_dir` fixture exists to avoid: on macOS that base is `/private/var/folders/.../T/pytest-of-runner/pytest-0/<test name>0/`, already past 104 bytes before `hsql/prod.sock` is appended. Two consequences, both reproducible on Linux by giving pytest a long base temp:

```
$ uv run pytest --basetemp=/tmp/ddd...(80 chars) tests/unit_tests/test_hsql_session.py -n0 -k "cannot_hold_a_socket or someone_else_could_reach"
test_hsql_session.py:539: AssertionError: assert 2 == <ExitCode.CONNECTION: 3>
test_hsql_session.py:564: OSError: AF_UNIX path too long
2 failed
```

The first is the new path-length check firing (exit 2) before the directory check gets a chance; the second is the test's own `listener.bind()`. Use the `runtime_dir` fixture (or `tempfile.mkdtemp(dir="/tmp")`) for the directory in both, the way the stub server does. CI has not reported on this head yet, so the macOS jobs are what will show it.

---

src/harlequin/hsql/session.py
```diff
@@ -0,0 +1,179 @@
...
+    connection string with a password in it. Under `XDG_RUNTIME_DIR` the system
+    guarantees the directory; under the `TMPDIR` fallback nothing does, and one
+    another user created first would receive all of it. `os.mkdir(mode=0o700)`
+    does nothing to a directory that already exists, so the server checks this
+    too rather than trusting its own mkdir.
+
+    Raises OSError if the directory cannot be read, and UnsafeRuntimeDir if it
+    is not a directory this user owns privately.
+    """
+    info = os.stat(path)
```
tconbeer: **`os.stat` follows symlinks, so the check can be passed with one.** Under the `/tmp` fallback an attacker who gets there first can make `/tmp/hsql-1000` a symlink to a directory the user owns privately (`~/.ssh`, say). The stat then reports the user's uid and mode 700 and this accepts it, and since the attacker owns the link in a sticky `/tmp` they can repoint it between this check and `connect()`. Verified on this branch:

```
>>> os.symlink(private_0700_dir, link); session.check_runtime_dir(link)
accepted
```

`os.lstat` closes it: a symlink then fails the `S_ISDIR` test that is already here, and a real directory in a sticky `/tmp` cannot be renamed or replaced by another user. Neither `$XDG_RUNTIME_DIR/hsql` nor `/tmp/hsql-<uid>` has a legitimate reason to be a link, and the server should refuse the same way before it binds.

---

src/harlequin/hsql/client.py
```diff
@@ -0,0 +1,336 @@
...
+report on. `tests/unit_tests/test_hsql_session.py` pins all three to the enum.
+"""
+
+STDIN_ARGUMENT = "-"
+FILE_OPTION = "--file"
+FILE_SHORT = "f"
+"""The one per-request flag the client has to understand: the server has no
+stdin, so bytes a caller piped in have to travel with the request."""
+
+BOOLEAN_SHORT_FLAGS = "tArx"
```
tconbeer, line 64: The cluster scan and its pin look right for hsql's own flags, and reading conservatively on `-cf -` is the correct direction. One limit worth recording for PR 2 rather than fixing here: the pin is against `bare_command()`, so a boolean short flag an adapter declares is not in this string, and `-zf -` for such a `-z` would be read as "`-z` takes a value" and forward no stdin. The way to make every scan miss visible, this one included, is for the server to refuse a request whose argv names `-f -` but carries no stdin section, rather than reading its own stdin. That is one check in PR 2 and it turns a silent empty script into an exit 2.

---

src/harlequin/hsql/client.py
```diff
@@ -0,0 +1,336 @@
...
+            stdin=stdin,
+            stdout_isatty=_isatty(sys.stdout),
+            stderr_isatty=_isatty(sys.stderr),
+        ),
+    )
+    return _relay(connection)
+
+
+def _isatty(stream: "object") -> bool:
+    """Whether `stream` is a terminal, which only this side can answer.
```
tconbeer, line 232: Nit: annotating the parameter as `"TextIO"` (under the existing `TYPE_CHECKING` block) lets mypy see `isatty()` and drops the `type: ignore`. Both call sites pass `sys.stdout`/`sys.stderr`, and the `AttributeError` branch still covers a replaced stream at run time.

---

tests/unit_tests/test_hsql_session.py
```diff
@@ -0,0 +1,769 @@
...
+def test_a_refused_socket_is_reported_and_left_alone(
+    environ: dict[str, str], capsys: pytest.CaptureFixture[str]
+) -> None:
+    """A running server also refuses between its `bind()` and its `listen()`,
+    and on the BSDs whenever its backlog is full, so a client that deleted the
+    file on one refusal would take a live session away from everyone after it.
+    The server unlinks a stale socket when it starts."""
+    path = session.socket_path("prod", environ)
+    bound = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    bound.bind(path)
```
tconbeer, line 549: Nit, same convention as before: "used to escape `run()` as one, and exit 1" describes the code this commit removed, and `test_a_session_flag_with_no_value_names_itself` at line 746 says "until PR 2 declares it". Both read as live facts to the next person. "Any errno from `connect()` other than nothing-is-listening is a diagnostic and exit 3" and "the client refuses it; click does not declare the option" say the same without the history.

---

tconbeer (commented):
## Follow-up on 7639299

**Scope.** Re-read the whole fix commit against my previous findings, reran `ruff`, `mypy`, `lint-imports` (5/5) and the three affected test modules here (151 passed, including with `HSQL_SESSION=devbox`), and reproduced each behavioral claim in the replies that I could reproduce on Linux.

**Every prior finding is addressed as described**, and the responses match the code: the connect errors are diagnostics with exit 3, the directory is checked before anything is sent, the assembled path is measured against 104 bytes, the stale unlink is gone, bundled short flags are read left to right with a pin against the real command, the request carries the two `isatty` bits and drops `TERM`, Ctrl-C exits 130, `run_python` strips `HSQL_SESSION`, the comments are cleaned up, and the contract is the whole-package shape. The three minor items without threads are done too. The design doc moved with the code in the four places it needed to.

**Two things to fix before merge, both in the new code:**

1. **Two new tests fail on macOS** because they build socket paths under `tmp_path` instead of the file's `runtime_dir` fixture. Reproduced on Linux with a long `--basetemp`: one asserts exit 3 and gets the new exit 2, the other's own `bind()` raises. Inline at `test_hsql_session.py:561`. CI has not reported on this head yet.
2. **`check_runtime_dir` uses `os.stat`, which follows symlinks**, so an attacker-owned link to a directory the user owns privately passes the check and can be repointed before `connect()`. Reproduced. `os.lstat` closes it. Inline at `session.py:170`.

Three nits inline: a `TextIO` annotation in place of a `type: ignore`, two docstrings that describe removed behavior or PR sequencing, and a note for PR 2 that the server should refuse `-f -` with no stdin section so any scan miss is visible.

## Not checked
- I did not run on macOS; the path-length failure is inferred from the `tmp_path` layout there and reproduced on Linux with an equivalent length.
- I did not re-audit the design doc beyond the four sections the reply said changed.
- I did not re-verify the module-count budget; the test that pins it passed.

---

tests/unit_tests/test_import_hygiene.py
tconbeer: Agreed on leaving `XDG_RUNTIME_DIR` and `TMPDIR` inherited. Now that a bad one is a diagnostic rather than a traceback, the only tests that could see it name a session and supply their own runtime dir, and a child whose environment differs from the parent's on that axis would hide a platform problem rather than surface one. Verified the `HSQL_SESSION` fix here: 48 passed with `HSQL_SESSION=devbox`. Resolving.

---

tests/unit_tests/test_hsql_session.py
tconbeer: Right, and CI reported it while you were writing this — 5 macOS and 5 Windows jobs red on 7639299. Fixed in fdf5fd5, which was pushed before this review arrived.

The `runtime_dir` fixture's short base is now a `short_tmp_path` fixture that both tests use, and `runtime_dir` builds on it, so there is one place that knows why.

I reproduced it your way before and after, since the suite here runs on Linux:

```
$ pytest --basetemp=<86-byte macOS-like path> tests/unit_tests/test_hsql_session.py -n0 -q
2 failed, 93 passed        # with tmp_path, as CI had it
96 passed                  # with short_tmp_path
```

Windows had a third failure from the same commit, which you would not have seen on Linux: `_cannot_be_served` checked `AF_UNIX` before the empty name, so `hsql --session` there answered "sessions need a unix socket" instead of naming the flag. The checks are now ordered most specific first.

---

src/harlequin/hsql/session.py
tconbeer: Good catch — I reproduced it exactly as you describe, and it is fixed in d80471d with `os.lstat`.

```
before:  check_runtime_dir(link) -> ACCEPTED
after:   refused: /tmp/.../link is not a directory
         a real 0700 directory this user owns: still accepted
```

`test_a_symlink_to_a_private_directory_is_refused` pins it, and the docstring records why it is `lstat` so nobody "fixes" it back.

One scope note, since I would rather state it than imply more than the change does: this covers the **final** path component, which is where the attack lives — `/tmp/hsql-<uid>` under the fallback, `hsql` under `$XDG_RUNTIME_DIR`. An intermediate symlink is still resolved, but those components come from the caller's own `XDG_RUNTIME_DIR` or `TMPDIR`, so an attacker who controls them already controls the environment.

The design's §4.7 now says `lstat` and why, so PR 2's server refuses the same way before it binds.

---

src/harlequin/hsql/client.py
tconbeer: Agreed, and recorded rather than fixed here — it is the server's check to make. The design's §4.7 now carries it as a PR 2 obligation:

> **The server refuses a request whose argv names `-f -` and carries no stdin section.** The client's argv scan is a scan, not a parse, and its set of boolean short flags is hsql's own — a boolean short an *adapter* declares is not in it, so `-zf -` for such a `-z` would forward no stdin. One check on the server turns every miss of that shape, including ones nobody has thought of, into an exit 2 rather than a silently empty script. The server never reads its own stdin.

That is the better guarantee than making the scan cleverer, because it fails closed on the cases neither of us has enumerated.

---

src/harlequin/hsql/client.py
tconbeer: Taken in d80471d — `TextIO` under the existing `TYPE_CHECKING` block, and the `type: ignore` is gone. mypy is clean, and the `AttributeError`/`ValueError` branch still covers a replaced or closed stream at run time.

---

tests/unit_tests/test_hsql_session.py
tconbeer: Both rewritten in d80471d, close to your wording — "Any errno from `connect()` other than nothing-is-listening is a diagnostic and exit 3, not a traceback", and "The client refuses it; click does not declare the option, so its own answer would be 'No such option'."

Fair hit that I reintroduced the class I had just swept: describing what the commit removed is the same failure as describing PR staging, and I wrote both while writing the fix for the first round.

---

tconbeer:
## Standing down on an intermittent macOS failure — not this PR's

Every review finding is fixed and pushed (`d80471d`). One CI failure remains, and I do not think it belongs to this PR. Naming it rather than re-running quietly.

**What is failing.** On `fdf5fd5`, macOS 3.13:

```
       - ValueError: No Tab with id 'tab-1'
```

with the traceback through `EditorCollection.action_new_buffer` → Textual's `Tabs.refresh_active` → `Tabs.validate_active`, which raises because `query("#tabs-list > #tab-1")` is empty when the deferred activation runs.

**Why it is not this PR's.**

- The test file came from `main` (`6c67567`, #1129). My diff touches nothing under `src/harlequin/components/`, `src/harlequin/app.py`, `src/harlequin/ssh.py` or `tests/functional_tests/` — `git diff --stat origin/main...HEAD` over those paths is empty.
- `main` is green, so I cannot claim "red on the base branch". But #1129's own commit message records a macOS race in this same test family that was already fixed once: *"`wait_for_workers` returns when `get_catalog()` has answered, which is before the app has handled the message carrying its result... That raced on macOS 3.10."*
- **It moves between runners**, which is what makes me call it a race rather than a version bug: macOS 3.13 failed on `64478f5` and `fdf5fd5` and **passed** on `e534693`, `1832855` and `d80471d`; macOS 3.14 passed on all of those and failed on `d80471d`.
- 12/12 passes locally on Linux for that test alone; the full suite is green here on every commit.

**What I could not check:** the macOS 3.14 log for `d80471d` is truncated server-side (659 bytes retained), so I can only say that job failed with the same signature — the `Store snapshot report on failure` step ran, as it does for a snapshot-bearing failure — not that it was the same test. Everything above about the named test comes from the 3.13 log, which I have in full.

**Proposed patch, which I have not applied** because it is app code this PR does not touch:

```python
# components/code_editor.py, action_new_buffer
await self.tabs.add_tab(Tab(f"Tab {self.counter}", id=new_buffer_id))
if activate:
    self.tabs.active = new_buffer_id
```

`Tabs.add_tab` already schedules its own `refresh_active()` that sets `active` after the mount resolves, so the tab is activated twice, and the deferred one raises if the tab is not in the DOM when it fires — which is what a test tearing down its app can produce. Either awaiting the activation the mount already performs, or guarding `validate_active` upstream, would close it. It is Textual's `Tabs`, not one of the component libraries we own, so the guard likely belongs here rather than upstream.

Happy to take that on in a separate PR if you want it, rather than widening this one.

I am re-running the failed job once now, which is the single re-run I will spend on this.

---

tconbeer:
The re-run is green — all 32 checks pass on `d80471d`.

The macOS flake is now #1133, which asks for a review of the stateful tests as a class rather than just this one. The `code_editor.py` double-activation fix I proposed above is written up there, so nothing about it is pending on this PR.

---

Issue #1134, opened 2026-09-04
Move the Windows tzdata check and download off the start-up path


tconbeer:
On Windows, `harlequin` calls `check_and_install_tzdata()` in `cli.py` before the app is built, and the app does not start until it returns. In the worst case that is two network downloads — the IANA `tzdata-latest.tar.gz` and CLDR's `windowsZones.xml` — plus a tar extraction, all in front of a blank terminal. On a slow or captive network it can be a long wait, and on a network that hangs rather than fails there is no timeout at all: `urlopen` is called with no `timeout=`, so it inherits the socket default of none.

Nothing on the way to the first paint needs the answer. The database is for pyarrow's `timestamptz` handling, which is not reached until a result set with a tz-aware timestamp arrives in the Results Viewer.

### What to do

Start the app first, then do the check and any download on a worker thread, and set the timezone db path when it finishes. The two messages it can produce already have places to go in a running app: the "downloading one now" warning becomes a notification, and the failure — today a `HarlequinTzDataError` that exits 2 before anything is drawn — becomes an error modal or a warning toast, since a session with no tzdata is perfectly usable until a `timestamptz` shows up.

Worth doing at the same time:

- Give both `urlopen` calls a timeout, so a network that hangs cannot hang the download thread indefinitely.
- Decide what a query that loads `timestamptz` should do while the download is still in flight. Waiting on the worker is probably right; crashing is what happens today if the database is absent.

`hsql` does not call this at all and should stay that way.

### Notes

- `--no-download-tzdata` / `no_download_tzdata` keeps its meaning; it just skips the worker instead of skipping an inline call.
- The check is cheap when a database is already present (one `pc.assume_timezone` call), so the common case is not what this is about — it is the first run on a machine, and every run on a machine where the download keeps failing.
- Related, both closed: #652, #662.

PR #1135, opened 2026-09-04
M4 6/14: one query log, written by both commands

tconbeer:
Groundwork for #429. Nothing reads the store yet, so this closes nothing on its own: `hsql --history` / `--history-search` (PR 7) and the History screen (PRs 8–9) are what close it.

> Updated after review. The four blocking findings, every "should fix" and nit, and three rounds of CI are addressed; the bullets below describe what is actually on the branch. The review thread has the round-by-round detail.

`harlequin.query_log` is a new module below both front ends — one SQLite database at `<user_state_dir>/harlequin/history.db`, one row per statement, written by `hsql` and by the IDE.

- **The schema and its migrations.** `pragma user_version` plus an ordered `MIGRATIONS` list, each entry a tuple of statements run inside the `begin immediate` that claimed the right to run them — so two processes opening a new store at once don't both create it. A newer store, or one at a version out of range, is left alone rather than half-migrated.
- **The pragmas are the design, not tuning.** `journal_mode=wal` + `synchronous=normal` are what make a write cheap; the busy timeout is what makes a second writer a non-event — 5s for `hsql`, 250ms for the IDE, where a missed row beats a stalled worker. A store that can't enter WAL (a network home directory has no shared memory) falls back to the default journal.
- **Two-phase write, in both commands.** `write()` records the statement as the database is asked to run it and returns a row id; `update()` completes it once the rows are known. So a session killed mid-fetch has still recorded the query, which the pickle written at `action_quit` does not. In the IDE both phases happen on the query workers — `_execute_query` writes, `_fetch_data` completes, `_cancel_query` marks whatever the fetch never finished — so no sqlite I/O touches the event loop. In `hsql`, `_execute_all` records through a generator `execute()` consumes one at a time, which is what leaves a row for an eager adapter cancelled inside `execute()`.
- **The store opens on the first write.** Constructing a `QueryLog` costs 3.4µs and touches no disk, so the IDE holds one from `__init__` without putting a mkdir, connect, pragmas and retention delete in front of the first paint.
- **Failure never fails a query.** A store that can't be opened, migrated or written disables itself and records why, in one place; `hsql` says so once on stderr, the IDE notifies once.
- **Retention** trims to the newest 100,000 rows, once per process, as a `delete`. The store is `0600` under a `0700` directory: it is an indexed log of every statement this machine has run.
- **Opt-out** is `--no-write-history`, a flag on both commands and so a profile key of the same name.
- **Redaction.** `sql` and `error` go through `redact.redact_sql()`, added here: `redact_text()` alone masks only what a caller registered, so a credential *typed into a query* — `attach 'postgres://u:pw@h'`, `create secret (… key_id '…', secret '…')`, `set s3_secret_access_key = '…'` — would have reached disk in plaintext. It now runs the URI and DSN patterns plus a `NAME 'literal'` pattern over the statement as well. An adapter's declared `connection_id` is masked too, since it reaches a column.

Two things had to move so the two commands key one connection identically — without which one store with two writers isn't one history:

- `get_connection_hash()` out of `catalog_cache` (which imports rich through `harlequin.history`, so `hsql` could not reach it) into `query_log`; `catalog_cache` re-exports it. New `connection_id()` is the single place either command derives the id, tunnel included.
- The IDE now drops the profile keys `hsql` reads for itself before it hashes what's left and hands the rest to the adapter — as `hsql` has always dropped the IDE's own — skipping any the adapter actually declares. And `no_download_tzdata` now comes off the config on every platform, not only Windows, where the `and` short-circuit was leaving it in the hash and passing it to the adapter as an undeclared kwarg.

Review also caught that `hsql` was hashing the `-c`/`-f` text into the id: for any adapter whose `connection_id` is `None` — the ABC default — every query text was a different connection and none matched the IDE's. Both keys now come off with the rest.

**SQLite rather than the JSONL the roadmap named**, and #429 is what decides it. Measured over 100,000 records: a tail read is a wash (8.2ms vs 16.7ms), a write is 14µs apart, and `sqlite3` imports *cheaper* than `json`. Search is not close — **396ms** as a full file scan against **35ms** as a `like`, and 400ms per keystroke is not a filter anyone can type into. Three second-order effects all point the same way: the reverse-scan reader never has to exist, a schema change becomes an `alter table` rather than a version bump that discards (wrong for the one store whose contents can't be re-fetched), and the file is a database you can open in Harlequin itself. What's given up is `tail -f`; `hsql --history --jsonl` (PR 7) gives back the format if not the follow.

**`--no-write-history` rather than `history = false`** (your call, and it replaced my first pass): only one of the two values changes behavior, and it keeps the word free for the `--history` mode that reads the store back — a key called `history` and a mode called `--history` would have been one word meaning two things. Being a real click parameter on both commands is also what let the `SHARED_ONLY_KEYS` machinery I'd added go away: the key now reaches `parse_profile_options()` and the generated JSON Schema the way every other option does.

**Two phases rather than one write at the end.** One write per statement would be simpler, but it can't say what happened to a statement that was executed and never fetched (`--result last`), and it loses the crash-safety that is half the point. The cost is a second sqlite op per fetched statement — 1.6ms for fifteen statements, end to end.

**Synchronous writes rather than a queue.** A writer thread would keep the query path clear of sqlite entirely, but a queue that has not flushed loses exactly the crash-safety the two-phase write exists for, and `hsql`'s `os._exit()` on timeout depends on every write being its own committed transaction.

**Trade-off accepted:** `hsql` now imports `sqlite3` on any run that connects (+4ms), because `connection_id()` lives in `query_log`. It is deferred out of the module scope, so `hsql --version` and `--help` still pay nothing — measured at 138ms, and importing neither `sqlite3` nor rich.

**Left for PR 8, deliberately:** the connection id is a hash of whatever is left of the config after each command pops its own keys, so an adapter declaring an option one command reads for itself would key differently in the two. Hashing the adapter's *declared* options instead makes it agree by construction; that rewrites every existing hash, and PR 8 bumps `CACHE_VERSION` anyway.

### Rebased onto #1136 (crash recovery)

The two meet in one place that needed fixing: **a crash report quotes the user's SQL, and it under-redacted it.** `build_crash_report()` masked its whole document with `redact_text()`, which reaches only values a caller registered — and a credential written into a statement registers nothing. So `attach 'postgres://u:pw@h'` in the active buffer, or in the `-c` that `hsql`'s argv carries, reached a file the report itself asks the user to paste into a public issue. Both places now use the `redact_sql()` this PR added, which is the same masking the store gets.

Two smaller seams: `adapter_name` arrived on `Harlequin.__init__` from both sides, and the mocked app in the connection-id test needs a real `return_code` now that the command exits with it. The recovery path itself needs nothing from the query log and the log needs nothing from it — every row is committed when the statement runs, which `test_a_crash_mid_fetch_keeps_the_query_it_was_running` now pins against the real handler rather than against a held-open fetch.

### Rebased onto #1138 (warm sessions)

A session's server runs `cli.run()` per request on a connection it opened once, so a served request never builds the adapter instance the connection id is derived from. Left alone, **every query a warm session ran would have been logged under no connection**, while the same database reached cold was logged under its hash — two lists for one database, which is the thing keying a connection exists to prevent. The session now derives the id when it connects and carries it on `Served`, and `_connection_for()` returns the pair whichever half of the session it came from. Fixing it once at serve time is sound rather than convenient: a served request is refused every option the id reads, so the answer cannot change between requests, and `--session-reset` reopens the same options.

`--no-write-history` joins `PER_REQUEST_OPTIONS`, which the partition test requires of every parameter. Per-request rather than the server's, because the server runs no SQL of its own, and because a profile is still read where the caller is — so `no_write_history = true` covers a whole session and the flag covers one invocation of it.

`--timeout`'s `os._exit()` is still what the cold path does, so the argument for a committed write per statement is unchanged; a server sets `abandon` and unwinds the request instead, which is strictly kinder to the store.

### Rebased onto #1140 (identity, rejection, `--session-status`)

A session now records the connection options it resolved at start-up, and a served request is refused the ones that differ. `--no-write-history` is not one of them: it is in `PER_REQUEST_OPTIONS`, and `connection_option_names()` is a positive test over hsql's connection keys plus the adapter's declarations, so the opt-out never reaches a session's identity. A session started without it still serves a request that carries it, and `--session-status` does not report it — which is what per-request means. The connection id is unaffected for the same reason: it is derived once at serve time from the options a served request is refused, so the identity and the id describe one snapshot.

Two seams in the diff: `redact_conn_str()` now masks a non-string item whole **and** masks a string's spans through the `_secret_spans()` split this PR added, and `Served` carries the session's `identity` beside its `connection_id`. #1140's `WarmSession` helper also grew a per-test `env`; it takes that alongside the `base_env` this PR added, so a session still runs with the redirected state directory the isolation fixture builds rather than writing into the developer's real history.

### Rebased onto #1147, #1150 and #1153 (cancellation, lifecycle, docs)

The warm-session series is complete, and two of its three code PRs touch the path the log writes from.

**Cancellation (#1147).** `run.stopped` used to mean one thing — the clock ran out — and now means that or *the caller gave up on a served request*. Both reach the log through the same two calls, so the rows a cancel leaves were already right; what was missing was any coverage from the served side, and one of the two orderings is load-bearing. `_execute_all` writes each statement's row from a generator `execute()` draws from, so #1147's "already cancelled before the first statement" guard has to sit **ahead** of that generator: behind it, a request cancelled while it queued would put a query the database never saw into the caller's history. Two tests, each failing against the code it pins — one that such a run records nothing (and does not even create the store, since the first write is what opens it), and one that a cancel between two fetches marks the statement in flight and every one after it, which is the served half of what `--timeout` already had.

**Lifecycle (#1150), and a real cost it exposed.** `RETENTION_ROWS` has said "trimmed once per process" since it was written, but the trim ran on every `QueryLog` that opened the store — one per command, which *was* one per process until a session started answering many. The walk it costs grows with the store, so a session's per-request sqlite work grew with the user's history: against a store at the 100,000-row cap, a request's first write measured **12.3ms, of which ~10ms was retention the session had already done**. Now trimmed at the first open of a store and **2.0ms** after, which is what the docstring claimed. Retention is a soft cap on a history kept for months, and this is what the IDE has always done with its one long-lived log.

Nothing else needed reconciling. `--idle-timeout` and `--max-lifetime` bring a session down the way a stop signal does — the request in flight still gets its answer — so its rows are completed by the run that wrote them, and the log is per-request rather than the session's, so a session coming down strands nothing. `--no-write-history` stays in `PER_REQUEST_OPTIONS` and out of the identity a served request is compared against, as of #1140.

Two seams in the diff: `_fetched`'s three cancel returns and `_execute_all`'s guard interleave with #1147's, and four of the `Server(...)` constructions its tests added predate the required `connection_id` — squashed into the commit that made it required, so each commit still typechecks. #1153 is docs, and touches nothing here. #1150 also landed its own fix for the `test_toggle_full_screen` race this branch was carrying, so that commit is gone.

- [ ] No.
- [x] Yes; I haven't opened a PR, but the gist of the change is: PR 10 of the plan is the docs PR for this release — where the history lives, what it holds, how to search it from either front end, that it's a SQLite file you can open in Harlequin, and `--no-write-history` on both command pages.

- [x] Yes.

`tests/unit_tests/test_query_log.py` covers the row round trip, the two-phase write, migration from `user_version = 0` and on a store already current, a store written by a newer Harlequin, retention from the oldest end, the lazy open, an unopenable store, an unwritable one, the WAL fallback, 4 processes × 100 concurrent inserts, and redaction of both a registered secret and a credential nothing registered. Beside those: what `hsql` records for every shape of statement, a statement cancelled inside `execute()`, the opt-out from the flag and from a profile in both commands, a non-boolean value refused rather than read as true, the "a log failure never fails the query" path, the IDE's row existing mid-fetch and being completed rather than duplicated, a cancel leaving a finished query alone, and two tests that pin the connection id — both of which fail if the `-c`/`-f` pops are reverted. A subprocess guard asserts a logging `hsql -c` run records the query **and** imports no rich.

The four rebases add nine: a crash report masking a credential typed into the active buffer, the same for one typed into `hsql -c`, a crash mid-fetch keeping the statement it was running, a served request logging under the session's connection, a served request opting out for itself, and — end to end, over a real socket — a warm run and a cold one against one database landing in one history under one id; then a served run cancelled before it started recording nothing, every statement a served cancel stopped saying so, and a second log on one store not trimming it again. Each fails against the code it pins.

Test isolation: an autouse fixture points `default_path()` at `tmp_path`, and the subprocess tests share an environment that redirects `XDG_STATE_HOME` and the `WIN_PD_OVERRIDE_*` names platformdirs actually reads on Windows — three of them had no `env` at all and were writing into the developer's real history. #1138's two new subprocess helpers (`hsql_subprocess` and the `--serve` process itself) had the same gap and now build their environment from the same `clean_env`.

Two test waits were also unsound before this PR and are fixed here, because the query log's writes were enough to expose them: `wait_for_workers` read `app.workers` once, though a worker's result arrives as a message whose handler starts the next one; and `test_toggle_full_screen` focused the Results Viewer without waiting for the query, so which widget went full screen depended on the table having landed. Both were reproduced under injected latency before being changed, and neither assertion moved — and #1150 has since landed the same fix for the second one.

---

Carried out of the thread: the plan's §3.2 and §4 are updated for the `--no-write-history` spelling; the Windows tzdata check that blocks start-up is filed as #1134; and the `EditorCollection`/`Tabs.add_tab` race that fails `test_a_refresh_that_reconnects_does_not_ask_for_a_second_one` is a `main` flake, red on the base commit and now reproduced on Linux — commented on in the thread, not fixed here.

https://claude.ai/code/session_017d34JLqrhGY3mMpi2Q61JU


tconbeer:
CI note: **`MacOs-latest - 3.14`** is red on `1f08878`, on `tests/functional_tests/test_ssh_recovery.py::test_a_refresh_that_reconnects_does_not_ask_for_a_second_one` — `ValueError: No Tab with id 'tab-1'`. The other 14 jobs pass, and the two non-test checks pass.

**This one isn't this PR's.** The same test fails the same way on `main` at `e278c0b`, the base commit ([run 33833836899](https://github.com/tconbeer/harlequin/actions/runs/33833836899)) — there on `MacOs-latest - 3.10` and `3.11`, while 3.14 passed. Here it's 3.14 while 3.10 and 3.11 pass. It moves between macOS Python versions run to run, which is what says it's a timing race rather than a version incompatibility, and it is red on the base branch either way. No fix for it exists yet to port, so nothing is carried into this PR.

**Where it actually is**, from the traceback — and it is not really in the SSH test:

```
EditorCollection.on_mount           code_editor.py:415
  action_new_buffer()               code_editor.py:456
    Tabs.add_tab(Tab(id='tab-1'))
      refresh_active()              textual/widgets/_tabs.py:497   <- await mount_await; self.active = ...
        validate_active()           textual/widgets/_tabs.py:580   <- query("#tabs-list > #tab-1") empty
```

Textual's `Tabs.add_tab` awaits the mount, then assigns `active`, and the validator re-queries the DOM for the tab it just mounted. On a loaded runner that query comes back empty and the assignment raises. Two things in the frame say the mount is the problem rather than the id: `mount_await` is `AwaitMount(parent=Horizontal(id='tabs-list'), widgets=[])` — nothing to wait for — and the `Tab` in `activated_message` already carries `id='tab-1'`, so the widget exists but is not in the queried tree.

That means it fires during **app start-up**, in `EditorCollection.on_mount`, which every functional test runs — so it isn't specific to this test, it just lands on whichever one the runner is slow enough to lose. That also fits it landing on a different Python version each run.

**Proposed fix, for whoever picks it up** (not pushed here — it's `main`'s to fix, and doing it here would widen this PR): the tab that `on_mount` creates is the one the collection always makes, so `action_new_buffer()` could mount the `Tab` and set `Tabs.active` in one step rather than letting `add_tab`'s post-mount validator race the DOM — or `on_mount` could await `self.tabs.query(Tab)` settling before the first activation. Given [AGENTS.md's "fix things upstream"](https://github.com/tconbeer/harlequin/blob/main/AGENTS.md), the real home is arguably a Textual issue: `add_tab` should not assign a reactive whose validator can fail against a DOM it just wrote to. Worth a look alongside #1129, which already fixed one macOS race in this same test.

One re-run of the failed job is queued to confirm — that is the single re-run, not a retry loop. If it comes back green the PR is green; if it fails again on a different macOS version, that is more evidence for the same conclusion, and I will not re-run a third time.

Separately, ignore the **"23 mismatched snapshots"** line in that job's log: snapshots are committed from 3.10 and this is 3.14, CI runs `--snapshot-warn-unused`, and exactly one test actually failed.

---

src/harlequin/hsql/cli.py
```diff
@@ -998,8 +1044,14 @@ def _connect(
         diagnostics.report_error(e)
         ctx.exit(ExitCode.USAGE)
 
+    keyed = connection_id(
```
tconbeer: **Bug: `-c` / `-f` text is hashed into hsql's connection id.** `command` and `file` are `expose_value=True` and nothing pops them from `values` before `_connect()`, so the fallback material here is `{"conn_str": ..., <adapter opts>, "command": ("select 1",)}`. Verified in-process: for `hsql -a duckdb --no-init :memory: -c "select 1"` this function receives `options={'no_init': True, 'command': ('select 1',)}`, where the IDE passes `{'no_init': True}`.

For any adapter whose `connection_id` is `None` (the ABC default, so every out-of-tree adapter that doesn't override it) every distinct query text becomes a distinct "connection", none of which ever matches the IDE's — which defeats the one-list goal and also fragments the catalog cache key hsql now shares.

Fix: `values.pop("command", None); values.pop("file", None)` alongside the other hsql-owned pops before `_connect`, and pin it with a test that uses an adapter whose `connection_id` is `None` (see the note on `test_the_two_commands_key_one_connection_the_same_way`).

---

src/harlequin/app.py
```diff
@@ -394,13 +406,38 @@ def pop_screen(self) -> "AwaitComplete":
...
+        error: BaseException | None = None,
     ) -> None:
         if self.history is None:
             self.history = History.blank()
         self.history.append(
             query_text=query_text, result_row_count=result_row_count, elapsed=elapsed
         )
+        # a negative count is how this app has always spelled "that one raised"
+        failed = result_row_count < 0
+        self.query_log.write(
```
tconbeer: **The IDE is not two-phase, so the changelog's "recorded as it runs" only holds for hsql.** This `write()` happens from `append_to_history`, whose callers are `QueriesExecuted` (DDL only), `ResultsFetched` and `QueryError` — i.e. after `_fetch_data` has drained the cursor. A session that hangs or crashes *during* a fetch (the case the PR description cites) loses that query, and a query the user cancels (`QueriesCanceled`) is never recorded at all. `QueryLog.update()` exists for exactly this and the IDE never calls it.

Suggest `write()` each statement from the `QueriesExecuted` handler (it has `message.cursors` with `ExecutedStatement.statement.sql`), keep the row ids on the app, and `update()` them in `load_tables` / `QueryError` / `QueriesCanceled` (status `canceled`). Otherwise narrow the changelog claim to hsql.

Minor, same spot: for a failed statement this writes `elapsed_ms=0.0` where hsql writes `NULL`; pass `None` when `failed` so the two commands agree.

---

src/harlequin/query_log.py
```diff
@@ -0,0 +1,380 @@
...
+        one leaves it out and gets now, in UTC.
+        """
+        moment = run_at if run_at is not None else datetime.now(timezone.utc)
+        values = (
+            moment.isoformat(),
+            self.program,
+            self.connection,
+            self.profile,
+            self.adapter,
+            # a query can carry a credential -- `attach 'postgres://u:pw@h'` --
```
tconbeer: **Security: this comment overclaims.** `redact_text()` is a literal `str.replace` over the values `hide_secrets_in()` registered; it never applies `_URI_PASSWORD` / `_DSN_PASSWORD`. So `attach 'postgres://u:pw@h/db'`, `create secret (type s3, key_id '…', secret '…')` and `set s3_secret_access_key = '…'` all reach disk in plaintext unless the same value happens to also be in the profile — and so does a driver error that quotes them. `test_a_secret_never_reaches_the_store` only passes because it calls `hide_secrets_in({"password": SECRET})` first.

The in-memory `History` pickle already stored raw SQL, so this isn't a regression, but this store is new, `sqlite3`-queryable, and the input to `hsql --history`. Suggest a `redact_sql()` (or a pattern pass inside `redact_text()`) that also runs `_mask_spans` with the URI/DSN patterns plus DuckDB's `(KEY_ID|SECRET|TOKEN|PASSWORD)\s+'…'` form, apply it to `sql` and `error` here and in `update()`, and make the test use an unregistered DSN. At minimum, correct the comment.

---

src/harlequin/hsql/cli.py
```diff
@@ -1622,9 +1718,14 @@ def _execute_all(
         run.statements += 1
         if item.error is not None:
```
tconbeer, line 2245: **A statement cancelled during its execute phase is never recorded.** `record()` runs only after `connection.execute()` has returned an item, and the `if run.stopped: break` above runs before it. With an eager adapter (SQLite's `cursor.execute()` steps the query, and it implements cancel) `hsql -a sqlite --timeout 1 -c "<heavy aggregate>"` raises inside execute, the yielded error item hits `break`, and there is no row. DuckDB is lazy, so the same timeout lands in `_fetched` and is recorded `canceled` — the history differs by adapter. Same root cause: a process killed while an eager statement runs leaves no row, contrary to "recorded as it runs".

Fix: record each statement *before* handing it to `execute()` (it's a lazy generator, so the caller controls the ordering), or at least record the item as `canceled` before the `break`.

Related nit in `_fetched` (line ~1818): when the clock runs out between fetches only the first unfetched item is marked `canceled`; the rest stay `ok`/`rows=NULL`, indistinguishable from `--result last`.

---

tests/unit_tests/conftest.py
```diff
@@ -55,6 +56,8 @@ def run_python(tmp_path: Path) -> Callable[[str], subprocess.CompletedProcess[st
         "XDG_CONFIG_HOME": str(tmp_path / "xdg"),
         "APPDATA": str(tmp_path / "appdata"),
         "LOCALAPPDATA": str(tmp_path / "localappdata"),
```
tconbeer, line 53: **Test isolation has two holes, and this PR adds the first *write* through them.**

1. Three existing tests in `test_hsql.py` spawn a real subprocess with no `env=` at all (`test_diagnostics_follow_the_data_they_describe` ~L426, `test_a_run_that_outlasts_the_grace_period_still_exits_4` ~L4294, `test_a_real_duckdb_query_times_out_and_says_so` ~L4346). The autouse `query_log_path` monkeypatch can't reach a child, so each of them now writes into the developer's real `~/.local/state/harlequin/history.db` (macOS: `~/Library/Application Support/harlequin/`). Give them this env (a shared fixture) or pass `--no-write-history`.

2. On Windows, `LOCALAPPDATA`/`APPDATA` in the env do not redirect platformdirs — it prefers `SHGetKnownFolderPath` via ctypes and only falls back to env vars when neither ctypes nor winreg import. So `test_a_run_that_records_its_queries_does_not_render_one` writes `select 1` into the real `%LOCALAPPDATA%\harlequin\harlequin\history.db`, reads it back from the same place, and passes. Setting `WIN_PD_OVERRIDE_LOCAL_APPDATA` / `WIN_PD_OVERRIDE_APPDATA` too would close it (check the platformdirs floor in pyproject supports them).

---

tests/unit_tests/test_hsql.py
```diff
@@ -4362,3 +4363,181 @@ def test_a_real_duckdb_query_times_out_and_says_so(tmp_path: Path) -> None:
...
+        "duck",
+        "-c",
+        "select 'hunter2-and-then-some' as a",
+    )
+    assert res.exit_code == ExitCode.OK
+    (row,) = logged(query_log_path)
+    assert "hunter2-and-then-some" not in row["sql"]
+
+
+def test_the_two_commands_key_one_connection_the_same_way(
```
tconbeer, line 4746: This test would pass with the keying broken, and did: duckdb `:memory:` has `connection_id == ""`, so `connection_id("", …)` returns `""` for both commands whatever is in the hashed options, and `get_connection_hash` is never exercised. Use a mock adapter with `connection_id = None` (as `test_cli.py` does) or a file-backed db with `connection_id` monkeypatched to `None`, and assert the two hashes agree for a profile that sets `limit`, `format`, `theme`, `no_download_tzdata`, `no_write_history`, plus a typed `-c`.

---

tests/unit_tests/test_cli.py
```diff
@@ -132,6 +132,8 @@ def test_default(
         adapter=mock_adapter.return_value,
         profile_name=None,
         connection_hash=mock_adapter.return_value.connection_id,
+        adapter_name="duckdb",
+        record_history=True,
```
tconbeer, line 175: Only the default is pinned on the IDE side. Nothing invokes `harlequin --no-write-history`, or a profile with `no_write_history = true`, and asserts `record_history=False` reaches `Harlequin(...)`; the functional test constructs `Harlequin(record_history=False)` directly despite its docstring naming the flag and the profile key, and the non-bool rejection is only tested through hsql. Two cases mirroring `test_default` (flag typed; profile key) plus one for the usage error would cover it.

---

src/harlequin/query_log.py
```diff
@@ -0,0 +1,380 @@
...
+    "error",
+)
+"""Every column a writer supplies, which is all of them but `id`."""
+
+_INSERT = "insert into queries ({}) values ({})".format(
+    ", ".join(f'"{column}"' for column in COLUMNS),
+    ", ".join("?" * len(COLUMNS)),
+)
+
+_BUSY_TIMEOUT_MS = 5000
```
tconbeer: In the IDE every `write()` runs synchronously on the Textual event loop (`append_to_history` is called from `@on` handlers), and the first one also does mkdir + `begin immediate` + migrate + a retention `delete`. Verified: with another connection holding `begin immediate`, `write()` blocks the full 5s and then disables the log. That's one 5s TUI freeze per process whenever someone has `sqlite3 history.db` open mid-transaction, an NFS lock is stuck, or a checkpoint is slow — and it's the one place in `app.py` that does DB work on the main thread. A much shorter timeout for the IDE (a few hundred ms; a missed row beats a frozen UI) or a `@work(thread=True)` writer would keep the invariant.

Also, the docstring's "eight of them, 200 inserts each … 0.15s" is a benchmark; per AGENTS.md that belongs in the PR description. `"""How long a writer waits on a lock held by another process."""` is enough.

---

src/harlequin/query_log.py
```diff
@@ -0,0 +1,380 @@
...
+        """The store, configured and migrated, or None having recorded why.
+
+        Called by the first write rather than by `__init__`, so that opening a
+        file, migrating it and trimming it is work a session pays for when it
+        runs a query and not while it is starting up.
+        """
+        store = self._path if self._path is not None else default_path()
+        db: sqlite3.Connection | None = None
+        try:
+            store.parent.mkdir(parents=True, exist_ok=True)
```
tconbeer: Low: `history.db`, `-wal` and `-shm` come out `0644` under a `0755` dir (verified). That matches `catalog_cache.py` / `editor_cache.py` today, so it doesn't lower the bar, but this is a world-readable, indexed log of every statement on a shared host or NFS home. `os.open(store, os.O_CREAT | os.O_RDWR, 0o600)` before `connect()` fixes the main file, and SQLite's unix VFS gives the sidecars the main file's mode; `mkdir(mode=0o700, …)` for the leaf dir. No-op on Windows.

---

src/harlequin/query_log.py
```diff
@@ -0,0 +1,380 @@
...
+
+    Under `begin immediate`, so that two processes opening a new store at once
+    do not both create it. A store written by a newer Harlequin is left alone:
+    its columns are a superset of these, so the inserts still name what they
+    mean.
+    """
+    db.execute("begin immediate")
+    try:
+        (version,) = db.execute("pragma user_version").fetchone()
+        for index, migration in enumerate(MIGRATIONS[version:], start=version):
```
tconbeer, line 284: Nit: a negative `user_version` (verified with `-3`) slices `MIGRATIONS[-3:]`, runs the create-table migration, and writes `user_version = -2`; the next process reruns it, hits `table queries already exists`, and logging is disabled for good with a file the user has to delete by hand. `version = max(version, 0)`, or treat anything out of range as "leave alone".

The transaction handling itself checks out: explicit `begin immediate` + transactional `user_version` + `commit()` is sound under the legacy sqlite3 semantics on 3.10–3.14, `journal_mode` is set outside as required, `lastrowid` survives the `with db:` commit, and hsql's `call_on_close(close)` can't overlap a worker-thread write because `Deadline.run` only returns after the worker's `finally`.

---

src/harlequin/query_log.py
```diff
@@ -0,0 +1,380 @@
...
+    through: Sequence[str] = (),
+) -> str:
+    """The id this connection is logged and cached under, in either command.
+
+    `declared` is what the adapter says its connection is, where it says
+    anything; the details it was built from are the fallback. Shared because
+    the two commands writing one store have to agree about which database a row
+    belongs to.
+    """
+    keyed = declared if declared is not None else get_connection_hash(conn_str, options)
```
tconbeer: Low: `declared` is written to the `connection` column verbatim, and the ABC docstring on `HarlequinAdapter.connection_id` says "typically a fully-hydrated connection string". Every installed adapter returns a path or `host:port/db`, so nothing leaks today, but a third-party adapter returning `self.conn_str[0]` would put a URI password on every row, indexed. Running the declared id through `redact_conn_str([declared])[0]` here is deterministic, so cache and log stay consistent; and the ABC docstring should say the id must not carry credentials.

---

src/harlequin/query_log.py
```diff
@@ -0,0 +1,380 @@
+"""Every query both commands have run, in one SQLite database.
+
+One store, at `<user_state_dir>/harlequin/history.db`, one row per statement,
+written as each statement runs -- so a session that never exits cleanly still
+has its history, and an agent's queries and a human's are one list rather than
+two.
+
+**SQLite because the store is searched.** A tail read is bounded by
```
tconbeer: **Comments (AGENTS.md "Conventions"): this module keeps the design argument in docstrings.** The 396ms/35ms/100,000-record benchmark, "SQLite because the store is searched", the append-only-file alternative, "State rather than cache" (`default_path`), "rather than a script" (`MIGRATIONS`), "rather than a rotation" (`RETENTION_ROWS`), "0.02ms"/"34x slower" (`_configure`), "Two phases because…" (`update`), and "Called by the first write rather than by `__init__`…" (`_open`) are all rationale or measurements the conventions route to the PR description or `docs/plans/` — and `docs/plans/m4-hsql-technical-plan.md` already carries most of them.

Suggested module docstring: *"One SQLite store of every query both commands run, written as each statement runs. Logging never fails a query: a store that cannot be opened or written disables itself and records why in `failure`, for the caller to report once."* Then one sentence of what plus at most one why on each function. The same "a store that cannot be written never fails the query" sentence also appears three times across `app.py`, `hsql/cli.py` and `diagnostics.py`; it should live once, here.

---

src/harlequin/query_log.py
```diff
@@ -0,0 +1,380 @@
...
+    conn_str: Sequence[str], config: Mapping[str, Any], *, through: Sequence[str] = ()
+) -> str:
+    """What a connection's cached catalog and query history are keyed by.
+
+    `through` is how it was reached, where that is not part of the details
+    themselves: two SSH tunnels front two databases that both look like
+    `localhost:15439`. Absent from the hashed material when there is none, so
+    an untunneled connection keys on its details alone.
+
+    Here rather than in `catalog_cache` because a headless caller keys the same
```
tconbeer: "Here rather than in `catalog_cache`" references where this used to live — exactly the "X rather than the old Y" the conventions forbid, since it reads as a live distinction once merged. Drop this paragraph; the first two were already the docstring.

---

src/harlequin/catalog_cache.py
```diff
@@ -1,39 +1,37 @@
...
 
 CACHE_VERSION = 2
 
+__all__ = [
+    "CatalogCache",
+    "get_catalog_cache",
+    "get_connection_hash",
+    "update_catalog_cache",
+]
+"""`get_connection_hash()` lives in `harlequin.query_log`, below the rich this
```
tconbeer: This isn't a docstring: a string expression after an assignment is a no-op, and nothing attaches it to `__all__`. Its content is also history ("lives in `harlequin.query_log`", "have always asked this module"). Replace with a trailing comment on the import: `from harlequin.query_log import get_connection_hash  # re-exported`.

---

src/harlequin/app.py
```diff
@@ -394,13 +406,38 @@ def pop_screen(self) -> "AwaitComplete":
...
+        result_row_count: int,
+        elapsed: float,
+        error: BaseException | None = None,
     ) -> None:
         if self.history is None:
             self.history = History.blank()
         self.history.append(
             query_text=query_text, result_row_count=result_row_count, elapsed=elapsed
         )
+        # a negative count is how this app has always spelled "that one raised"
```
tconbeer: "has always" is a history reference. `# a negative count means the query raised` says the same thing.

---

src/harlequin/cli.py
```diff
@@ -595,8 +607,24 @@ def inner_cli(
...
+        # to work here: its keys come off rather than being handed to an
+        # adapter that never declared them, as hsql drops this command's own.
+        # Which also keeps the two agreeing about a connection's id, since that
+        # is a hash of whatever is left. An option the adapter does declare is
+        # its own, however hsql spells the same word.
+        declared_by_adapter = {
+            sluggify_option_name(option.name)
+            for option in adapter_cls.ADAPTER_OPTIONS or []
+        }
+        for key in hsql_profile_keys() - harlequin_options - declared_by_adapter:
```
tconbeer, line 616: The six-line comment above is over the conventions' length and its middle sentence is a design consequence. `# drop hsql's own profile keys before the adapter sees the config; an option the adapter also declares is the adapter's` covers it. Same for the tzdata comment below: keep the original first sentence and `# popped on every platform: the remaining config is the adapter's`.

One question this block raises: the IDE keeps a profile key the adapter *declares* even when hsql spells the same word, but hsql pops `timeout`, `format`, `output`, `limit`, `result`, etc. unconditionally for itself. An adapter declaring an option named `timeout` would therefore hash (and be configured) differently between the two commands, splitting its history. No bundled adapter collides today; worth deciding explicitly rather than by accident.

---

src/harlequin/config.py
```diff
@@ -669,6 +669,26 @@ def refuse() -> HarlequinConfigError:
     return seconds
 
 
+def take_no_write_history(config: MutableMapping[str, Any]) -> bool:
```
tconbeer, line 682: "Spelled as a refusal in both commands, so that the value that changes behavior is the one a profile writes" is the naming decision from the plan; the conventions route decisions to the PR / `docs/plans/` (which already has it, with attribution). `"""Pop `no_write_history` off the config and return it. Raises: HarlequinConfigError if the value is not a bool."""` is the docstring.

---

tests/unit_tests/test_query_log.py
```diff
@@ -0,0 +1,367 @@
...
+    db.commit()
+    db.close()
+
+    assert log.write("select 2") is None
+    assert not log.enabled
+    assert log.failure is not None
+
+
+def test_a_disabled_log_writes_nothing(store: Path) -> None:
+    """`history = false`, which is the same object to everything upstream."""
```
tconbeer: Stale: `history = false` is the spelling this PR replaced with `--no-write-history`. Also in this file: "34x slower per write" (L303) and "45MB, and 35ms to search" (L366) are benchmarks, and `..._the_history_screen_has_always_shown` is a history reference in a test name — `test_the_retention_cap_matches_the_history_screen` reads the same. Same for "39% of a whole `hsql --version`" in `test_import_hygiene.py:194` and "used to lose" in `test_history_screen.py:104`.

---

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 ### Features
 
 - Press `alt+e` in the Query Editor to edit the current buffer in the editor named by `$VISUAL` or `$EDITOR`; quitting the editor with a non-zero status (like `:cq`) discards the changes ([#1102](https://github.com/tconbeer/harlequin/issues/1102)).
+- Harlequin and hsql now record every query they run as it runs, in a shared history at `<user state dir>/harlequin/history.db` — so a session that crashes keeps its history, and hsql's queries and Harlequin's are one list. Pass `--no-write-history` (or set `no_write_history = true` in a profile) to record nothing ([#429](https://github.com/tconbeer/harlequin/issues/429)).
```
tconbeer: Per `.claude/rules/changelog.md` this carries a path and a rationale clause. Suggest: *"Harlequin and hsql now record every query as it runs in one shared history, so a crashed session keeps its history. Pass `--no-write-history` (or set `no_write_history = true` in a profile) to opt out ([#429](https://github.com/tconbeer/harlequin/issues/429))."* — and "as it runs" only once the IDE is actually two-phase (see `app.py`). The store's location can go to the harlequin-web docs PR.

---

tconbeer (commented):
**Verdict: changes needed before merge** (submitted as a comment because GitHub won't let this account request changes on its own PR).

Reviewed at 8fb9168 under four lenses: correctness/logic, security/data safety, test isolation, and comment/changelog hygiene against AGENTS.md. Each was a separate narrow pass, then verified by hand. On this branch `ruff`, `mypy` (strict), `lint-imports` (4 contracts kept) and the PR's tests (138 passed) are all green.

**Blocking**

1. **hsql hashes `-c` / `-f` text into the connection id** (`hsql/cli.py:1047`). Verified in-process. For any adapter with `connection_id = None` every query text is a different "connection" and none matches the IDE's — the opposite of the PR's goal, and it now fragments the catalog cache key too. The test meant to pin this (`test_the_two_commands_key_one_connection_the_same_way`) uses duckdb `:memory:`, whose id is `""`, so it never exercises the hash.
2. **The IDE is not two-phase** (`app.py:422`): writes happen on `ResultsFetched`, so a session that dies mid-fetch loses the query and a cancelled query is never recorded. `QueryLog.update()` exists for this and the IDE never calls it. The changelog claims "as it runs" for both commands.
3. **Credentials typed inside a query reach disk in plaintext** (`query_log.py:181`): `redact_text()` only replaces values registered by `hide_secrets_in()`; it applies no DSN patterns. `attach 'postgres://u:pw@h'` and DuckDB `create secret` are not masked, contrary to the comment, and the test passes only because it registers the secret first.
4. **Merge conflict with `main`** (`pyproject.toml`, `tests/unit_tests/conftest.py`, `tests/unit_tests/test_import_hygiene.py`) — all three are adjacent additions from #1132, so the merge should be mechanical.

**Should fix**

- Execute-phase cancels under `--timeout` are never recorded for eager adapters (`hsql/cli.py:1719`).
- Three existing subprocess tests in `test_hsql.py` pass no `env=`, so they now write to the developer's real history; on Windows the `LOCALAPPDATA` env override does not redirect platformdirs at all (`tests/unit_tests/conftest.py:58`).
- No IDE-side test for `--no-write-history` typed, or `no_write_history = true` from a profile, reaching `Harlequin(record_history=False)`.
- Every IDE write is synchronous on the event loop with a 5s `busy_timeout` (`query_log.py:99`) — a locked store freezes the TUI for 5s, and it is the one DB write in `app.py` not on a worker.

**Low / nits** (inline): world-readable `history.db` (matches existing caches, but this one is a searchable query log); adapter-declared `connection_id` written unredacted; negative `user_version` bricks logging permanently; the `harlequin_options`/`declared_by_adapter` asymmetry between the two commands is a decision worth making explicitly.

**Comments and changelog** — the PR does not match the AGENTS.md conventions: seven benchmarks in docstrings, the SQLite-vs-file / state-vs-cache / refusal-spelling decisions recorded in code, "here rather than in `catalog_cache`" referencing the old home, "has always" / "used to" history references, a no-op string after `__all__`, a stale `history = false` docstring, and the same "never fails the query" sentence in three places. The inline comments carry replacement wording; `docs/plans/m4-hsql-technical-plan.md` already holds the rationale.

**Verified as correct, no action:** `_migrate`'s explicit `begin immediate` + transactional `user_version` + `commit()` on 3.10–3.14; `_trim`'s offset arithmetic; `lastrowid` after the `with db:` commit; `Statement.index` uniqueness for `rows_written`; hsql's `call_on_close(close)` cannot overlap a worker-thread write; `os._exit` on timeout loses nothing since every write is its own committed transaction; every `append_to_history` caller is on the event-loop thread; `hide_secrets_in()` runs before the first write in both commands, CONN_STR included; `read_only`, `adapter`, `no_write_history`, `no_download_tzdata`, `ssh_*` and `TUI_ONLY_KEYS` are removed on both sides before hashing; all SQL is parameterized; `str(HarlequinError)` matches hsql's `_message_for`.

## Not checked
- I did not examine the History screen or `hsql --history` read path (PRs 7–9), nor whether `run_at` as ISO text with offset sorts correctly for them.
- I did not run the full suite on 3.12–3.14, macOS or Windows; relied on 3.10/Linux locally and on the platformdirs source for the Windows finding.
- I did not verify `connection_id` for adapters not installed here (odbc, bigquery, databricks…); spot-checked duckdb, sqlite, postgres, mysql.
- I did not test the WAL fallback on a real network filesystem; `pragma journal_mode = wal` typically returns `delete` there rather than raising, so the fallback path the test simulates may not be the one that runs.
- I did not check `SshTunnel.cache_material()` contents for secrets, nor `hsql --config init` writing `no_write_history`.
- I did not assess whether `json.dumps` without `sort_keys` can order the hash material differently between commands for CLI-typed adapter options.

---

tconbeer:
Thanks — the first two findings were real bugs I'd have shipped. All four blocking items and every "should fix" and nit are addressed in `8123a69`. One item I did **not** change, with reasoning, at the bottom.

## Blocking

**1. `-c`/`-f` hashed into the connection id.** Confirmed and fixed — `command` and `file` now come off `values` with the other keys hsql reads for itself. You were right that the test couldn't fail: duckdb `:memory:` returns `""`, so `connection_id("")` short-circuits and the hash never runs. It now monkeypatches `DuckDbAdapter.connection_id` to `None` (the ABC default), and there's a second test that two different queries against one database are one connection. I verified both by reverting the pops: both fail, then pass again.

**2. IDE two-phase.** Done, and it also moved both writes off the event loop, which folds in your `busy_timeout` finding. `_execute_query` writes a row per statement as the database is asked to run it; `_fetch_data` completes it with the rows it returned or marks it `error`; `_cancel_query` marks whatever the fetch never completed. `append_to_history` is back to the in-memory history alone, so the `elapsed_ms=0.0`-vs-`NULL` divergence is gone too.

Writing that turned up a bug of my own: `logged_rows` persisted across runs, so a cancel with nothing running would have rewritten a *finished* query as `canceled`. Rows are now popped as they complete. Two tests: one holds a fetch open and asserts the row exists mid-fetch with `rows IS NULL` and that the same `id` is completed rather than a second row written; one asserts a cancel after a finished run leaves it `ok`.

**3. Credentials in query text.** You're right that the comment overclaimed — `redact_text()` is a literal replace over registered values only. Added `redact.redact_sql()`, which also runs `_URI_PASSWORD`/`_DSN_PASSWORD` plus a new `NAME 'literal'` pattern for `create secret (… key_id '…', secret '…')` and `set s3_secret_access_key = '…'` (`key_id` isn't in `_SECRET_NAME`, so it's named explicitly). Applied to `sql` and `error`, in both `write()` and `update()`. The tests use a credential nothing registered, and one asserts `select … from tokens where name = 'alice'` is stored verbatim, since over-redaction shouldn't eat the log.

One thing that fell out: two patterns can match the same literal, and masking it twice ate the closing quote (`set … = '********`). `_mask_spans` now skips a span it has already covered.

**4. Merge conflict.** Already resolved — the merge landed in `1f08878`, after the `8bf9168` you reviewed. The `run_python` docstring was the only textual conflict.

## Should fix

- **Execute-phase cancels.** Fixed as you suggested: `_execute_all` records through a generator `execute()` consumes one at a time. That needed `query.execute()` to take an `Iterable` rather than a `Sequence` — which is what it has always done with it, and the docstring now says so. Pinned with sqlite (eager, cancels inside `execute()`) asserting `canceled`, and every unfetched statement is now marked, not just the first.
- **Subprocess isolation.** All three now share a `clean_env` fixture with `run_python`. It also sets `WIN_PD_OVERRIDE_LOCAL_APPDATA`/`WIN_PD_OVERRIDE_APPDATA` — thank you for that, I'd assumed `LOCALAPPDATA` was enough. Those need platformdirs ≥ 4.2 and the floor in `pyproject.toml` is `>=3.10`; the lock has 4.11, so CI is covered, but the floor is arguably wrong for a different reason and I left it alone.
- **IDE-side `--no-write-history` tests.** Added three: flag typed, profile key, and the non-bool refusal.
- **5s freeze.** `UI_BUSY_TIMEOUT_MS = 250` for the IDE. With both writes now on workers this is belt-and-braces, but a missed row still beats a stalled worker.

## Nits

`0600` file under a `0700` directory (via `os.open` before `connect()`, as you suggested); `max(version, 0)` so a negative `user_version` can't half-migrate; the declared `connection_id` goes through `redact_conn_str()` before it reaches the column — guarded with `isinstance(str)`, because adapters are third-party code and a `MagicMock` promptly proved the point.

Also added a `threading.Lock` around the connection: two workers write it now, and `sqlite3` doesn't serialize that for us under `check_same_thread=False`.

## Comments and changelog

All of it — the seven benchmarks, the SQLite-vs-file / state-vs-cache / refusal-spelling decisions, "here rather than in `catalog_cache`", the "has always"/"used to" references, the no-op string after `__all__`, the stale `history = false`, and the sentence that appeared in three places. The module docstring is now your two sentences. `docs/plans/m4-hsql-technical-plan.md` keeps the rationale. Changelog is your wording; "as it runs" is now true for both commands.

## The one I didn't change

**The `harlequin_options` / `declared_by_adapter` asymmetry.** You're right that it's an accident rather than a decision, and that an adapter declaring `timeout` would split its own history. But I don't think either command should give way: AGENTS.md makes hsql's flags "the frozen part", so `timeout` in a profile has to be hsql's there; and the IDE's adapter options predate that rule, so making it drop `timeout` unconditionally would take a working option away from IDE users of such an adapter.

The principled fix is to stop hashing "whatever is left of the config" and hash **the adapter's declared options** instead — then it doesn't matter who pops what, and the two agree by construction. That changes every existing connection hash, so it belongs in PR 8, which bumps `CACHE_VERSION` to 3 anyway and gets the invalidation for free. Happy to do it here instead if you'd rather not carry the asymmetry through two releases.

## On your "not checked" list

Two I can close: `run_at` is `datetime.isoformat()` in UTC with a `+00:00` offset and a zero-padded fixed-width prefix, so lexical `order by run_at` matches chronological for every row this writes — but `--history` should order by `id` regardless, which is what the plan's §3.3 specifies. And `json.dumps` without `sort_keys` is fine as long as both commands build the dict in the same order, which they do (`{"conn_str": …, **config}` from a TOML-ordered profile) — but it's fragile, and folding it into the declared-options fix above would settle it properly. The rest I haven't verified either.

CI: the one red job is the `EditorCollection.on_mount` / `Tabs.add_tab` race from my earlier comment. It reproduced on Linux 3.10 locally during this work, which rules out "macOS only" — I ran the functional suite 3× with and 3× without these changes and got 143/143 and 141/141 both ways, so it isn't this PR's, and it now has a Linux repro for whoever picks it up.

---

tconbeer:
The last push turned two Windows jobs red, and **both were mine** — fixed in `2dd7109`.

**`_report_query_log_failure()` had lost its caller.** Moving the writes onto the workers left the notification orphaned, so a store the IDE could not write would have gone unmentioned — the exact thing the review asked to be surfaced. It is now called from the two handlers those workers post to, which are the ones allowed to touch the UI.

**`wait_for_workers` was never sound.** It read `app.workers` once, but a worker's result arrives as a *message* and its handler starts the next worker: `_execute_query` posts `QueriesExecuted`, whose handler starts `_fetch_data`. So the wait returned before the second worker existed. That was invisible while every worker was fast — and the query log's writes are enough to stop being fast on a slow disk. `test_history_screen` (Windows 3.12, both snapshots) and `test_toggle_full_screen` (Windows 3.10) are that. It now loops until the set is empty, giving the message pump a turn between passes, with a ceiling so a self-restarting worker cannot hang the suite.

I reproduced it rather than guessing: making every write sleep 20ms — about a thousand times a real disk — fails `test_history_screen` locally on the old wait, and the whole functional suite (143) passes on the new one. No assertion changed; the test waits for the work it always claimed to wait for.

For the record on the underlying cost, since two-phase writes were a review request: the fifteen statements that test runs cost **1.6ms** of query log in total, so this is a CI-runner amplification rather than something a user would feel. If you'd still rather the IDE not do sqlite I/O inside the query workers at all, the alternative is a queue drained by a writer thread — I did not do that, because a queue that has not flushed loses exactly the crash-safety the two-phase write exists for, and `hsql`'s `os._exit()` on timeout depends on every write being its own committed transaction.

Local: ruff, mypy, `lint-imports` (5 contracts), 1648 passed / 7 skipped.

The macOS `Tabs.add_tab` failure in the same run is still the base-branch one from earlier — no third comment on it, as promised.

---

tconbeer:
`2dd7109` fixed Windows 3.10 and 3.12, but Windows **3.11** then failed `test_toggle_full_screen` three runs out of three — deterministic on that runner, so `flaky`'s retries didn't hide it. Fixed in `337193f`.

**The mechanism.** The test runs a query and then focuses the Results Viewer without waiting for it:

```python
app.editor.text = "select 1"
await pilot.press("ctrl+j")

app.results_viewer.focus()      # <- nothing to focus yet
await pilot.press("f10")
assert app.editor_collection.disabled
```

The viewer takes focus as its table lands. Until then `focus()` doesn't stick, `self.focused` is still the editor, and `watch_full_screen` walks up from it and makes the **editor** full screen — so `editor_collection.disabled` is False and the assertion fails.

I did not just add a wait and hope. My first repro (slowing the query-log writes) did **not** reproduce this one, so I discarded it rather than claiming it. Slowing `fetch` to 300ms — which is what "the table lands late" actually means — reproduces it exactly: **3 failures out of 3 without the wait, 3 passes out of 3 with it**, matching the Windows 3/3 signature.

I also swept the other functional tests that submit a query. Every one either waits for the workers or polls for the visible table; the two unwaited `ctrl+j` presses in `test_data_catalog.py` are the catalog's insert binding, not a run. So this was the only test depending on a query finishing without waiting for it, and there is no similar fix pending elsewhere.

Local: ruff, mypy, `lint-imports`, 1648 passed / 7 skipped.

Worth saying plainly, since this is the third push chasing CI: these two test fixes (`wait_for_workers` looping through worker hand-offs, and this wait) are corrections to waits that were never sound. Both tests assert exactly what they asserted before. The query log made them fail by making the query path slower by ~1.6ms per fifteen statements, which was enough to move a race that was already there.

---

src/harlequin/app.py
```diff
@@ -1312,6 +1357,12 @@ def _cancel_query(self) -> None:
                 header="Harlequin could not cancel your queries.",
                 error=e,
             )
+        # whatever the fetch has not completed. A cancelled cursor comes back
+        # empty and error-free, so only this can tell those rows from a query
+        # that matched nothing.
+        while self.logged_rows:
```
tconbeer: **Ordering: the rows are marked after `connection.cancel()` returns, so a cancelled query can still be recorded as `ok, rows=0`.** This runs on the `query_cancellers` worker thread while `_fetch_data` runs on the `query_runners` one. Once `cancel()` returns, the fetch unblocks with an empty, error-free result, reaches its completion loop, and `pop(id_)`s the row as `ok` — and only then does this loop run and find nothing left. That is exactly the ambiguity `canceled` exists to remove, and it's the same order bug hsql avoids by setting `_expired` *before* `connection.cancel()` in `Deadline._cancel`.

Fix: take the pending rows atomically before cancelling — `pending, self.logged_rows = self.logged_rows, {}` — mark them `canceled`, then call `cancel()`; the fetch worker's later `pop(id_, None)` on the fresh dict is a no-op. That swap also removes the second problem here: `while self.logged_rows: … popitem()` is check-then-pop across two threads, so the fetch worker draining the last entry between the test and the `popitem()` raises `KeyError` on this worker.

`test_a_cancel_leaves_a_finished_query_alone` can't see either, because the query has finished before the cancel; a test that cancels while `fetch` is held open (the `slow_fetch` harness in the file above) and asserts `canceled` would.

---

src/harlequin/app.py
```diff
@@ -1403,9 +1454,21 @@ def _fetch_data(
...
         elapsed = time.monotonic() - submitted_at
+        for id_, result in results.items():
+            # taken off as it is completed, so that a later cancel marks only
+            # what was still running
+            self.query_log.update(
+                self.logged_rows.pop(id_, None),
+                rows=result.fetched_row_count,
+                truncated=result.truncated,
+                elapsed_ms=elapsed * 1000,
+            )
```
tconbeer: **Rows are completed in a second loop after *every* fetch has finished, so a cancel mid-batch marks statements that already returned as `canceled`.** For `select 1; select 2; select <slow>`, the first two are fetched and sitting in `results`, but still in `logged_rows`, when the user cancels the third — the canceller pops all three, and the completion loop here then finds nothing to complete, so the two that finished are recorded `canceled` with `rows` NULL. (With the swap-before-cancel fix above this gets more visible, not less.) The comment says "taken off as it is completed", but they're taken off as the *batch* completes.

Fix: pop and `update()` each row inside the fetch loop, right after its `fetch()` returns — and use `result.elapsed` for `elapsed_ms` rather than the batch clock. That is what hsql writes (`result.elapsed * 1000` in `_Run.record_result`), so the column would mean the same thing in both commands; today an IDE row carries the whole batch's time from submission. The batch `elapsed` stays what `ResultsFetched` reports.

---

src/harlequin/hsql/cli.py
```diff
@@ -1713,24 +1825,31 @@ def _fetched(
...
             run.failure = e
+            run.record_result(item.statement, status="error", error=e)
             diagnostics.report_error(e)
             if on_error == "stop":
                 return
             continue
         if run.stopped:
             # the cancel landed inside that fetch, so the rows it returned are
             # the ones it had rather than the ones the query has
+            run.record_result(item.statement, status="canceled")
```
tconbeer: The between-statements path above now marks everything after it, but the other two return paths — the cancel that raised out of `fetch()` (line ~1839) and this one, where it landed inside the fetch — still mark only `item`, so `selected[position:]` stay `ok`/NULL and are indistinguishable from `--result last`. `test_every_statement_a_cancel_stopped_says_so` puts the slow statement last, so it can't tell. A small helper that marks `selected[position - 1:]` used on all three returns closes it; putting the slow statement in the middle of that test pins it.

---

tests/unit_tests/test_query_log.py
```diff
@@ -0,0 +1,394 @@
...
```
tconbeer: Still `history = false` — this one didn't get changed in `8123a69` though the reply says it did. `"""`--no-write-history`: a disabled log is a real object whose writes do nothing."""`. Same family, not flagged before: `tests/conftest.py:89` "Kept on rather than disabled: …" is an alternative weighed; *"Point the query log at a throwaway store so a test run never writes the developer's own history."* is enough.

---

src/harlequin/query_log.py
```diff
@@ -0,0 +1,364 @@
...
+        for holding one.
+        """
+        store = self._path if self._path is not None else default_path()
+        db: sqlite3.Connection | None = None
+        try:
+            store.parent.mkdir(parents=True, exist_ok=True, mode=0o700)
+            # every statement this machine has run, so not world-readable.
+            # SQLite gives -wal and -shm the main file's mode. No-op on Windows.
+            os.close(os.open(store, os.O_CREAT | os.O_RDWR, 0o600))
+            # not thread-bound: hsql writes from the worker thread `--timeout`
```
tconbeer: Now stale: this comment says one connection "is only ever written by one thread at a time in either command", while the `__init__` comment on `_writing` (line 125) says the opposite and is the one that's true now. Keep just `# not thread-bound: the lock in _run() is what serializes writers` here, or drop it and let line 125 carry it.

Also `UI_BUSY_TIMEOUT_MS` (line 79): "a front end that would otherwise stop redrawing" no longer describes anything, since both IDE writes are on workers now; *"What a front end waits, so a worker is not held on a lock another process owns."* is what it does.

---

tconbeer (commented):
Follow-up review at `337193f` (the three response commits, `8123a69`..`337193f`, plus the merge of `main` in `1f08878`). Scope this round: the response diff only, under correctness and comment hygiene, with the merge checked separately.

**Verdict: close, one more round.** All four blocking findings from the first review are addressed and verified, but the IDE's new two-phase path has two ordering bugs of the same kind the first round found in hsql, so a cancelled query can still be recorded as `ok`.

**Verified fixed**

- `-c`/`-f` no longer hashed; both new keying tests are real (I confirmed `connection_id` is `None` under the monkeypatch, so the hash actually runs).
- IDE writes on `_execute_query`, completes on `_fetch_data`; the mid-fetch test genuinely holds the fetch open and sees the row with `rows IS NULL`.
- `redact_sql()` masks unregistered URI passwords, `create secret (… key_id, secret)` and `set …_secret_… = '…'`; probed with escaped quotes, `tokens where name = 'alice'`, repeated literals and duplicate matches — no under-redaction, and the double-mask quote-eating is fixed. `secret_name = '…'` is over-redacted, which is the right direction.
- Execute-phase cancel is recorded via the `recorded()` generator; sqlite test pins it.
- `clean_env` shared by all subprocess tests, `WIN_PD_OVERRIDE_*` set; IDE flag/profile/non-bool tests added; `0600`/`0700`; `max(version, 0)`; declared id through `redact_conn_str`; write lock; `UI_BUSY_TIMEOUT_MS`; changelog wording; module docstring and the removed/history references.
- **Merge:** relative to `main`, the three conflicted files carry only this PR's additions and no `main`-only file was touched.
- Locally on this head: ruff, mypy strict, `lint-imports` (5 kept), and the full suite — **1648 passed, 7 skipped**.

**Open (inline)**

1. `_cancel_query` marks rows *after* `connection.cancel()` returns, on a different thread from the fetch — the fetch completes the row as `ok, rows=0` first. Swap the dict out and mark before cancelling (as hsql's `Deadline._cancel` sets its flag before cancelling). The `while … popitem()` across threads can also raise `KeyError`.
2. `_fetch_data` completes rows in a second loop after the whole batch, so a mid-batch cancel marks already-fetched statements `canceled`. Complete each row inside the fetch loop, and use `result.elapsed` so `elapsed_ms` means the same thing as hsql's.
3. `_fetched` still marks only the in-flight statement on two of its three cancel returns.
4. Two stale comments (`query_log.py:227` contradicts line 125; the `history = false` docstring wasn't actually changed) and one docstring whose rationale no longer applies (`UI_BUSY_TIMEOUT_MS`).

**Accepted as decided:** deferring the `harlequin_options`/`declared_by_adapter` asymmetry to PR 8's declared-options hash and `CACHE_VERSION` bump is a reasonable call, and it's now a decision rather than an accident. The `wait_for_workers` and `test_toggle_full_screen` waits are test-infra fixes slightly outside the PR's scope; the reproductions described justify them, and neither assertion moved.

## Not checked
- I did not re-run the three security/correctness/comment sub-passes on the response diff; this round is my own read plus targeted probes (redaction, cancel ordering by code inspection, merge diff).
- I did not reproduce the cancel-ordering race empirically; the finding is from the thread assignment of the two workers and the order of the two calls.
- I did not run on 3.12–3.14, macOS or Windows; the `WIN_PD_OVERRIDE_*` behaviour is from the platformdirs source, not a Windows run.
- I did not verify the macOS `Tabs.add_tab` failure is red on `main`; I'm taking the linked base-branch run at its word.
- Unchanged from round one: the History screen / `hsql --history` read path, uninstalled adapters' `connection_id`, and the WAL fallback on a real network filesystem.

---

tconbeer:
All four addressed in `71c80b3`. Both ordering findings were real, and you were right that they're the same shape as the hsql one from round one — I moved the write into the workers and reproduced the bug I'd just fixed.

**1. `_cancel_query` marked after `cancel()` returned.** Fixed as suggested: the rows are taken and marked before the cancel, and the dict is swapped in one step (`pending, self.logged_rows = self.logged_rows, {}`) rather than drained with `popitem()`, which was check-then-pop across two threads.

My first attempt at the test was worthless — it passed with the bug restored, because releasing the held fetch *after* `_cancel_query()` never puts the completion in the window. The test now patches `connection.cancel` with a fake that releases the fetch and waits for its row to complete, which puts the completion *inside* `cancel()` deterministically instead of racing for it. Verified: `ok` with the old order, `canceled` with the new.

**2. `_fetch_data` completed rows after the whole batch.** Each row is now completed inside the fetch loop, right after its `fetch()` returns — and with `result.elapsed` rather than the batch clock, so `elapsed_ms` means the same thing as hsql's. The batch `elapsed` stays what `ResultsFetched` reports. The stale "taken off as it is completed" comment goes with it.

**3. `_fetched`'s other two cancel returns.** One `canceled_from(position)` helper on all three, marking the in-flight statement and everything after it. `test_every_statement_a_cancel_stopped_says_so` now puts the slow statement in the *middle* — with only the in-flight one marked it fails at index 2 (`ok` where `canceled` is expected), which is the `--result last` ambiguity you named.

**4. Stale comments.** The `check_same_thread` one contradicted the lock above it and is now just the pointer to it; `UI_BUSY_TIMEOUT_MS` describes what it does rather than a freeze that can't happen now both writes are on workers; `test_a_disabled_log_writes_nothing` gets your wording — you were right that it didn't change in `8123a69` despite my reply saying it had; and `tests/conftest.py`'s fixture no longer weighs an alternative.

Local: ruff, mypy strict, `lint-imports` (5 kept), **1649 passed / 7 skipped**.

On your not-checked list, one I can close: the macOS `Tabs.add_tab` failure is red on `main` at `e278c0b` in [run 33833836899](https://github.com/tconbeer/harlequin/actions/runs/33833836899) — macOS 3.10 and 3.11 there, while 3.14 passed; on this branch it has since landed on 3.11, 3.12 and 3.14 across four runs. It also reproduced on Linux 3.10 locally under the full suite, so it isn't macOS-specific either. Not taking my own word for it: that run is `main`'s own push build.

---

tconbeer:
Windows 3.14 went red on `71c80b3`, and it was one of the tests I added last round. Fixed in `aa1d048`.

```
FAILED test_a_statement_cancelled_while_it_runs_is_still_recorded
  - worker 'gw3' crashed while running it
```

A *crash*, not an assertion. The test ran `hsql --timeout 0.3` **in process**, and a run that will not stop within the grace period ends in `os._exit()` — which in process takes the xdist worker with it. SQLite happened to stop inside the grace period on every other runner, so the hazard only surfaced on the one slow enough to miss it. `timeout.py`'s docstring says exactly this, and the timeout tests already in the file are subprocesses for exactly this reason; I didn't follow them.

Both timeout tests this branch added now run through a child and read the store back out of the temporary state directory `clean_env` already points them at — found with `tmp_path.rglob("history.db")` rather than by reimplementing platformdirs' per-platform layout. Assertions unchanged, and `test_every_statement_a_cancel_stopped_says_so` still fails when `_fetched` marks only the in-flight statement.

I checked the rest of the file for the same hazard while I was there: every other in-process `--timeout` test uses the fake `stuck` adapter, which cancels cleanly, or a timeout long enough never to fire. The two that drive a real driver to expiry — `test_a_run_that_outlasts_the_grace_period_still_exits_4` and `test_a_real_duckdb_query_times_out_and_says_so` — were already subprocesses. So these two were the only ones, and there is no third waiting to fire.

Local, under xdist as CI runs it: ruff, mypy strict, **1649 passed / 7 skipped**.

---

Issue #745, opened 2025-02-06
Fix crash caused by DuckDB connection error


tconbeer:
When opening the same db in two terminals:

```bash
~/open/harlequin fix/infinity-timestamps ≢  ~3  3.9 ❯ make serve
textual run --dev -c harlequin -P dev -f . f1.db
╭─ DuckDB couldn't connect to your database. ──────────────────────────────────────────────────────────────────────────╮
│ IO Error: Could not set lock on file "/home/tco/open/harlequin/f1.db": Conflicting lock is held in                   │
│ /home/tco/.local/share/uv/python/cpython-3.9.20-linux-x86_64-gnu/bin/python3.9 (PID 26697). See also                 │
│ https://duckdb.org/docs/connect/concurrency                                                                          │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭───────────────────────────────────────── Traceback (most recent call last) ──────────────────────────────────────────╮
│ /home/tco/open/harlequin/src/harlequin/components/code_editor.py:224 in on_mount                                     │
│                                                                                                                      │
│   221 │   async def on_mount(self) -> None:                                                                          │
│   222 │   │   if self.startup_cache is not None:                                                                     │
│   223 │   │   │   for _i, buffer in enumerate(self.startup_cache.buffers):                                           │
│ ❱ 224 │   │   │   │   await self.action_new_buffer(state=buffer)                                                     │
│   225 │   │   │   │   # we can't load the focus state here, since Tabs                                               │
│   226 │   │   │   │   # really wants to activate the first tab when it's                                             │
│   227 │   │   │   │   # mounted                                                                                      │
│                                                                                                                      │
│ ╭───────────────────────────────────────────────── locals ─────────────────────────────────────────────────╮         │
│ │     _i = 0                                                                                               │         │
│ │ buffer = BufferState(                                                                                    │         │
│ │          │   selection=Selection(start=(3, 39), end=(3, 39)),                                            │         │
│ │          │   text="-- select current_date, current_timestamp::timestamp, '2025-02-05 22:07:42.90456"+175 │         │
│ │          )                                                                                               │         │
│ │   self = EditorCollection()                                                                              │         │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────╯         │
│                                                                                                                      │
│ /home/tco/open/harlequin/src/harlequin/components/code_editor.py:275 in action_new_buffer                            │
│                                                                                                                      │
[...]

PR #1136, opened 2026-09-04
Crash recovery and bug reporting

### LINKED ISSUE #687: Never crash from an uncaught exception; gracefully exit (save buffers) and show error message with instructions to open an issue
this probably means writing a log file and pointing users to it.

Include the trace and other info from the env, if possible.

would be nice to do this inside the app so we can try to preserve the contents of the buffer (or maybe we should just checkpoint that every minute or so anyway)

tconbeer:
Closes #687

Implements `docs/plans/crash-recovery-design.md`, as its seven ordered commits. Each is independently green under `make check`.

1. **`write_cache` is atomic and never raises.** Both writers go through `_write_pickle()`: a pid-named temp file in the same directory, `fsync`, then `os.replace`. It returns a bool, because the two new callers write from places where an exception would itself take the app down.
2. **A per-process recovery file, and rename-before-replay.** Checkpoints go to `recovery-<pid>.pickle`, never to `cache-1.pickle`, so two Harlequins keep today's "last clean quit wins" semantics. `adopt_recovery()` renames the file it picks to `.replayed` *before* unpickling it — that one line is the whole crash-loop guard. It adopts a crash handler's `recovered-*` always, and a `recovery-*` only once its mtime has gone stale, which is how a live sibling keeps its own.
3. **A 60s checkpoint timer**, so a `kill -9`, a closed terminal or a dropped ssh session no longer loses every open buffer. Skipped when the content hasn't changed, and skipped when every buffer is blank.
4. **`environment.py`**, the facts a diagnostic reports, shared by `hsql --info` and the crash report. `--info`'s document is byte-identical; `terminal_facts()` and `install_facts()` are new and outside it.
5. **The handler.** `AppBase._handle_exception` saves the user's work, writes a crash report, and prints the same panel every other Harlequin error uses. `cli.py` forwards the app's return code.
6. **`hsql` exits 70** (`sysexits.h`'s `EX_SOFTWARE`) and writes a report, instead of letting Python print a traceback and exit 1 — which is the code for SQL the database rejected.
7. **`.github/ISSUE_TEMPLATE/crash_report.md`**, which the message and `crash.ISSUE_URL` point at.

Beyond the design: the exit code is also added to hsql's generated CLI reference and to the skill's exit-code list and troubleshooting reference, which are agent-facing contracts that would otherwise be wrong.

### What a user sees

A real crash, provoked by raising `KeyError("column_index")` from a key handler with a `warehouse` profile whose password is `hunter2-and-more`:

```
╭─ Harlequin crashed. ─────────────────────────────────────────────────────────────────────────────╮
│ KeyError: 'column_index'                                                                         │
│                                                                                                  │
│ Your buffers have been saved, and Harlequin will offer them back the next time you start it.     │
│                                                                                                  │
│ Please report this crash to help improve Harlequin.                                              │
│ https://github.com/tconbeer/harlequin/issues/new?template=crash_report.md                        │
│                                                                                                  │
│ The file you will need for the crash report is at:                                               │
│ /home/tco/.local/state/harlequin/log/crash-20260904T123337Z-1746.log                             │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
```

The URL is an OSC-8 hyperlink, so it is clickable in a terminal that supports one and still readable as a URL in one that doesn't. The rest of the message is escaped before rendering, so a driver that puts `[brackets]` in an exception doesn't lose them to rich's markup parser. `echo $?` is `1`.

And the report at that path, in full — this is what pastes into the template's fenced block:

```
harlequin crash report, written 2026-09-04 12:33:53Z.

Review and redact as necessary before sharing this file: it holds your
configuration, with passwords masked, and the SQL that was in your
active buffer.
Report this at https://github.com/tconbeer/harlequin/issues/new?template=crash_report.md


========================================================================
=== ENVIRONMENT ===
========================================================================

version: 2.13.0
python:
  version: 3.10.15
  implementation: CPython
  executable: /home/tco/.local/share/uv/tools/harlequin/bin/python3
platform:
  system: Linux
  release: 6.18.44
  machine: x86_64
terminal:
  term: xterm-256color
  term_program: WezTerm
  term_program_version: 20240203-110809
  colorterm: truecolor
  shell: /bin/bash
  wsl_distro: None
  ssh: False
  locale: en_US.UTF-8
  size: 120x36
install:
  installer: uv
  in_venv: True


========================================================================
=== CONTEXT ===
========================================================================

adapter: duckdb
adapter_class: DuckDbAdapter
adapter_module: harlequin_duckdb.adapter
adapter_distribution: harlequin 2.13.0
profile: warehouse
keymaps: vscode
theme: harlequin
connected: True
buffers: 1
size: Size(width=120, height=36)
recovery_file: /home/tco/.cache/harlequin/recovered-20260904T123337Z-1746.pickle


========================================================================
=== TRACEBACK ===
========================================================================

Traceback (most recent call last):
  File "/home/tco/.local/share/uv/tools/harlequin/lib/python3.10/site-packages/harlequin/app.py", line 19, in go
    raise KeyError("column_index")
KeyError: 'column_index'


========================================================================
=== SQL IN THE ACTIVE BUFFER ===
========================================================================

select customer_id, sum(amount)
from orders
group by 1
```

Nothing in it names `hunter2-and-more`, and there are no locals. (The paths and terminal above are edited to read like a user's machine; everything else is verbatim from a real run in CI's container.)

hsql's is the same file under a different program name, and its stderr is three plain lines rather than a panel — `hsql` writes no box drawing, and no escape sequences either, so no link markup there:

```
hsql: error: hsql hit a bug in itself and could not finish the run.
note: please report this crash to help improve Harlequin: https://github.com/tconbeer/harlequin/issues/new?template=crash_report.md
note: the file you will need for the crash report is at /home/tco/.local/state/harlequin/log/crash-20260904T123400Z-1815.log
```

`echo $?` is `70`. Its report has no buffer section (hsql has no buffers) and a `CONTEXT` of the argv, adapter and profile, with a DSN typed on the command line masked in place:

```
argv: postgres://tco:********@warehouse:5432/analytics -c select 1
adapter: None
profile: None
```

The design doc has the full reasoning. The three decisions worth repeating here:

- **A per-process recovery file rather than checkpointing into the shared cache.** A shared file would be thrashed every minute by every running Harlequin, and a checkpoint could overwrite what another process wrote when it quit cleanly. The separate file also means the last clean-quit state stays on disk: if the recovered buffers are themselves what crashed, `cache-1.pickle` is untouched.
- **Rename before replay, rather than a session sentinel or a PID-liveness check.** #745 crashed *inside* `EditorCollection.on_mount` while replaying cached buffers, so saving buffers on crash without a guard would turn a one-time crash into a crash loop — worse than what we have now. Renaming first is one line, needs no state, and is correct across concurrent processes by construction. Staleness by mtime does the same job for the liveness question, which a PID check would get wrong on Windows and under PID reuse.
- **The message is a `panic()` panel on stderr, not a modal.** `panic()` calls `_close_messages_no_wait()`, so the pump is gone by the time we handle the exception; an in-app modal is unreachable, and attempting one re-enters an app that has already proven it is broken.

Tradeoffs, all deliberate and covered in the doc's "Known risks":

- **The active buffer's SQL is in the report.** It is usually what is needed to reproduce a TUI bug, and the file is local — but SQL holds table names and literals, so it goes last, under an explicit heading, and the report's own header asks the reporter to review and redact as necessary. The whole report goes through `redact.redact_text()` once at the end, and there is no `show_locals`.
- **Both commands now exit non-zero where they exited 0.** `harlequin` returns 1 on a crash and 2 on a config or connection error (which previously exited 0 — a live bug this closes); `hsql` returns 70 where Python let it exit 1. User-visible, so both have a changelog line.
- **Partial-replay convergence.** A crash partway through replaying recovered buffers can write a `recovered-*` file holding the subset that loaded. It converges and salvages what it can, but may take more than one restart.

This leans on five Textual internals (`textual==8.2.8`, pinned exactly). Each use site has a comment naming what it is for, and each has a test that fails loudly rather than silently reverting us to raw tracebacks — in particular the crash tests are written as `with pytest.raises(RuntimeError)`, because `run_test` re-raises from `_exception`/`_exception_event` and a handler that did not set them would turn every crashing functional test silently green.

Two places where the implementation departs from the doc, both in the tests:

- The doc's functional test asserts on `_exit_renderables[0]`, but `_print_error_renderables()` clears that list, so there is nothing to index by the time the test body resumes. The assertions are on captured stderr instead, which is what the user actually sees.
- Injecting through `_handle_exception` has to happen from inside an `except` block: Textual's own renderer reads `sys.exc_info()`, so an exception injected from outside one exercises a path no real crash takes.

- [ ] No.
- [x] Yes; I haven't opened a PR, but the gist of the change is: the troubleshooting page should say that a crash writes a report, where the reports live per-OS, and to file it against the new Crash Report template; and `harlequin`'s and `hsql`'s exit codes are now documented as non-zero on failure (`harlequin`: 1 on a crash, 2 on a config or connection error; `hsql`: a new `70`).

- [x] Yes.

New: `tests/unit_tests/test_editor_cache.py`, `test_environment.py`, `test_crash.py`, and `tests/functional_tests/test_crash.py`; plus recovery and checkpoint cases in `test_cache.py` and the hsql boundary cases in `test_hsql.py`.

Two pieces of test infrastructure worth calling out, because without them the suite would write into the developer's own directories: an autouse, session-wide `crash_reports_go_to_tmp` fixture (every failing functional test goes through `run_test`'s re-raise path, and so through `_handle_exception`), and `no_use_buffer_cache` extended to patch the new seams. `test_cli.py`'s and `test_ssh.py`'s `MagicMock(spec=Harlequin)` now set `return_code = 0`, since the command exits with it and a MagicMock is not an exit code.

Verified with `make check` — ruff, mypy (140 files), the full 3.10 suite (1548 passed, 7 skipped, 148 snapshots) and the 3.12 `py12` run — plus `lint-imports` (4/4 contracts kept, with `harlequin.crash` and `harlequin.environment` added to the headless contract and to `HEADLESS_IMPORTS`). Also ran the design's by-hand checks, which is where the examples above come from: an injected IDE crash writes `recovered-*.pickle` and a crash log, whose buffer the next start restores, leaving the file `.replayed`.

https://claude.ai/code/session_0131Sdv9zwvjBkuqWxE1bqHG

### DIFF
--- a/.github/ISSUE_TEMPLATE/crash_report.md
+++ b/.github/ISSUE_TEMPLATE/crash_report.md
@@ -1,0 +1,49 @@
+---
+name: Crash Report
+about: Harlequin or hsql exited unexpectedly and wrote a crash report.
+title: 'Crash: '
+labels: 'bug'
+assignees: ''
+
+---
+**Before proceeding, please acknowledge**:
+- [ ] I have searched Issues and Discussions in this repo for this error.
+
+**Paste your crash report below.**
+
+Harlequin and `hsql` write a crash report and print its path when they exit
+unexpectedly. If you still have that message on screen, the path is in it.
+Otherwise, the reports are here:
+
+| OS | Location |
+|---|---|
+| macOS | `~/Library/Logs/harlequin/` |
+| Linux | `~/.local/state/harlequin/log/` |
+| Windows | `%LOCALAPPDATA%\harlequin\Logs\` |
+
+The newest `crash-*.log` is the one you want.
+
+> [!IMPORTANT]
+> Please read the report before pasting it. It contains your configuration
+> (with passwords masked) and the SQL that was in your active buffer.
+
+<details>
+<summary>Crash report</summary>
+
+```
+PASTE YOUR CRASH REPORT HERE
+```
+
+</details>
+
+**What were you doing when it crashed?**
+Steps to reproduce it, if you know them.
+
+**Anything else?**
+Screenshots, or anything the report doesn't cover.
+
+**Contributing**
+Are you interested in contributing a fix?
+- [ ] Yes
+- [ ] Maybe
+- [ ] No
--- a/.github/ISSUE_TEMPLATE/crash_report.md
+++ b/.github/ISSUE_TEMPLATE/crash_report.md
@@ -8,6 +8,7 @@ assignees: ''
 ---
 **Before proceeding, please acknowledge**:
 - [ ] I have searched Issues and Discussions in this repo for this error.
+- [ ] I have redacted secrets and PII from the crash report pasted below.
 
 **Paste your crash report below.**
 
--- a/src/harlequin/app.py
+++ b/src/harlequin/app.py
@@ -77,7 +77,12 @@ from harlequin.config import (
 )
 from harlequin.copy_formats import HARLEQUIN_COPY_FORMATS, WINDOWS_COPY_FORMATS
 from harlequin.driver import HarlequinDriver
-from harlequin.editor_cache import Cache, clear_recovery
+from harlequin.editor_cache import (
+    CHECKPOINT_INTERVAL_SECONDS,
+    Cache,
+    clear_recovery,
+    write_recovery,
+)
 from harlequin.editor_cache import write_cache as write_editor_cache
 from harlequin.exception import (
     HarlequinBindingError,
@@ -305,6 +310,8 @@ class Harlequin(AppBase):
                 ),
             )
         self.query_timer: Union[float, None] = None
+        self._last_checkpointed_cache: Cache | None = None
+        """What the recovery file holds, so an idle session stops rewriting it."""
         self.connection: HarlequinConnection | None = None
         self._recovery_lock = threading.Lock()
         """Held across reopening the tunnel and the connection through it.
@@ -421,6 +428,7 @@ class Harlequin(AppBase):
         self._connect()
         self._load_catalog_cache()
         self.action_bind_keymaps(*self.keymap_names)
+        self.set_interval(CHECKPOINT_INTERVAL_SECONDS, self._checkpoint_editor_cache)
 
     @on(Button.Pressed, "#run_query")
     def submit_query_from_run_query_bar(self, message: Button.Pressed) -> None:
@@ -1122,6 +1130,38 @@ class Harlequin(AppBase):
 
     def action_focus_results_viewer(self) -> None:
         self.results_viewer.focus()
+
+    def _build_editor_cache(self) -> Cache | None:
+        """The open buffers, or None when there is nothing worth writing.
+
+        Blank buffers are nothing: a crash before the editor mounts leaves the
+        collection empty, and writing that over a recovery file would destroy
+        exactly the work this exists to save.
+        """
+        editor_collection = getattr(self, "editor_collection", None)
+        if editor_collection is None or not editor_collection.is_mounted:
+            return None
+        cache = Cache(
+            focus_index=editor_collection.active_buffer_index,
+            buffers=editor_collection.buffers,
+        )
+        if not any(buffer.text.strip() for buffer in cache.buffers):
+            return None
+        return cache
+
+    def _checkpoint_editor_cache(self) -> None:
+        """Write the open buffers to this session's recovery file, if they moved.
+
+        Synchronously, on the message pump: pickling a few SQL buffers is
+        microseconds, and a thread worker would only add the hazard of two
+        checkpoints racing. `Cache` compares by content, so no dirty flag has
+        to stay in sync with tab swaps.
+        """
+        cache = self._build_editor_cache()
+        if cache is None or cache == self._last_checkpointed_cache:
+            return
+        if write_recovery(cache):
+            self._last_checkpointed_cache = cache
 
     async def action_quit(self) -> None:
         write_editor_cache(
--- a/src/harlequin/app.py
+++ b/src/harlequin/app.py
@@ -77,7 +77,7 @@ from harlequin.config import (
 )
 from harlequin.copy_formats import HARLEQUIN_COPY_FORMATS, WINDOWS_COPY_FORMATS
 from harlequin.driver import HarlequinDriver
-from harlequin.editor_cache import Cache
+from harlequin.editor_cache import Cache, clear_recovery
 from harlequin.editor_cache import write_cache as write_editor_cache
 from harlequin.exception import (
     HarlequinBindingError,
@@ -1130,6 +1130,7 @@ class Harlequin(AppBase):
                 buffers=self.editor_collection.buffers,
             )
         )
+        clear_recovery()
         update_catalog_cache(
             connection_hash=self.connection_hash,
             catalog=None,  # TODO: cache completions instead.
--- a/src/harlequin/app.py
+++ b/src/harlequin/app.py
@@ -6,10 +6,12 @@ import os
 import sys
 import threading
 import time
+from datetime import datetime, timezone
 from functools import partial
 from pathlib import Path
 from typing import (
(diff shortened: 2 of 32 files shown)


.github/ISSUE_TEMPLATE/crash_report.md
```diff
@@ -0,0 +1,49 @@
...
+name: Crash Report
+about: Harlequin or hsql exited unexpectedly and wrote a crash report.
+title: 'Crash: '
+labels: 'bug'
+assignees: ''
+
+---
+**Before proceeding, please acknowledge**:
+- [ ] I have searched Issues and Discussions in this repo for this error.
+
```
tconbeer, line 12: Also: I have redacted secrets and PII from the crash logs pasted below

src/harlequin/app.py
```diff
@@ -1123,13 +1148,110 @@ def action_focus_query_editor(self) -> None:
...
         self.results_viewer.focus()
 
+    def _build_editor_cache(self) -> Cache | None:
+        """The open buffers, or None when there is nothing worth writing.
+
+        Blank buffers are nothing: a crash before the editor mounts leaves the
+        collection empty, and writing that over a recovery file would destroy
+        exactly the work this exists to save.
+        """
+        editor_collection = getattr(self, "editor_collection", None)
```
tconbeer, line 1168: Why not just self.editor_collection?

tconbeer (commented):
Add example crash message and an example report to the PR body please

.github/ISSUE_TEMPLATE/crash_report.md
tconbeer: Added in 286335c, as a third acknowledgement:

```markdown
- [ ] I have redacted secrets and PII from the crash report pasted below.
```

Worth having even though the report masks what it can: the masking covers registered secrets and the credentials inside a DSN, and nothing else — a table name, a literal in the active buffer's SQL, or a hostname in a traceback are all things only the reporter can judge.

---

src/harlequin/app.py
tconbeer: Because `self.editor_collection` raises rather than being falsy — `compose()` assigns it, so before the app mounts the attribute does not exist at all:

```
>>> app = Harlequin(DuckDbAdapter((":memory:",), no_init=True))
>>> app.editor_collection
AttributeError: 'Harlequin' object has no attribute 'editor_collection'
```

That is not a hypothetical here: `_save_work_on_crash()` calls this, and a crash during startup — before or during `compose()` — is one of the crashes the handler runs for. #745 was exactly that shape. `is_mounted` alone doesn't cover it, since there's no object to ask.

The `except Exception` in `_save_work_on_crash` would swallow the `AttributeError`, so a plain attribute access wouldn't *break* anything — but it would mean the pre-mount case is handled by accident rather than on purpose, and the same helper is called every 60s by the checkpoint timer, where I'd rather not rely on exception flow.

I've kept the `getattr` and split the docstring in 286335c so the two guards read as the separate things they are — the attribute may not exist yet, and blank buffers are not worth writing:

```python
def _build_editor_cache(self) -> Cache | None:
    """The open buffers, or None when there is nothing worth writing.

    `editor_collection` is assigned in `compose()`, so before the app
    mounts the attribute does not exist at all -- and a crash before that
    is one of the crashes the caller runs for.

    Blank buffers are nothing either: writing them over a recovery file
    would destroy exactly the work this exists to save.
    """
```

`test_a_crash_before_the_app_is_mounted_saves_nothing` pins it. Happy to switch to `self.editor_collection: EditorCollection | None = None` in `__init__` instead if you'd prefer the attribute always exist, but that makes every other use site (about ten, all post-mount) answer to mypy for a None that can't happen there.

---

PR #1138, opened 2026-09-04
Warm sessions 2/6: the server, the flag partition, and byte-equivalence

tconbeer:
No issue; this is PR 2 of the six in [`docs/plans/hsql-warm-sessions-design.md`](https://github.com/tconbeer/harlequin/blob/main/docs/plans/hsql-warm-sessions-design.md) §9, on top of #1132.

`hsql --serve NAME [CONN_STR]` runs in the foreground, connects once, and answers `hsql --session NAME ...` invocations sent to it over an `AF_UNIX` socket — so a caller pays neither the imports nor the connection again. `select 1` against a warm DuckDB session answers in ~25ms against ~350ms cold, and against a server-backed adapter the whole of `connect()` is what stops being paid per invocation.

- **`harlequin/hsql/server.py`** — the accept loop, one worker at a time, the socket lifecycle, the uid check. A request is run by `cli.run(argv, served=...)`, the same command the cold path builds, writing into a recorder that stands in for the process's streams; what goes back is those bytes and the exit code. The server formats nothing and parses nothing of its own.
- **`--session`, `--session-reset` and `--queue-timeout`** become click options beside `--serve`. `--session-reset` closes the connection and opens a fresh one without restarting the process — the escape hatch for a session left in a strange state, and what gives each equivalence test a fresh one.
- **The §4.1 flag partition, on both sides.** Every option is in exactly one of four groups (connection-time, per-request, server-lifetime, role), pinned by a test that checks the four partition the command's own parameters. `--serve` refuses per-request options; a served request refuses `--queue-timeout`. Each refusal names the invocation the option belongs on:
  ```
  hsql: error: --command is a per-request option, and --serve takes none. Start the
  session, then send it the request: 'hsql --serve prod ...', then 'hsql --session
  prod --command ...'.
  ```
- **`session` and `serve` join `CLI_ONLY_SESSION_KEYS`**, refused from a config file the way `ssh_allow_reuse` is, and dropped from the generated schema. `session` is read before any config file is, so a profile setting it would name a session the invocation was never sent to; a profile answering `serve` would turn a query into a daemon.
- **Cancellation that doesn't land.** The cold path ends the process when cancelled work outlasts its grace period, because a thread still inside a driver aborts an interpreter exiting around it. A server can't do that, so `Deadline` takes an `abandon` hook: the request exits 4, and the session refuses to hand its connection to anyone until a reset.

POSIX only — native Windows has no `AF_UNIX`, which the client from #1132 already handles. WSL2 is Linux and gets it.

The design doc argues the whole shape (§3 rejects more import optimization, a batch mode, MCP-instead, and a connection pool); what this PR decided on top of it:

- **The server runs `cli.run()` rather than its own path.** A second execution path would be a second product, and §8 names that as the risk this feature carries. The mitigation is structural: the server has no formatting and no flag handling, so it *can't* drift. Byte-equivalence is the test that fails if it ever does.
- **One request at a time, via a turnstile rather than a lock.** Two threads on one DuckDB connection fail outright, and the adapter contract promises nothing about thread-safety — the IDE has always run exclusive workers. A lock would wake waiters in whatever order the interpreter chose; a turnstile hands out tickets so the client that waited longest goes next, and so a waiter can give up on `--queue-timeout` and be told it *never reached the database* — a different fact from a query that ran too long, and one a caller retries differently.
- **A `flock` beside the socket.** Distinguishing a live session from a stale socket file by connecting races another server starting under the same name. The lock makes "am I the session called `prod`" atomic, and the second server is refused rather than quietly stealing the name.
- **A served request refuses connection options outright, rather than comparing them.** §4.4's differing-options rejection needs the recorded connection identity, which is PR 3. Until then the choice is between ignoring them — which §4.4 rejects, because a client that typed `-a postgres` and got DuckDB has been lied to — and refusing all of them. Refusing is the conservative half and is forward-compatible: PR 3 relaxes it to "refuse only if it differs," so an identical `-a duckdb` starts being served.
- **The signal handler is installed before the socket listens.** Found by a test: `listen()` makes the socket connectable, so a stop signal arriving before the handler was installed killed a process a client could already reach, exiting 130 on what should be a clean stop.
- **Two equivalence parametrizations, not one.** The claim is "given the same starting state, one invocation produces the same bytes cold or warm" — but a *cold* invocation carries the connection on its command line and a warm one does not, and appending `--no-init :memory:` to `hsql --spec` changes what that invocation means. So invocations that connect compare `hsql CONN ARGS` against `hsql --session NAME ARGS`, and the modes that read no database compare identical argv both ways. Splitting them is what keeps the assertion honest rather than exempting the cases that wouldn't pass.

Tradeoffs worth naming: the server swaps the process's cwd, forwarded environment and streams for the length of each request, which is only safe *because* requests are serialized — that invariant is load-bearing beyond concurrency. And a session is stateful in ways a cold invocation is not (§5), which is a feature and a sharp edge; the `session_divergent` marker below is the enumerated version of it.

- [ ] No.
- [x] Yes; I haven't opened a PR, but the gist of the change is: the "Headless & Agents" topic needs a session section — written per §5.1 as *"a database session you can send commands to,"* not *"hsql, but faster,"* since a reader holding the first model predicts the temp-table behavior correctly and a reader holding the second files a bug — plus a `SessionStart`-hook example and the note that `--idle-timeout 0` is the answer to "my session died between calls." That is PR 6 of the plan, which is docs-only and deliberately not last in spirit.

- [x] Yes.

Three kinds, per §6:

1. **`tests/unit_tests/test_hsql_equivalence.py`** — every single invocation run twice, cold and warm, asserting identical stdout bytes, identical stderr, identical exit code, and identical files for the `-o` cases. 51 cases covering the formats, NULLs, truncation, the row cap, multi-result selection, `--on-error`, `--stats`, color, the catalog modes, and the usage errors. Each gets a fresh session via `--session-reset`, so the suite is the primary consumer of the escape hatch §5 proposes for humans.
2. **`@pytest.mark.session_divergent`** — registered beside `online` and `py12`. A marked test must assert what the warm behavior *is*, not merely that it differs: a temp table is `does not exist` cold and a result set warm, and both are in the assertions.
3. **`tests/unit_tests/test_hsql_serve.py`** — the partition (including the test that the four groups partition the command's parameters), the turnstile's ordering and its deadline, the recorder's stream interleaving, the socket lifecycle (stale socket, a second server, an unsafe runtime dir, clean shutdown), and the wire from a client that is not ours (chunked stdout, the `-f -`-with-no-stdin refusal, a cwd that can't be entered, `--color auto` reading the caller's terminal).

Full suite: 1542 unit passed, 139 functional passed. mypy clean over 142 files, all five import-linter contracts kept, ruff clean. The client's cold-start budget is unchanged at 58 modules — the client never imports the server. `docs/generated/hsql-reference.md` and `src/harlequin/schemas/config-v1.json` regenerated.

https://claude.ai/code/session_01SCbCzJWyU8yqSiUeDTxvzb

---
_Generated by [Claude Code](https://claude.ai/code/session_01SCbCzJWyU8yqSiUeDTxvzb)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
-- hsql can now hold a database connection open between invocations: `hsql --serve NAME [CONN_STR] [connection options]` starts a session in the foreground, and `hsql --session NAME -c ...` (or `HSQL_SESSION=NAME`) sends queries to it, paying neither the start-up nor the connection cost again. A session keeps its temp tables, settings and open transactions between invocations; `hsql --session NAME --session-reset` reconnects it. Not available on native Windows.
+- `hsql --serve NAME [CONN_STR]` holds a database connection open as a named session, and `hsql --session NAME -c ...` (or `HSQL_SESSION=NAME`) sends queries to it without paying start-up or connection cost again. Temp tables, settings and open transactions persist between invocations; `--session-reset` reconnects. POSIX only.
 - Press `alt+e` in the Query Editor to edit the current buffer in the editor named by `$VISUAL` or `$EDITOR`; quitting the editor with a non-zero status (like `:cq`) discards the changes ([#1102](https://github.com/tconbeer/harlequin/issues/1102)).
 - Harlequin now saves your open buffers every minute, and offers them back the next time you start if it exited without a clean quit ([#687](https://github.com/tconbeer/harlequin/issues/687)).
 - When Harlequin hits a bug in itself, it now exits with a short message instead of a traceback, saves your buffers, and writes a crash report you can attach to an issue ([#687](https://github.com/tconbeer/harlequin/issues/687)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- hsql can now hold a database connection open between invocations: `hsql --serve NAME [CONN_STR] [connection options]` starts a session in the foreground, and `hsql --session NAME -c ...` (or `HSQL_SESSION=NAME`) sends queries to it, paying neither the start-up nor the connection cost again. A session keeps its temp tables, settings and open transactions between invocations; `hsql --session NAME --session-reset` reconnects it. Not available on native Windows.
 - Press `alt+e` in the Query Editor to edit the current buffer in the editor named by `$VISUAL` or `$EDITOR`; quitting the editor with a non-zero status (like `:cq`) discards the changes ([#1102](https://github.com/tconbeer/harlequin/issues/1102)).
 - Harlequin now saves your open buffers every minute, and offers them back the next time you start if it exited without a clean quit ([#687](https://github.com/tconbeer/harlequin/issues/687)).
 - When Harlequin hits a bug in itself, it now exits with a short message instead of a traceback, saves your buffers, and writes a crash report you can attach to an issue ([#687](https://github.com/tconbeer/harlequin/issues/687)).
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -243,6 +243,11 @@ markers = [
     "online: requeries internet connection (deselect with '-m \"not online\"')",
     "use_cache: override the autouse fixture the disables the buffer cache",
     "py12: Only run on Python 3.12 or higher",
+    # a test whose invocations are *meant* to see different bytes warm than
+    # cold -- because the session kept state between them -- and which asserts
+    # what the warm behavior is. The set carrying it is the enumerated version
+    # of docs/plans/hsql-warm-sessions-design.md's section 5.
+    "session_divergent: asserts behavior a warm session has and a cold run has not",
 ]
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -243,11 +243,7 @@ markers = [
     "online: requeries internet connection (deselect with '-m \"not online\"')",
     "use_cache: override the autouse fixture the disables the buffer cache",
     "py12: Only run on Python 3.12 or higher",
-    # a test whose invocations are *meant* to see different bytes warm than
-    # cold -- because the session kept state between them -- and which asserts
-    # what the warm behavior is. The set carrying it is the enumerated version
-    # of docs/plans/hsql-warm-sessions-design.md's section 5.
-    "session_divergent: asserts behavior a warm session has and a cold run has not",
+    "session_divergent: asserts state a warm session keeps between invocations that a cold run does not",
 ]
 
 
--- a/src/harlequin/hsql/cli.py
+++ b/src/harlequin/hsql/cli.py
@@ -136,12 +136,20 @@ SHORTHANDS = {
 SOURCES = f"{__name__}.sources"
 """Context key under which `-c` and `-f` record themselves, in order."""
 
-CONNECTION_OPTIONS = frozenset(
-    {"conn_str", "adapter", "profile", "config_path", "read_only", *SSH_KEYS}
-)
+CONNECTION_OPTIONS = frozenset({"conn_str", "adapter", "read_only", *SSH_KEYS})
 """Answered once, when a connection is opened -- so `--serve` takes them and a
 served request may not. Every adapter option is one too, by
 `_is_connection_option()`: the command declares none of them itself."""
+
+CONFIG_OPTIONS = frozenset({"profile", "config_path"})
+"""Which file and which profile the other options are read from.
+
+Not connection-time, though a profile usually holds connection-time keys:
+these name where values come from rather than being values, so which group
+one belongs to is decided by what it resolves to. A profile of nothing but
+`format` and `limit` is per-request whichever way it was named, which is what
+lets a served `-P` behave the way a discovered `default_profile` does.
+"""
 
 SERVER_OPTIONS = frozenset({"queue_timeout"})
 """Answered once, and only a server has one to answer."""
@@ -636,7 +644,12 @@ def build_cli(argv: Sequence[str]) -> click.Command:
             _refuse_unservable_name(ctx, serve)
         elif served is not None:
             _refuse_the_sessions_options(
-                ctx, served, typed=typed_options, connects=connects
+                ctx,
+                served,
+                typed=typed_options,
+                connects=connects,
+                profile=profile,
+                profile_config=profile_config,
             )
         elif "queue_timeout" in explicitly_set:
             diagnostics.error(
@@ -1276,8 +1289,7 @@ def _serve(
 ) -> ExitCode:
     """Connect, and answer for the session called `name` until stopped.
 
-    The server is what `--session-reset` reconnects through, so it is handed
-    the adapter and not just the connection.
+    The server reconnects for `--session-reset`, so it is handed the adapter.
     """
     # here rather than at module scope: sockets and threads are the one
     # invocation in many that serves, and every other one would pay for them
@@ -1319,7 +1331,7 @@ def _is_connection_option(name: str) -> bool:
     an adapter's options are all connection-time.
     """
     return name in CONNECTION_OPTIONS or name not in (
-        PER_REQUEST_OPTIONS | SERVER_OPTIONS | ROLE_OPTIONS
+        PER_REQUEST_OPTIONS | SERVER_OPTIONS | ROLE_OPTIONS | CONFIG_OPTIONS
     )
 
 
@@ -1371,14 +1383,26 @@ def _refuse_unservable_name(ctx: click.Context, name: str) -> None:
 
 
 def _refuse_the_sessions_options(
-    ctx: click.Context, served: "Served", *, typed: set[str], connects: bool
+    ctx: click.Context,
+    served: "Served",
+    *,
+    typed: set[str],
+    connects: bool,
+    profile: str | None,
+    profile_config: Mapping[str, Any],
 ) -> None:
-    """Exit with an error if a served request typed an option the session owns.
+    """Exit with an error if a served request asked for what the session owns.
 
     A connection option describes a connection the session already opened,
-    so a request that typed one either agrees with it, and asked for nothing,
-    or differs, and asked for a different session. A server-lifetime option
-    describes the server, which is up.
+    and a server-lifetime option describes a server that is up. A typed `-P`
+    is judged by the keys its profile holds rather than by being one: a
+    profile of per-request keys applies, and one that names a database is
+    refused under the key that names it.
+
+    Only a *typed* profile, because that is the caller asserting this profile
+    for this invocation -- the rule `merge_profile_with_cli()` reads a command
+    line by. One discovered in the caller's directory says nothing about the
+    session, so its connection-time keys are moot rather than wrong.
     """
     for key in sorted(typed):
         spelling = _spelling(ctx, key)
@@ -1393,6 +1417,17 @@ def _refuse_the_sessions_options(
                 f"{spelling} is a connection option, and the session named "
                 f"{served.name!r} connected when it started. Drop it, or start "
                 f"a session with it: '{PROGRAM} --serve NAME {spelling} ...'."
+            )
+            ctx.exit(ExitCode.USAGE)
+    if profile is None or "profile" not in typed:
+        return
+    for key in sorted(profile_config):
+        if key in SERVER_OPTIONS or (connects and _is_connection_option(key)):
+            diagnostics.error(
+                f"the profile {profile!r} sets {key}, which the session named "
+                f"{served.name!r} answered when it started. Use a profile of "
+                f"per-request options here, or start a session with this one: "
+                f"'{PROGRAM} --serve NAME -P {profile}'."
             )
             ctx.exit(ExitCode.USAGE)
 
--- a/src/harlequin/hsql/cli.py
+++ b/src/harlequin/hsql/cli.py
@@ -27,6 +27,19 @@ because `CONN_STR` is positional: `hsql catalog` and a DuckDB file named
 `catalog` would have needed a rule, and `--catalog` needs none. They are
 mutually exclusive, and each lives in `harlequin.hsql.modes`, imported by the
 callback when it is chosen.
+
+A **session** is the fourth shape, and it is two roles of this one command.
+`--serve NAME` connects and then holds the connection, answering invocations
+sent to it over a socket; `--session NAME` is such an invocation, and it is the
+client in `harlequin.hsql.client` that carries it there, forwarding argv
+verbatim. So the server parses every served request with this same command,
+and what tells the callback it is answering for a session is `ctx.obj`. Every
+option here is in exactly one of three groups, decided by when its question
+can be answered: connection-time (`CONN_STR`, `-a`, `-P`, the SSH options,
+every adapter's), per-request (`-c`, `--format`, every mode), and
+server-lifetime (`--queue-timeout`). `--serve` refuses the second group and a
+served request refuses the third, and each refusal names the invocation the
+option belongs on.
 """
 
 from __future__ import annotations
@@ -54,9 +67,11 @@ from typing import (
 import click
 
 from harlequin.config import (
+    CLI_ONLY_SESSION_KEYS,
     CLI_ONLY_SSH_KEYS,
     DEFAULT_ADAPTER,
     DEFAULT_SSH_TIMEOUT,
+    SSH_KEYS,
     TUI_ONLY_KEYS,
     UNLIMITED,
     merge_profile_with_cli,
@@ -86,6 +101,7 @@ from harlequin.redact import hide_secrets_in
 
 if TYPE_CHECKING:
     from harlequin.adapter import HarlequinAdapter, HarlequinConnection
+    from harlequin.hsql.server import Served
     from harlequin.hsql.timeout import Deadline
     from harlequin.layout import LayoutOptions
     from harlequin.navigate import CatalogPath
@@ -120,6 +136,54 @@ SHORTHANDS = {
 SOURCES = f"{__name__}.sources"
 """Context key under which `-c` and `-f` record themselves, in order."""
 
+CONNECTION_OPTIONS = frozenset(
+    {"conn_str", "adapter", "profile", "config_path", "read_only", *SSH_KEYS}
+)
+"""Answered once, when a connection is opened -- so `--serve` takes them and a
+served request may not. Every adapter option is one too, by
+`_is_connection_option()`: the command declares none of them itself."""
+
+SERVER_OPTIONS = frozenset({"queue_timeout"})
+"""Answered once, and only a server has one to answer."""
+
+ROLE_OPTIONS = frozenset({"serve", "session"})
+"""The two spellings that say which process an invocation is."""
+
+PER_REQUEST_OPTIONS = frozenset(
+    {
+        "command",
+        "file",
+        "output",
+        "format",
+        *SHORTHANDS,
+        "vertical",
+        "tuples_only",
+        "no_align",
+        "no_header",
+        "no_footer",
+        "null_string",
+        "timeout",
+        "config_mode",
+        "catalog",
+        "catalog_search",
+        "path",
+        "spec",
+        "info",
+        "skill",
+        "session_reset",
+        "limit",
+        "display_rows",
+        "result",
+        "on_error",
+        "stats",
+        "color",
+        "version",
+    }
+)
+"""Answered on every invocation, so a session's client sends them and
+`--serve` takes none. `tests/unit_tests/test_hsql_serve.py` pins the four
+groups to the command: every parameter is in exactly one."""
+
 
 def build_cli(argv: Sequence[str]) -> click.Command:
     """Build the hsql command, importing at most one adapter to do it.
@@ -140,6 +204,7 @@ def build_cli(argv: Sequence[str]) -> click.Command:
             click.Option(["--spec"], is_flag=True),
             click.Option(["--info"], is_flag=True),
             click.Option(["--skill"], is_flag=True),
+            click.Option(["--session-reset"], is_flag=True),
         ],
         # A mode reports on what is installed or configured rather than
         # connecting with it, so there is no profile to read for one. `--config`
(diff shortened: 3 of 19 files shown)


src/harlequin/hsql/server.py
```diff
@@ -0,0 +1,623 @@
...
+            write_through=True,
+        )
+
+    def stderr(self) -> io.TextIOWrapper:
+        return io.TextIOWrapper(
+            _Stream(self, protocol.STDERR, isatty=self._stderr_isatty),
+            encoding="utf-8",
+            errors="backslashreplace",
+            write_through=True,
+        )
```
tconbeer, line 157: **Blocking (Windows CI):** both `TextIOWrapper`s take the default `newline=None`, which on write translates `\n` to `os.linesep`. That is why `test_the_recorder_keeps_the_order_the_streams_were_written_in` and `test_click_writes_binary_output_to_the_recorder` fail on every Windows job with `b'one\r\ntwo\n'`.

The server never runs on Windows, but the Recorder is pure and its tests run everywhere. More to the point, CPython builds `sys.stdout`/`sys.stderr` with `newline="\n"` on POSIX, so passing `newline="\n"` here is what makes the stand-in match the stream it replaces, rather than relying on `os.linesep` happening to be `\n`:

```python
return io.TextIOWrapper(
    _Stream(self, protocol.STDOUT, isatty=self._stdout_isatty),
    encoding="utf-8",
    newline="\n",
    write_through=True,
)
```

(same for `stderr()`).

---

tests/unit_tests/test_hsql_serve.py
```diff
@@ -0,0 +1,905 @@
...
+    styled, _ = raw_request(warm, argv=argv, stdout_isatty=True)
+    assert b"\x1b[" not in b"".join(data for _, data in plain)
+    assert b"\x1b[" in b"".join(data for _, data in styled)
+    honored, _ = raw_request(
+        warm, argv=argv, stdout_isatty=True, environ={"NO_COLOR": "1"}
+    )
+    assert b"\x1b[" not in b"".join(data for _, data in honored)
+
+
+def test_the_peer_uid_is_ours_on_a_socketpair() -> None:
```
tconbeer, line 999: **Blocking (Windows CI):** `socket.socketpair(socket.AF_UNIX)` raises `AttributeError: module 'socket' has no attribute 'AF_UNIX'` on Windows. Needs `@needs_unix_sockets`.

---

tests/unit_tests/test_hsql_serve.py
```diff
@@ -0,0 +1,905 @@
...
+
+
+@pytest.mark.parametrize(
+    "name,reason",
+    [
+        ("", "needs a name"),
+        ("bad name", "is not a session name"),
+        ("../etc", "is not a session name"),
+        ("x" * 65, "is not a session name"),
+    ],
```
tconbeer, line 166: **Blocking (Windows CI):** `cannot_be_served()` checks `AF_UNIX` before it checks the name, so on Windows the three name cases get `hsql sessions need a unix socket, which native Windows has not.` and fail the `reason in res.stderr` assertion. The empty-name case passes because it is ordered first.

The ordering in `cannot_be_served()` is deliberate and right; it is the test that needs to know. Mark the three:

```python
pytest.param("bad name", "is not a session name", marks=needs_unix_sockets),
```

---

src/harlequin/hsql/server.py
```diff
@@ -0,0 +1,623 @@
...
+
+if TYPE_CHECKING:
+    from harlequin.adapter import HarlequinConnection
+
+ACCEPT_POLL_SECONDS = 0.5
+"""How often the accept loop looks up from the socket to see if it was stopped."""
+
+REQUEST_TIMEOUT_SECONDS = 30.0
+"""How long a connected client has to send its request before it is dropped,
+so that a client that connects and stalls cannot hold a thread forever."""
```
tconbeer: **Bug: this timeout breaks `-f -` typed at a terminal.** The client reads stdin *after* it receives `HELLO` (correctly, so an ambient fallback still has its stdin), and only then sends `REQUEST`. So the 30s clock here starts while a human is still typing. Reproduced end to end with the real console script and a stdin that arrives after 32s:

```
$ hsql --session slow -tAf -        # stdin written 32s later
hsql: error: the session named 'slow' did not answer: [Errno 32] Broken pipe
(exit 3)
```

Server log: `dropped a connection that sent no request: timed out`. Cold, the same invocation waits as long as the human does.

What the timeout buys is small: `_attend` only enters the turnstile *after* the request frame arrives, so a stalled client holds a daemon thread blocked in `recv`, not a turn at the connection. I'd drop it (or make it hours), since the client already waits indefinitely for its answer in the other direction. If keeping it, the client would have to read stdin before connecting, which costs the ambient fallback.

---

src/harlequin/hsql/cli.py
```diff
@@ -990,21 +1209,222 @@ def _connect(
...
+    """
+    for key in sorted(typed):
+        spelling = _spelling(ctx, key)
+        if key in SERVER_OPTIONS:
+            diagnostics.error(
+                f"{spelling} is a --serve option, and the session named "
+                f"{served.name!r} already has one."
+            )
+            ctx.exit(ExitCode.USAGE)
+        if connects and _is_connection_option(key):
```
tconbeer, line 1415: **Design question: `-P` on a served request.** `profile` is in `CONNECTION_OPTIONS`, so `hsql --session prod -P csv-out -c ...` is refused here, even when that profile carries only per-request keys (`format`, `limit`, `color`). Meanwhile a *default* profile discovered in the client's cwd is applied (as `test_a_config_file_where_the_client_is_applies_its_per_request_keys` pins). The design doc (§4.4, around "`-P` needs no special case on either side") reads as: a client's `-P` resolves against the client's cwd, its per-request keys apply, and only its connection-time keys go through the comparison.

If the intent is "refuse until PR 3 can compare", fine, but then PR 3's comparison is on option *values*, and `-P` is a name that resolves to a bag of both kinds, so it likely needs its own rule there rather than falling out of `_is_connection_option`. Worth deciding now whether `-P` belongs in this set, or whether the check should look at what the profile *resolved to* (its connection-time keys, via the same `_is_connection_option` filter over the merged values).

Related, one line up at 647: a profile's `queue_timeout` on a served request is popped and silently ignored, while a typed `--queue-timeout` is refused. Small asymmetry; a note or a refusal would make them agree.

---

src/harlequin/hsql/server.py
```diff
@@ -0,0 +1,623 @@
...
+            stdout_isatty=request.stdout_isatty, stderr_isatty=request.stderr_isatty
+        )
+        try:
+            with self._as(request, recorder):
+                code = cli.run(request.argv, served=Served(self, request))
+        except NotADirectoryError as e:
+            diagnostics.error(str(e), stream=recorder.stderr())
+            return recorder.segments, ExitCode.USAGE
+        except Exception as e:  # noqa: BLE001 -- the server outlives a request
+            traceback.print_exc(file=self._stderr)
```
tconbeer: Low: this is the one line on the server's stderr that bypasses `diagnostics._write()` and so `redact_text()`. A driver exception that quotes a DSN back (the case `redact_text` exists for, per AGENTS.md) lands on the operator's terminal unredacted here, while the same message reaches the client redacted on the next line. The operator typed the DSN, so it is their own secret, but it breaks the "every message leaves through that one line" promise. `traceback.format_exc()` through `redact_text()` before writing would keep it.

---

tests/unit_tests/test_hsql_serve.py
```diff
@@ -0,0 +1,905 @@
...
+    # the abandoned connection is left to its thread, not closed under it
+    assert not opened[0].closed  # type: ignore[attr-defined]
+    session.reset()
+    assert session.connection() is opened[-1]
+
+
+# --- the turnstile ---------------------------------------------------------------
+
+
+def test_the_turnstile_serves_in_arrival_order() -> None:
```
tconbeer, line 492: Two test notes:

1. **Coverage gap:** the `abandon` path is exercised only with `_FakeConnection` and a `threading.Event`. Nothing sends a real served request with `--timeout` that expires against DuckDB and asserts (a) the client exits 4 with the timeout message, (b) the *next* request gets `cancelled a query that did not stop` / exit 3, and (c) `--session-reset` recovers it. That chain crosses `Deadline.run` → `Served.abandon` → `Server.connection()` → `_connection_for` → `_reset_session`, and each hop is tested in isolation only. A `select` over a large `range()` with `--timeout 0.1` against the `warm` fixture would cover it.

2. **Timing:** this test orders tickets with `time.sleep(0.05 * number)` and asserts `queued == 2` after a fixed `0.3s`. On a loaded runner thread 2 can take its ticket before thread 1, and the assertion on `[1, 2]` fails for reasons unrelated to the turnstile. A `Barrier`/`Event` per thread that releases the next only after the previous has called `enter()` (observable via `queued`) makes the order deterministic.

---

pyproject.toml
```diff
@@ -243,6 +243,11 @@ markers = [
     "online: requeries internet connection (deselect with '-m \"not online\"')",
     "use_cache: override the autouse fixture the disables the buffer cache",
     "py12: Only run on Python 3.12 or higher",
+    # a test whose invocations are *meant* to see different bytes warm than
+    # cold -- because the session kept state between them -- and which asserts
+    # what the warm behavior is. The set carrying it is the enumerated version
+    # of docs/plans/hsql-warm-sessions-design.md's section 5.
+    "session_divergent: asserts behavior a warm session has and a cold run has not",
```
tconbeer: Conventions: this comment relays the design doc ("the enumerated version of docs/plans/…'s section 5"), which AGENTS.md says belongs in the PR or the plan, not the code. The other three markers carry no comment. Suggest deleting the four comment lines and letting the marker text carry it:

```toml
"session_divergent: asserts state a warm session keeps between invocations that a cold run does not",
```

---

tests/unit_tests/test_hsql_serve.py
```diff
@@ -0,0 +1,905 @@
...
+        ["other.db"],
+    ],
+)
+def test_a_served_request_may_not_type_a_connection_option(
+    hsql: Hsql, in_process_server: Server, args: list[str]
+) -> None:
+    """The session connected when it started, so its connection is fixed: a
+    request that typed a connection option would otherwise run on the session's
+    connection while believing it had changed it. (Comparing the value against
+    the session's, so an identical one is served, is PR 3's refinement.)
```
tconbeer: Conventions: "(Comparing the value against the session's … is PR 3's refinement.)" narrates a future PR, which AGENTS.md rules out of docstrings. Delete the parenthetical; the PR description already carries it.

Same pattern, smaller, elsewhere in this PR:
- `cli.py` `_serve` docstring: "so it is handed the adapter and not just the connection" → "The server reconnects for `--session-reset`, so it is handed the adapter."
- `Turnstile` docstring: "A lock would serve requests in whatever order…" is an alternative weighed. "Hands out tickets so the longest-waiting client goes next and a waiter can give up on its own deadline." says the same without it.
- The "a thread still inside a driver aborts an interpreter that exits around it" rationale appears in `timeout.py`, twice in `server.py`, and in a test docstring. Once, in `timeout.py`, is enough.
- `test_hsql_equivalence.py:44-47`: "This is the drift surface the design cares about" → cut.

---

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- hsql can now hold a database connection open between invocations: `hsql --serve NAME [CONN_STR] [connection options]` starts a session in the foreground, and `hsql --session NAME -c ...` (or `HSQL_SESSION=NAME`) sends queries to it, paying neither the start-up nor the connection cost again. A session keeps its temp tables, settings and open transactions between invocations; `hsql --session NAME --session-reset` reconnects it. Not available on native Windows.
```
tconbeer: Three sentences and the option grammar; `.claude/rules/changelog.md` asks for one or two. A trim that keeps the user-facing facts:

> `hsql --serve NAME [CONN_STR]` holds a database connection open as a named session, and `hsql --session NAME -c ...` (or `HSQL_SESSION=NAME`) sends queries to it without paying start-up or connection cost again. Temp tables, settings and open transactions persist between invocations; `--session-reset` reconnects. POSIX only.

---

tests/hsql_sessions.py
```diff
@@ -0,0 +1,105 @@
...
+            [
+                sys.executable,
+                "-c",
+                "import sys\n"
+                f"sys.argv = ['hsql', '--serve', {name!r}, *{list(serve_argv)!r}]\n"
+                "from harlequin.hsql import main\n"
+                "main()\n",
+            ],
+            stdout=subprocess.PIPE,
+            stderr=self.stderr_path.open("wb"),
```
tconbeer: Nit: this file object is handed to `Popen` and never closed in the parent, so each `WarmSession` leaks a descriptor until GC (`ResourceWarning` under `-W error`). Keep it on `self` and close it in `stop()` alongside `process.stdout`.

---

tconbeer (commented):
## Scope

Lenses applied: correctness of the server lifecycle and the served request path (turnstile, signals, stream/cwd/env swap, `abandon`, exception escape from `cli.run()`), the flag partition, security of the socket (uid check, runtime dir, lock, what the request can make the server do), test assertions vs. their names, and docstring/comment/CHANGELOG conventions from AGENTS.md. Verified locally on Linux: the hsql-related unit tests (702 passed), mypy, ruff, and all five import-linter contracts. Two things reproduced end to end against a real server in a fresh interpreter.

## Verdict: changes needed before merge

(Submitted as a comment because GitHub does not allow "request changes" on one's own PR.)

One hard reason: **Windows CI is red on all five Python versions** (6 tests). All three causes are test-side or a one-argument fix, none is a design problem, and they are pinned inline:

- `Recorder.stdout()/stderr()` need `newline="\n"` (matches CPython's POSIX `sys.stdout`; fixes 2 tests).
- `test_the_peer_uid_is_ours_on_a_socketpair` needs `@needs_unix_sockets`.
- Three name cases in `test_serve_refuses_a_name_no_client_could_reach` need `needs_unix_sockets` marks, since the AF_UNIX check correctly precedes the name check.

One **real bug**, reproduced: `REQUEST_TIMEOUT_SECONDS` (30s) starts before the client has read stdin, so `hsql --session NAME -f -` typed at a terminal for more than 30s exits 3 with "did not answer: Broken pipe". Since a stalled client never enters the turnstile, the timeout protects only a daemon thread in `recv`; I'd drop it.

One **design question** to settle now rather than in PR 3: `-P` on a served request is refused as a connection option even when the profile carries only per-request keys, while a discovered default profile applies; the design doc says `-P` needs no special case.

The rest is a coverage gap (no end-to-end `--timeout` → abandon → reset test), one timing-based test, an unredacted traceback on the operator's stderr, and conventions (a marker comment that cites the design doc, a docstring that narrates PR 3, a three-sentence CHANGELOG entry).

What I checked and found sound: the turnstile's FIFO order and its use at shutdown (queued requests *are* answered before the connection closes, as claimed); signal handler installed before `listen()`; `_as()` restores cwd/env/streams on every path; the flock-then-unlink stale-socket sequence (and leaving the `.lock` file behind is the correct choice; unlinking under flock is the classic two-servers race); uid check on accept plus the 0700 directory; `cli.run()` catches `ClickException`/`Abort`/`KeyboardInterrupt` and `ctx.exit()` under `standalone_mode=False`, and nothing on the served path calls `sys.exit()`; the equivalence suite is xdist-safe (one runtime dir per worker); import hygiene holds (client never imports the server, contract kept).

## Not checked

- I did not run the suite on macOS or Windows; the Windows analysis is from the CI logs, and the `LOCAL_PEERCRED` / `struct xucred` layout on macOS is from reading, not from a run.
- I did not verify byte-equivalence under a non-UTF-8 locale: cold `sys.stderr` is locale-encoded and the Recorder's is UTF-8, so a diagnostic containing non-ASCII could differ there. The suite runs under `en_US.UTF-8`.
- I did not test a served `--timeout` that expires against a real driver (the coverage gap noted inline), SIGTERM mid-request, or the repeated-Ctrl-C `halt()` path.
- I did not review `docs/generated/hsql-reference.md` or `schemas/config-v1.json` beyond confirming the pinning tests pass.
- I did not evaluate the design doc's choices themselves (one request at a time, `flock`, refuse-vs-compare); I reviewed the implementation against the doc.
- A clean review is necessary, not sufficient: the merge call is yours.

---

src/harlequin/hsql/server.py
tconbeer: Fixed in 017fc00 — both stand-ins now take `newline="\n"`.

Your framing is the one I took into the docstring: the point isn't that the tests are cross-platform, it's that the stand-in should match the stream it replaces, and the bytes are the contract either way.

```
segments: [(3, b'one\ntwo\n'), (4, b'note\n')]
```

---

tests/unit_tests/test_hsql_serve.py
tconbeer: Fixed in 017fc00 — `@needs_unix_sockets` added.

---

tests/unit_tests/test_hsql_serve.py
tconbeer: Fixed in 017fc00 — the three name cases carry `marks=needs_unix_sockets`, with a comment noting the empty-name case is answered before the platform is because a flag with no value is the caller's typo everywhere.

---

src/harlequin/hsql/server.py
tconbeer: Good catch, and thanks for reproducing it — dropped the timeout in 017fc00.

Your analysis of what it bought is what decided it: `_attend` blocks in `recv` *before* it enters the turnstile, so a stalled client costs a daemon thread and not a turn at the connection. That is not worth a clock that can expire on a human. The alternative you named — reading stdin before connecting — trades a real bug for a worse one, since it would make the ambient fallback consume stdin it may not need.

Your repro, on the new head:

```
$ hsql --session slow -tAf -        # stdin written 32s later
typed slowly
(exit 0)
```

and the server log has no `dropped a connection` line.

---

src/harlequin/hsql/cli.py
tconbeer: Decided it your way, in 017fc00: the check now looks at what the profile resolved to.

The inconsistency you pointed at is the argument — a typed `-P csv-out` was refused while `default_profile = "csv-out"` in the same directory applied, which is the same profile reaching the same session and getting two answers. So `profile` and `config_path` leave `CONNECTION_OPTIONS` for a group of their own:

```python
CONFIG_OPTIONS = frozenset({"profile", "config_path"})
"""Which file and which profile the other options are read from.

Not connection-time, though a profile usually holds connection-time keys:
these name where values come from rather than being values, so which group
one belongs to is decided by what it resolves to. ...
"""
```

A typed `-P` is then refused per *key*, and the message names the key rather than the flag:

```
hsql: error: the profile 'prod' sets adapter, which the session named 'inproc'
answered when it started. Use a profile of per-request options here, or start a
session with this one: 'hsql --serve NAME -P prod'.
```

One thing I decided against, and it is worth flagging since it is the half your comment leaves open: the check applies only to a **typed** `-P`, not to whatever the merged values happen to hold. Refusing on resolved values generally would refuse every warm invocation run inside a project whose `default_profile` names a database — which is most projects that have a config file. A typed `-P` is the caller asserting *this profile, for this invocation*; a discovered one is ambient and says nothing about the session. Same asymmetry as `--session` against `HSQL_SESSION`, and the rule `merge_profile_with_cli()` already reads a command line by.

That also makes PR 3's job cleaner than it would have been: it replaces "refuse" with "compare" over exactly the connection-time keys, and `-P` needs no rule of its own — the resolution already happened.

On `queue_timeout`: it is in `SERVER_OPTIONS`, and the profile loop checks that set too, so a typed `-P` carrying it is now refused alongside the connection keys. A *discovered* profile's `queue_timeout` is still popped silently on a non-serve run — same treatment as `TUI_ONLY_KEYS`, which is the established pattern for a key belonging to the other role, and the same reasoning as above for why an ambient value should not fail an invocation.

New tests: `test_a_served_request_takes_a_profile_of_per_request_options`, and `test_a_served_request_refuses_a_profile_that_names_a_connection` over `adapter`/`read_only`/`conn_str`.

---

src/harlequin/hsql/server.py
tconbeer: Fixed in 017fc00 — it goes through `redact_text()` now:

```python
diagnostics.note(
    f"request failed: {redact_text(traceback.format_exc())}",
    stream=self._stderr,
)
```

You're right that the operator typed the DSN, but a server is the one process where "the operator's terminal" and "a log someone else reads" are the same stream for eight hours, so the promise is worth more here than the exemption.

---

pyproject.toml
tconbeer: Fixed in 017fc00 — took your wording verbatim and deleted the four comment lines.

---

tests/unit_tests/test_hsql_serve.py
tconbeer: Both addressed in 017fc00, the first not quite as asked — worth saying why.

**2 (timing):** rewritten to order by observation. Each thread starts and the loop waits until `turnstile.queued == n` before starting the next, so the arrival order is the loop's rather than the scheduler's, and the fixed `0.3s` is gone.

**1 (coverage):** I added the end-to-end test, but it asserts (a) and a session that still works — not (b) and (c). The chain you listed can't be reached with a real adapter: DuckDB's `cancel()` *lands*, so `finished.wait(self.grace)` returns and `abandon` is never called. `abandon` only fires when a cancel is ignored for the whole grace period, which no adapter that implements cancellation does. Reaching it against DuckDB would mean a query its own interrupt can't stop, which I don't think exists, or patching `GRACE_SECONDS` in the server subprocess, which would be testing the patch.

So the new test covers the path a real `--timeout` actually takes, which was untested and is the common one:

```python
proc = send(["--timeout", "0.1", "-c", "select count(*) from range(2000000000)"])
assert proc.returncode == ExitCode.TIMEOUT
assert b"timed out after 0.1s" in proc.stderr
assert send(["-tAc", "select 'still up'"]).stdout == b"still up\n"
```

That crosses `Deadline.run` → cancel → attribution → the session surviving, end to end. The remaining hops (`Served.abandon` → `Server.connection()` → `_reset_session`) stay unit-tested against a fake, and the docstring now says why rather than leaving it looking like an oversight. If you'd rather have the full chain even at the cost of a fake adapter in the server subprocess, say so and I'll add it.

---

tests/unit_tests/test_hsql_serve.py
tconbeer: All five fixed in 017fc00: the PR-3 parenthetical and the `-P`/`--config-path` note are gone (the latter no longer true anyway — see the `-P` thread), `_serve`'s docstring is your sentence, the `Turnstile` docstring states what it does, the interpreter-abort rationale is now only in `timeout.py`, and the "drift surface" line is cut.

---

CHANGELOG.md
tconbeer: Fixed in 017fc00 — used your wording verbatim.

---

tests/hsql_sessions.py
tconbeer: Fixed in 017fc00 — kept on `self._stderr_file` and closed in `stop()` beside `process.stdout`.

---

PR #1140, opened 2026-09-04
Warm sessions 3/6: identity, rejection, and --session-status

tconbeer:
No issue; this is PR 3 of the six in [`docs/plans/hsql-warm-sessions-design.md`](https://github.com/tconbeer/harlequin/blob/main/docs/plans/hsql-warm-sessions-design.md) §9, on top of #1138.

§4.4's promise, delivered: **the name is the caller's and the identity is the server's.** `--serve` records the connection options it resolved at start-up, and a served request's own are compared against that record rather than refused wholesale.

- **The recorded identity.** `connection_options_in()` takes the connection-time half of the merged values, from one snapshot on both sides — `--serve` records it, a served request builds the same thing from its own values, and the two are compared key by key. It holds the keys a profile or a command line named; a key named by neither takes its value from the adapter's own default, which core cannot enumerate.
- **The rejection.** Equal values are served — so #1138's promise that "an identical `-a duckdb` starts being served" lands here. Differing ones exit 2, naming the flag and *both* values, since what the session was started with is the half a caller cannot see:
  ```
  hsql: error: --adapter is 'sqlite'. The session named 'prod' was started with
  'duckdb', and its connection is fixed. Drop it here, or start a session with
  it: 'hsql --serve NAME --adapter ...'.
  ```
  A typed `-P` is compared key by key the same way, under the key that differs rather than as a flag — so the profile a session was *started* with, the one a caller is most likely to name, is usable on a request to it. Both values go through `harlequin.redact` first: a connection option is exactly where a password is typed.
- **`--session-status`.** JSON on stdout — session, pid, version, adapter, redacted connection and connection options, uptime, request count, state, queue depth, transaction mode, ssh — reported while a query is running, which is what makes "is it hung or is it slow" a question with an answer.
- **`HARLEQUIN_CONFIG_PATH` now travels with a request**, joining `NO_COLOR` in `protocol.FORWARDED_ENV_VARS`. It also joins `CLI_ONLY_SESSION_KEYS` alongside `session` and `serve`, and is dropped from the generated schema.
- **`docs/cli-options.md`** documents the five option groups and the checklist for adding an option, linked from AGENTS.md and CONTRIBUTING.md.

Three decisions this PR made on top of the doc:

- **`--session-status` is not a request, and this was the PR's real design question.** §4.5 promises it is answerable mid-query, and a request cannot be: a served invocation borrows the process's cwd, environment and streams for its whole run — the invariant #1138 called load-bearing beyond concurrency — so a second one parsed alongside a query would take them out from under it. And a status that waited its turn would report only once the query it was asked about had finished, which is exactly when nobody wants it. So the client sends it as its own frame (a third thing its argv scan knows about, beside `--session` and `-f -`) and the server reports from its own bookkeeping. That is the shape PR 4's cancel wants anyway, so it pays for itself twice.

  It stays a declared click option, so `--help`, `--spec` and the flag partition still describe it and a cold invocation is refused rather than ignored — the same shape `--session` itself has. The frame carries the caller's argv, so a flag typed beside it is refused *by name*; no parser sees a status request, so silence there would be the one place this could mislead. And it joins `CLI_ONLY_SESSION_KEYS` because the client reads it off argv and not from config: a profile could otherwise set it only for the runs that never reach a session, which is a divergence between the two paths rather than a setting.

  `transaction_mode` is the one field read from the connection, so it takes a zero-timeout turn at the turnstile rather than reading beside a request — the adapter contract says nothing about thread-safety. A busy session reports it as null without waiting. `idle_timeout_s` and `expires_in_s` are null rather than absent until PR 5 fills them, so the document's shape does not change under a caller already parsing it.

- **`HARLEQUIN_CONFIG_PATH` travels.** The doc left this open, noting it widens `FORWARDED_ENV_VARS` past "what `--color auto` needs," which is the line #1132 drew. It should be widened: a caller who typed `--config-path` and a caller who exported the variable named the same file for the same reason, and a session that honored one and ignored the other would read a config file *neither* of them asked for — the same failure §4.2 forwards the client's cwd to avoid. The replacement line is narrower than it looks: both variables on it name *where a value comes from* rather than being one, and neither is a credential. Still not the caller's whole environment — an adapter option's `envvar` resolves against the server's, and shipping the rest across a socket is a credential leak with no upside.

- **An option the session never named is refused, not assumed to match.** The alternative was recording every declared option's click default in the identity, so a request typing a value equal to the default would be served. Rejected: click's default for an adapter option is frequently `None` while the adapter's real default is something else, so it would trade an honest *"the session named 'prod' was started without `--port`"* for a confusing *"was started with `None`."*

Tradeoffs worth naming: the comparison is of what each side was *given*, not of what a path resolves to, so `--serve prod ./f1.db` followed by `--session prod f1.db` is refused even though both name the same file. Normalizing would be wrong for the DSNs that are not paths, and refusing something that would have worked is the safe direction — a caller should not be passing the connection to a client at all. And the status document carries `connection_options`, a key beyond §4.10's sketch: it is the identity the rejection compares against, which is what §4.7 means by "the redacted connection identity," and a caller who has just been refused wants to see it.

**From review.** Two rounds, both worth the space here:

- `TUI_ONLY_KEYS` were classified as connection options, so a typed `-P` was refused whenever its profile held a `theme` — the feature failing on the input it was built for. The cause was `_is_connection_option()` working by *exclusion*. It is now `connection_option_names()`, a positive test: hsql's own keys plus the adapter's declared ones. That deletes `NOT_AN_ADAPTERS_OPTION`, `_is_connection_option` and the `TUI_ONLY_KEYS` exception with it.
- `_send_status` had no exception net, so a bug there put a server traceback into an *unrelated* caller's stderr; it now answers exit 70 with a crash report, like the request path.
- A typed profile's `queue_timeout` had stopped being refused; `conn_str` as a scalar string never matched an identical command line, and the message then printed the same value on both sides; a non-string `conn_str` crashed to exit 70 where a usage error was owed. All fixed, with `redact_conn_str` hardened so it cannot raise — `harlequin.crash` calls it.
- Docstrings and error messages were rewritten across the diff: what the code does, not what it does not do, and no sessions asking or answering. Every user-facing message ends with what to do about the failure.

- [ ] No.
- [x] Yes; I haven't opened a PR, but the gist of the change is: the session section the "Headless & Agents" topic still needs (PR 6 of the plan) should now also cover `--session-status` as the answer to "what is my session doing," and say that a request may carry the connection options the session already has but not different ones — with `-P` as the spelling to show, since a profile of connection keys is now usable on both sides. `docs/cli-options.md` in this PR is contributor documentation and stays in-repo.

- [x] Yes.

- **Identity and rejection** (`test_hsql_serve.py`) — a request naming a *different* connection refused with both values asserted; one naming the connection the session has, served; an option the session never named, refused as such; a typed profile on both sides of the same rule; an adapter's `secret=` declaration masked in a refusal; a mistyped profile value as exit 2 rather than a crash; and a test pinning the identity to the partition.
- **`--session-status`** — the document end to end against a real session; `busy` while a request holds the connection, and again while duckdb is actually executing; a flag beside it refused by name; the `=` form not queueing behind a query; a bug in the status path as exit 70 and not another caller's stderr; the busy short-circuit pinned against a connection that counts driver reads.
- **`HARLEQUIN_CONFIG_PATH`** — a caller's variable reaching the session, and the *operator's* not reaching a request that did not set one. Plus an end-to-end test for §4.2's typed-`-P`-resolves-in-the-client's-cwd claim.

Full suite green: 1823 passed / 7 skipped on 3.10, the py12 subset green, mypy clean over 148 files, all five import-linter contracts kept, ruff clean. The client's module budget is untouched — `requests_status` and `pack_status` live in the two modules it already imports. `docs/generated/hsql-reference.md` regenerated; `schemas/config-v1.json` unchanged, since `session_status` is a CLI-only key.

https://claude.ai/code/session_018w66UxvaY8Z9mF5rf2kRW9


src/harlequin/hsql/cli.py
```diff
@@ -1390,66 +1465,136 @@ def _refuse_the_sessions_options(
...
+    if not connects:
+        # a mode that reads no database is not the session's business: it runs
+        # where the caller is, with the options the caller typed
+        return
+    typed_keys = {key for key in typed if _is_connection_option(key)}
+    profile_keys: set[str] = set()
+    if profile is not None and "profile" in typed:
+        profile_keys = {
+            key for key in profile_config if _is_connection_option(key)
+        } - typed_keys
```
tconbeer: **A typed `-P` is refused whenever its profile holds any IDE key — which defeats the feature this PR adds.**

`values` has `TUI_ONLY_KEYS` popped at :634-635, *before* the identity snapshot at :684, so a session structurally cannot record `theme`, `locale`, `keymap_name`, `show_files`, `show_s3`, `no_download_tzdata` or `viewer_max_rows`. But `profile_keys` here is built from the raw `profile_config`, and none of those seven is in `PER_REQUEST_OPTIONS | SERVER_OPTIONS | ROLE_OPTIONS | CONFIG_OPTIONS` — so `_is_connection_option()` returns `True` for every one of them:

```
theme -> True    keymap_name -> True    show_files -> True    show_s3 -> True
locale -> True   no_download_tzdata -> True   viewer_max_rows -> True
```

Each is then compared against an identity that can never contain it, and takes the `not known` branch. End to end, with a profile that names the session's own connection *and* a theme:

```
$ hsql --serve prod -P prod &
$ hsql --session prod -P prod -tAc "select 1"
hsql: error: the profile 'prod' sets theme, a connection option the session
named 'prod' did not ask for. ...
exit=2
```

The session was started from that exact profile. The misclassification predates this PR, but on `main` a typed `-P` carrying any connection key was refused outright, so it was unreachable; this PR's stated goal is "the profile a session was *started* with, which is the one a caller is most likely to name, is now usable on a request to it." `TUI_ONLY_KEYS`' own docstring says "One profile serves both commands, so a profile written for the IDE has to work headless" — so a profile with a `theme` in it is the normal case, not the exotic one.

`test_a_served_request_takes_a_profile_that_names_the_sessions_connection` passes only because its fixture profile carries no IDE key.

Fix belongs in `_is_connection_option` (or by subtracting `TUI_ONLY_KEYS` here), so that both halves agree on what a connection option is.

---

src/harlequin/hsql/server.py
```diff
@@ -489,6 +583,34 @@ def _attend(self, connection: socket.socket) -> None:
...
+                stream=recorder.stderr(),
+            )
+            self._answer(connection, recorder.segments, ExitCode.USAGE)
+            return
+        print(
+            json.dumps(self.status(), separators=(",", ":"), default=str),
+            file=recorder.stdout(),
+        )
+        self._answer(connection, recorder.segments, ExitCode.OK)
+        diagnostics.report_status_answered(stream=self._stderr)
```
tconbeer: **No exception net here, and the traceback lands in a *different* caller's terminal.**

`_attend`'s `try` catches only `(protocol.ProtocolError, OSError)`. The REQUEST path gets a second layer from `_run`'s `except Exception` → crash report → exit 70. `_answer_status` has neither, so anything `status()` raises — `redact_profile`/`redact_conn_str` over adapter-supplied identity values, or `json.dumps` — escapes the thread.

That matters more than usual because of what the escaping traceback hits: `threading.excepthook` writes to `sys.stderr`, and `_as` has swapped `sys.stderr` process-wide to the **in-flight request's** recorder. Demonstrated in-process with `status()` patched to raise while a stand-in request held the streams:

```
what the status client received: None
what the *other* request's stderr now carries:
  kind=4 b'Exception in thread Thread-1 (_attend):\nTraceback (most recent call last):\n ...'
```

So a bug in the status path delivers a server-side traceback with absolute paths into an unrelated caller's diagnostics, and the status client sees a closed socket → exit 3 (`CONNECTION`, "the database cannot be reached") for what is a bug in hsql, where the request path would give 70.

Wrapping this the way `_run` is wrapped fixes both halves.

---

src/harlequin/hsql/cli.py
```diff
@@ -1390,66 +1465,136 @@ def _refuse_the_sessions_options(
...
     for key in sorted(typed):
-        spelling = _spelling(ctx, key)
         if key in SERVER_OPTIONS:
             diagnostics.error(
-                f"{spelling} is a --serve option, and the session named "
-                f"{served.name!r} already has one."
+                f"{_spelling(ctx, key)} is a --serve option, and the session "
+                f"named {served.name!r} already has one."
             )
             ctx.exit(ExitCode.USAGE)
```
tconbeer: **Regression: a typed profile that sets `queue_timeout` is no longer refused — it is silently dropped.**

`main`'s profile loop was `if key in SERVER_OPTIONS or (connects and _is_connection_option(key))`. The rewrite checks `SERVER_OPTIONS` only over `typed` (here), then returns early on `not connects` (:1496-1499), and the profile loop that follows filters on `_is_connection_option` alone (:1503-1505). Nothing looks at a profile's server-lifetime keys any more.

With `[profiles.slow] queue_timeout = 30`:

```
# origin/main
$ hsql --session prod -P slow -tAc "select 1"
hsql: error: the profile 'slow' sets queue_timeout, which the session named
'prod' answered when it started. ...        exit=2

# this branch
$ hsql --session prod -P slow -tAc "select 1"
1                                           exit=0
```

The value is popped at :680 and parsed at :901, then only ever handed to `_serve` — so on a served request it is dead. Two follow-on oddities: the rule now depends on *where* the value came from (typed `--queue-timeout` is still refused) and on whether the mode connects (typed `--queue-timeout --info` is still refused; profile `queue_timeout` + `--info` is not); and an **invalid** profile value still fails the request under a flag it could not have used:

```
[profiles.prod] queue_timeout = "soon"
$ hsql --session prod -P prod -c "select 1"
hsql: error: --queue-timeout='soon' is not a positive number of seconds. ...
```

This docstring still says "A server-lifetime option describes a server that is up, and is refused **outright**" — true only for typed options now. Nothing covered this before and nothing covers it now, which is why the removal produced no test failure; it belongs in the `test_a_served_request_refuses_a_profile_that_names_another_connection` parametrize, beside `test_a_served_request_may_not_type_a_server_option`.

---

src/harlequin/hsql/cli.py
```diff
@@ -1382,6 +1433,30 @@ def _refuse_unservable_name(ctx: click.Context, name: str) -> None:
...
+
+
+def _comparable(value: Any) -> Any:
+    """One connection option's value, in the shape both sides spell it.
+
+    A profile writes `conn_str = ["f1.db"]` and a command line produces a
+    tuple; nothing else about a connection option's value differs by where it
+    came from, since both sides parsed it with the same adapter's options.
+    """
+    return tuple(value) if isinstance(value, list) else value
```
tconbeer: **A scalar-string `conn_str` in a profile never matches, and the refusal then prints the same value on both sides.**

`conn_str` is not an adapter option, so `parse_profile_options` skips it as a `command_option` and a profile's value reaches `values` exactly as TOML wrote it. A bare string is fully supported — :692-694 does `if isinstance(conn_str, str): conn_str = (conn_str,)`, and `conn_str = ":memory:"` runs fine cold. So a session started from such a profile records `":memory:"` while the CLI produces `(":memory:",)`, and list→tuple alone never reconciles them.

Worse, `redact_profile` normalizes *both* sides to a list before printing, so the message contradicts itself:

```
$ hsql --serve prod -P prod &        # profile has conn_str = ":memory:"
$ hsql --session prod :memory: -tAc "select 1"
hsql: error: CONN_STR says [':memory:'], and the session named 'prod' connected
with [':memory:']. Its connection is fixed, so drop it here, ...
exit=2
```

The mirror case (`--serve prod f1.db`, request `-P prod` with `conn_str = "f1.db"`) produces the same "these two identical things differ." Folding a string `conn_str` to a 1-tuple here would fix it; comparing the normalized values would additionally make it impossible for the message to disagree with the comparison.

---

src/harlequin/hsql/cli.py
```diff
@@ -1390,66 +1465,136 @@ def _refuse_the_sessions_options(
...
-                f"'{PROGRAM} --serve NAME -P {profile}'."
+                f"the profile {profile!r} sets {key}, a connection option the "
+                f"session named {served.name!r} did not ask for. Use a profile "
+                f"of per-request options here, or start a session with this "
+                f"one: '{PROGRAM} --serve NAME -P {profile}'."
             )
-            ctx.exit(ExitCode.USAGE)
+        ctx.exit(ExitCode.USAGE)
+    mine = redact_profile({key: asked}, options)[key]
+    theirs = redact_profile({key: has}, options)[key]
```
tconbeer: **A non-string `conn_str` in a profile turns this usage error into a crash report (exit 70).**

`redact_profile` routes anything under `conn_str` through `redact_conn_str`, which assumes strings. `hide_secrets_in` guards with `isinstance(item, str)` (`redact.py:72-78`); `redact_profile` does not. Nothing validates a profile's `conn_str` type — `Profile = Dict[str, Any]`, and `parse_profile_options` skips command options — so:

```python
>>> redact_profile({"conn_str": 5})
TypeError: expected string or bytes-like object
```

Confirmed for `5`, `True`, `None`, `3.5`, `(1, 2)`, and any list containing a non-str. With `[profiles.bad] conn_str = 5`:

```
$ hsql --session prod -P bad -c "select 1"
→ TypeError escapes into main()'s crash handler → ExitCode.CRASH (70) + bug-report prompt
```

Expected is the ordinary USAGE refusal naming the differing `conn_str`. A caller gets told they found a bug in Harlequin when they made a typo in their own config. Guarding in `redact_conn_str` (mirroring `hide_secrets_in`) would fix this call site and any future one.

---

src/harlequin/hsql/session.py
```diff
@@ -129,6 +137,22 @@ def without_session_option(argv: "Sequence[str]") -> "list[str]":
...
+    An option *value* that is the literal string is the same ambiguity
+    `requested_session()` has, and fails as visibly: the whole argv travels
+    with the ask, and the session refuses one carrying anything else.
+    """
+    for argument in argv:
+        if argument == "--":
+            break
+        if argument == STATUS_OPTION:
+            return True
+    return False
```
tconbeer, line 154: **`--session-status=1` isn't recognized, so the one flag that must never wait, waits.**

This scan is strictly weaker than its two siblings — `requested_session()` (:87) and `_reads_stdin()` (`client.py:285`) both handle the `=` form. Here, `--session-status=1` falls through, so the invocation is packed as a **REQUEST** and takes a turn at the connection:

```
$ hsql --session rev --session-status=1
Error: Option '--session-status' does not take a value.   exit=2
```

Against an *idle* session that's just a confusing exit. Against a **busy** one, the typo blocks for the length of the running query (or `--queue-timeout`) before producing that message — on the flag whose whole purpose, per this module and `server.py`, is that it "does not wait." Matching `argument.startswith(STATUS_OPTION + "=")` and letting the server refuse it by name keeps the behavior uniform.

Two doc inaccuracies while you're here:
- The value-skipping gap does fail visibly, as the docstring says, but names the wrong culprit: `hsql --session prod -c '--session-status'` reports `... takes nothing else; -c is for an invocation that runs` — `server.py:598` picks the *first* non-matching element, which for a false positive is always the flag whose value triggered the misread, never the value itself.
- This docstring doesn't mention the `--` break the body implements (`hsql --session x -- --session-status` becomes a `CONN_STR` and is refused by the identity check), and `pack_status`'s docstring (`protocol.py:231`) says the ask carries "the argv the caller typed" when `client.py:219` passes it through `without_session_option`.

---

src/harlequin/hsql/server.py
```diff
@@ -248,6 +278,67 @@ def requests(self) -> int:
...
+            "pid": os.getpid(),
+            "version": protocol.VERSION,
+            "adapter": identity.pop("adapter", self.adapter),
+            "connection": " ".join(redact_conn_str(list(conn_str))) or None,
+            "connection_options": redact_profile(identity, self._adapter_options),
+            "uptime_s": round(time.monotonic() - self._started, 1),
+            "requests": self._requests,
+            "state": self._state(busy),
+            "queued": self._turnstile.queued,
+            "transaction_mode": None if busy else self._transaction_mode(),
```
tconbeer: **Check-then-act: `busy` is sampled under the lock, then released before it's acted on — so the invariant the docstring states absolutely isn't enforced.**

`Turnstile.busy` (:108-112) acquires and releases `self._condition`. Between that read at :300 and `_transaction_mode()` at :312 there's ~6-10 µs of `redact_conn_str`/`redact_profile`/dict-building (measured), which a GIL preemption widens to milliseconds. Nothing stops a request entering the turnstile in that window, so `connection.transaction_mode` can be read from the status thread while a query runs on that connection on another. With an instrumented fake connection (2 query threads, 2 status threads, 5 s):

```
transaction_mode reads: 296
overlapping entries into the connection: 331
```

The docstring at :284-289 says "nothing here may touch the connection while another thread holds it." Benign for every adapter in tree — duckdb inherits the base `return None`, sqlite returns a plain attribute — but `HarlequinConnection.transaction_mode` is public API documented as "the user-facing label of the currently-active transaction mode," and nothing forbids an implementation that asks the driver. That's precisely the case this guard exists for.

Separately, `busy` (:300) and `queued` (:311) are two independent acquisitions, so one document can report `"state":"idle","queued":1` — a session described as idle with requests in line, which is what a monitoring script branches on. Measured contradiction rate: 0.3% at 50 ms requests, 44% under a burst — i.e. it misfires exactly when someone asks "is this session saturated?"

One `Turnstile` helper that samples `(busy, queued)` and runs the mode read under a single acquisition fixes both.

---

tests/unit_tests/test_hsql_serve.py
```diff
@@ -681,13 +792,170 @@ def test_a_config_file_where_the_client_is_applies_its_per_request_keys(
...
+) -> None:
+    """The reason a status is not a request: it takes no turn at the
+    connection, so "is it hung or is it slow" has an answer while the query
+    that raised the question is still running."""
+    proc = send(["--session-status"])
+    assert proc.returncode == ExitCode.OK
+    status = json.loads(proc.stdout)
+    assert status["state"] == "busy"
+    # a busy session does not ask its driver a question on a second thread
+    assert status["transaction_mode"] is None
```
tconbeer: **This assertion is vacuous — delete the busy short-circuit at `server.py:312` and it still passes.**

`HarlequinConnection.transaction_mode` returns `None` in the base class (`adapter.py:203-212`) and `harlequin_duckdb` doesn't override it, so an **idle** DuckDB session reports `None` too. The comment says "a busy session does not ask its driver a question on a second thread," but nothing here distinguishes "suppressed because busy" from "the adapter has no transaction mode." Pinning it needs an adapter that returns non-`None` (SQLite on 3.12+) or an in-process `Server` with a stub connection.

Two more things about this test, which is the PR's headline claim and so worth the scrutiny:

**It doesn't pin "while a query is running."** The `blocked` fixture (:1031-1076) stalls in the FIFO read of its `-f` script — inside `Server._run`, but before any SQL is parsed or executed. So what's pinned is "busy while a request holds the turnstile," which is weaker than what `cli.py:524` and the module docstring both promise. The distinction is the one failure mode a caller cares about: a driver call holding the GIL or a process-wide lock would serialize the status answer, and a FIFO read (which releases the GIL) can't detect that. There's already a real multi-second query in this file for it — `select count(*) from range(2000000000)` at :1131.

**A regression here hangs rather than fails.** If status started taking a turn, `hsql_subprocess` passes no `timeout` to `subprocess.run`, `queue_timeout` defaults to `None`, and no `pytest-timeout` is configured — so CI would hang instead of reporting.

To be clear, the concurrency itself is real, not faked: a second process asks over the real socket while a real request thread holds the turnstile, and the test would fail if status entered the turnstile. It's the two assertions on top that don't hold up.

---

tests/unit_tests/test_hsql_serve.py
```diff
@@ -238,27 +277,74 @@ def test_session_is_a_profile_key_the_ide_leaves_alone() -> None:
...
+    request that asserted a different one would otherwise run on the session's
+    connection while believing it had changed it. The refusal names both
+    values, because what the session has is the half a caller cannot see."""
     res = hsql(*args, "-c", "select 1", obj=served_by(in_process_server))
     assert res.exit_code == ExitCode.USAGE
     assert res.stdout == ""
-    assert "is a connection option, and the session named 'inproc' connected" in (
+    assert f"says {differs}, and the session named 'inproc' connected with" in (
         res.stderr
     )
```
tconbeer: **The docstring says "names both values"; the assertion stops one word before the second one.**

`differs` is only the *asked* half, and the match ends at `connected with`. The real messages:

```
--read-only says True, and the session named 'inproc' connected with False.
CONN_STR says ['other.db'], and ... connected with [':memory:'].
--adapter says 'sqlite', and ... connected with 'duckdb'.
```

Only the adapter case has its second half asserted anywhere (:955). A regression that printed `None`, `''`, or echoed the asked value back for `theirs` would keep `read_only` and `conn_str` green — and per the finding on `_comparable`, "printing the same value on both sides" is a live failure mode, not a hypothetical one. This is the parametrized test for the claim, and the claim's distinctive half is the untested one.

Same shape at :375-395 for the profile wording (`f"the profile 'prod' sets {key} to"`), though it matters less there.

---

src/harlequin/hsql/server.py
```diff
@@ -248,6 +278,67 @@ def requests(self) -> int:
...
+            "ssh": self._ssh,
+            # `--idle-timeout` and `--max-lifetime` are not built yet, and a
+            # key that appears in a later release is worse for a caller
+            # reading this than one that is null until it can be answered
```
tconbeer: Per AGENTS.md, this is the case the conventions name twice: it records a tradeoff ("X is worse for a caller reading this than Y") and it points at code that doesn't exist yet — the comment equivalent of "keep this until #123 lands." It'll read as stale the day PR 5 lands.

```python
# null until --idle-timeout and --max-lifetime can answer them
```

The rationale for the shape belongs in `docs/plans/hsql-warm-sessions-design.md`, where this PR has already put it.

---

src/harlequin/hsql/protocol.py
```diff
@@ -51,15 +51,33 @@
...
 
+STATUS = 6
+"""Client to server: what the session is doing, as its own argv.
+
+Not a request, and so not a turn at the connection: a status answered by the
+same command the cold path builds would have to wait for the query it is being
+asked about, and "is it hung or is it slow" is a question with no use once the
+query is over. So the server answers this one off its own bookkeeping, while a
+request runs.
+"""
```
tconbeer: One design argument — *"a status must be answerable mid-query, or it answers nothing worth asking"* — is restated **six times** across this diff:

- `server.py:16-19` (module docstring)
- `server.py:284-285` (`status()`)
- `server.py:589-591` (`_answer_status`)
- here, `protocol.py:57-61`
- `session.py:34-35` (`STATUS_OPTION`)
- `client.py:213-215` (inline)

plus a seventh restatement in `diagnostics.py:156`'s one-line docstring ("which took no turn to be answered"). It already has a home — `docs/plans/hsql-warm-sessions-design.md` §4.5 and the PR-3 note this diff adds to it — and AGENTS.md sends it there: "never relay the design behind a change, the alternatives weighed, or the instructions the author was working from."

For this constant specifically, every neighbouring frame kind is one line (`"""Server to client, last: the code the client exits with."""`). Suggest:

```python
STATUS = 6
"""Client to server: what the session is doing, as its own argv.

Not a request and so not a turn at the connection: the server answers it off
its own bookkeeping, while a request runs.
"""
```

The same trim applies to the other five. Two of them additionally argue against a rejected alternative, which the conventions rule out outright:

- `server.py:24-26` and `cli.py:1479-1481` carry the *identical* sentence — "Not a reconnect -- that is a different session -- and not silence, which lies to a caller who typed `-a postgres` and got DuckDB." Drop it from both.
- `server.py:589` opens "Written here rather than run through the command, because…" The one fact there that isn't elsewhere and *is* load-bearing is the shared-process-state constraint: "A served invocation borrows this process's cwd, environment and streams for its whole run, so this one is answered here rather than parsed beside it."

And `server.py:326-331` (`_transaction_mode`) explains neither what the function does nor why it's written that way — "The sharp edge of a session being state: a caller who ran `BEGIN` and exited…" is commentary on session semantics attached to a getter. The summary line alone is enough; the `# adapters are third-party code` inline below it is exemplary and should stay.

`_adapter_options`' `"""What the adapter declares, so that `secret=` decides what prints."""` (:259) is the model the rest of the diff should follow.

---

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 ### Features
 
 - `hsql --serve NAME [CONN_STR]` holds a database connection open as a named session, and `hsql --session NAME -c ...` (or `HSQL_SESSION=NAME`) sends queries to it without paying start-up or connection cost again. Temp tables, settings and open transactions persist between invocations; `--session-reset` reconnects. POSIX only.
+- `hsql --session NAME --session-status` reports what a session is doing -- its adapter, connection, uptime, request count and whether it is busy -- as JSON, and answers while a query is running. A request that names a different database, adapter or connection option than the session connected with is now refused rather than run on the session's connection.
```
tconbeer: Three things, per `.claude/rules/changelog.md`:

1. **"is now refused rather than run on the session's connection"** describes a fix to behavior nobody has seen. `--serve` itself is unreleased — it's the entry directly above, in this same `[Unreleased]` block — so no released hsql ever ran a mismatched request on a session's connection. "now" and the contrast both imply a regression being corrected.
2. **The field list is a schema dump.** "its adapter, connection, uptime, request count and whether it is busy" is more than a user needs to decide whether they care.
3. **Dash convention:** the file's prevailing dash for a prose aside is `—` (lines 51, 201, 260); `--` appears once, at 286.

Suggested:

```markdown
- `hsql --session NAME --session-status` reports what a session is doing as JSON — adapter, connection, uptime, requests, busy or idle — and answers even while a query is running. A request that names a different database, adapter or connection option than the session connected with is refused.
```

Omitting the issue link is right — I searched for an issue covering warm sessions or session status and found none, and the sibling `--serve` entry carries none either.

---

tconbeer (commented):
## Scope of this review

Five narrow passes, each with the others' lenses excluded: (1) identity recording and served-request rejection in `cli.py`/`config.py`; (2) `--session-status` concurrency, protocol and env forwarding in `server.py`/`protocol.py`/`client.py`/`session.py`; (3) whether the tests' *assertions* pin the claims; (4) docstring/comment/CHANGELOG conventions; (5) security, which I did myself.

Claims were verified empirically — live `hsql --serve` sessions over real sockets, in-process `Server` instances with instrumented fake connections, and A/B runs against an `origin/main` worktree. `test_hsql_serve.py` + `test_hsql_session.py`: **198 passed**.

The design is sound and the two central decisions are right: a status genuinely cannot be a request (a served invocation owns the process's cwd, environment and streams for its whole run), and comparing identity key-by-key rather than refusing wholesale is what makes `-P` usable. The concurrency in the headline test is real, not faked. What follows is about the seams.

## Blocking

**1. A typed `-P` is refused whenever its profile holds any IDE key.** `TUI_ONLY_KEYS` are popped from `values` before the identity snapshot, but the request side builds `profile_keys` from the raw `profile_config` — and all seven (`theme`, `locale`, `keymap_name`, `show_files`, `show_s3`, `no_download_tzdata`, `viewer_max_rows`) classify as connection options. So `hsql --serve prod -P prod` followed by `hsql --session prod -P prod -c ...` is refused with *"the profile 'prod' sets theme, a connection option the session named 'prod' did not ask for."* That's the session's own profile. Since `TUI_ONLY_KEYS`' docstring says one profile serves both commands, a profile with a `theme` is the normal case — which makes this the PR's headline feature not working on the input it was built for. The existing test passes only because its fixture profile carries no IDE key.

**2. `_answer_status` has no exception net.** Anything `status()` raises escapes the thread, and `threading.excepthook` writes to a `sys.stderr` that `_as` has swapped to the **in-flight request's** recorder — so a server-side traceback with absolute paths is delivered into an unrelated caller's terminal, and the status client gets exit 3 ("the database cannot be reached") for an hsql bug where the request path gives 70.

## Worth fixing before merge

**3. Regression:** a typed profile setting `queue_timeout` is no longer refused — silently dropped. `main` refused it; verified A/B. The docstring still says server-lifetime options are "refused outright."

**4.** A scalar-string `conn_str` in a profile never matches a tuple, and the message then prints *the same value on both sides*: `CONN_STR says [':memory:'], and the session ... connected with [':memory:']`.

**5.** A non-string `conn_str` in a profile (`conn_str = 5`) crashes `redact_profile` → exit 70 and a bug-report prompt, where a usage error is owed.

**6.** `--session-status=1` isn't recognized by the argv scan, so it's packed as a request and **blocks behind a running query** — on the one flag documented as never waiting.

**7. Two assertions in the headline test don't hold up.** `assert status["transaction_mode"] is None` is vacuous (DuckDB returns `None` idle too, so the busy short-circuit could be deleted). And the `blocked` fixture stalls in a FIFO read before any SQL executes, so "while a query is running" isn't what's pinned. A regression here would *hang* CI, not fail it — no timeout anywhere in that path.

## Non-blocking

Check-then-act on `Turnstile.busy` (measured: 331 overlapping entries into a fake connection under load) and `busy`/`queued` sampled separately, so one document can say `"state":"idle","queued":1`; one `Turnstile` snapshot fixes both. Comment sprawl — one design argument restated six times, plus a comment pointing at unbuilt PR-5 code, both against AGENTS.md; concrete replacements inline. CHANGELOG describes a fix to unreleased behavior. Test gaps: the `secret=True` declaration path is never exercised (only the name backstop), no adapter-declared option is tested on the *refusal* side, and the `not known` profile-worded branch is unreachable in the current parametrize.

**Security: clean.** Forwarded `HARLEQUIN_CONFIG_PATH` stays inside the existing same-uid socket boundary, is set/popped/restored correctly per request (verified live in both directions), and `SSH_KEYS` holds no credential, so the unredacted `ssh` field is safe. DSN and declared-secret masking both work in the status document.

## Not checked

- I did not review PRs #1132 or #1138, on which this builds; I took `Turnstile`, `Recorder`, `_as` and the socket/uid model as given except where this diff's new code depends on them.
- I did not run the full suite, the functional/snapshot tests, mypy, `lint-imports`, or the cold-start budget — I ran only the two changed test files. CI covers the rest and is green apart from the item below.
- I did not verify the client's module-count budget claim, or that `docs/generated/hsql-reference.md` matches its generator.
- I did not test on macOS or Windows, or with any out-of-tree adapter — the adapter-option findings are reasoned from `harlequin_duckdb`/`harlequin_sqlite` plus the ABC, and spot-checked against `harlequin-postgres`' published `transaction_mode`.
- I did not audit `harlequin.redact` beyond the paths this diff calls, nor re-derive whether `_SECRET_NAME` covers what adapters actually use.
- I did not review the `docs/plans/` prose for accuracy against the implementation, beyond noting that §4.10's example shows `idle_timeout_s`/`expires_in_s` populated where the code returns null, and shows a `read_only` key that a session which never typed `--read-only` won't record.

**CI:** one failure, `test_ssh_recovery.py::test_a_refresh_that_reconnects_does_not_ask_for_a_second_one` on macOS 3.13 only (`ValueError: No Tab with id 'tab-1'`), passing on the other 13 legs. It's TUI/SSH-recovery code this diff doesn't touch and reads as an unrelated flake — but I did not confirm that against `main`.

A clean review is necessary, not sufficient: whether the identity semantics are the ones you want is a judgment I can't make from the diff. The merge call is yours.

---

src/harlequin/hsql/cli.py
tconbeer: Confirmed and fixed in `be0f2b0`. All seven classified `True`, exactly as you show.

Fixed in `_is_connection_option`, as you suggested, rather than by subtracting at the call site — the two halves have to agree on the question, not on one answer to it. There is now a `NOT_AN_ADAPTERS_OPTION` frozenset carrying `TUI_ONLY_KEYS` alongside the four groups, so the identity snapshot and the request-side comparison read the same rule.

Two tests: `test_the_ides_own_profile_keys_are_not_connection_options` pins the classification, and `test_a_served_request_takes_a_profile_the_ide_also_reads` runs the end-to-end case over `theme`, `locale` and `viewer_max_rows` — the fixture profile now carries an IDE key, so the gap that let this through is closed.

---

src/harlequin/hsql/server.py
tconbeer: Fixed in `be0f2b0`. The cross-talk is the part that makes this worse than a missing net, and I had not seen it.

`_answer_status` now wraps the whole answer the way `_run` is wrapped: the note goes to the server's own stderr, the client gets `report_crash` into its own recorder and exit 70. Extracted `_status_into()` so the `try` covers the document and the argv refusal both. `_crash_report()` now takes an argv rather than a `protocol.Request`, since a status has no request — it only ever used `request.argv`.

`test_a_bug_in_the_status_path_is_exit_70_and_not_someone_elses_stderr` pins it over a socketpair with `status()` patched to raise: last frame is `EXIT` 70, stderr carries the crash line, and no `Traceback`.

---

src/harlequin/hsql/cli.py
tconbeer: A real regression, and you're right that the docstring was asserting the thing I'd just removed. Fixed in `be0f2b0`.

The function now builds `from_profile` once and refuses server-lifetime keys from both sources before the `connects` return, so the rule no longer depends on where the value came from or on whether the mode connects. The profile wording gets its own message naming the key and pointing at `--serve NAME -P`.

That also settles the follow-on you named: the refusal runs long before `_timeout_seconds()`, so `queue_timeout = "soon"` is now refused as a profile key rather than failing the request under a flag it could not have used.

`test_a_served_request_may_not_take_a_server_option_from_a_profile` covers it, beside the typed one.

---

src/harlequin/hsql/cli.py
tconbeer: Fixed in `be0f2b0`, and I took the second half of your suggestion too — it's the stronger fix.

`_comparable()` now takes the key and folds `conn_str` to a tuple from a string, list or tuple alike, matching what the callback does at :692-694 before it connects. And the request side now builds its own snapshot through `connection_options_in()` rather than reading `values` directly, so **the message prints the values the comparison used**. A message that disagrees with its own comparison is no longer expressible, rather than fixed case by case.

The parametrized refusal test now asserts both halves (see the thread on it), which is what would have caught the "two identical things differ" output.

---

src/harlequin/hsql/cli.py
tconbeer: Reproduced (`TypeError: expected string or bytes-like object`) and fixed in `redact_conn_str`, as you suggested — in `be0f2b0`.

An item that is not a string is masked whole rather than parsed for spans. That direction rather than passing it through: this is the last thing between a value and a terminal, `harlequin.crash` is one of its callers, and the module already documents over-redaction as the safe way to be wrong. It stays `list[str]`, so no caller's type changes.

The served request now takes the ordinary path: exit 2, `the profile 'bad' sets conn_str to '********', and the session named 'inproc' connected with [':memory:']`. `test_a_profile_value_of_the_wrong_shape_is_a_usage_error` covers that; `test_redact.py` covers `5`, `True`, `None`, `3.5`, and that one bad item does not take the valid DSNs beside it.

---

src/harlequin/hsql/session.py
tconbeer: Fixed in `be0f2b0` — the scan now matches `STATUS_OPTION + "="`, so it is no weaker than its two siblings. The server refuses the `=` form with `--session-status is a flag and takes no value`, which is what click says cold.

`test_a_mistyped_status_does_not_wait_for_the_query_it_asks_about` runs it against the `blocked` fixture with a 30s subprocess timeout, so a regression fails rather than hangs.

All three doc points taken:
- `asks_for_status`' docstring no longer claims the false positive names the value; it now says the whole invocation travels with the ask and the session refuses one carrying anything else.
- The `--` break is in the docstring and the body carries the same `# everything after it is an argument, not an option` comment `requested_session()` uses.
- `pack_status` now says "the invocation that carried it" rather than "the argv the caller typed."

---

src/harlequin/hsql/server.py
tconbeer: Both fixed in `be0f2b0`. `Turnstile.snapshot()` returns `(busy, queued)` under one acquisition, so the `"state":"idle","queued":1` document is no longer expressible.

For the mode read I did not hold `_condition` across the call — that would put third-party code inside the lock `enter()`/`leave()` need, trading a race for a possible wedge. Instead `_transaction_mode()` **takes a turn**: `self._turnstile.enter(0)`, read, `leave()`. `enter(0)` is already a non-blocking try-acquire, so a busy session still reports `null` without waiting, and an idle one holds the connection only for the length of the adapter's own call — the same exposure any request has. The docstring now says that rather than the absolute claim it couldn't back.

`test_a_busy_session_does_not_ask_its_driver_for_the_mode` pins it with a connection that counts reads: idle reports `Manual`, and with the turnstile held the read count does not move — not merely that the field is null.

---

tests/unit_tests/test_hsql_serve.py
tconbeer: All three taken, in `be0f2b0`.

**The vacuous assertion** is gone, replaced by `test_a_busy_session_does_not_ask_its_driver_for_the_mode` — an in-process `Server` over a stand-in connection that counts reads and returns `HarlequinTransactionMode(label="Manual")`. Idle reports `Manual`; with the turnstile held the count does not move. Deleting the short-circuit now fails it.

**"While a query is running"** is now two tests rather than one overloaded one. The FIFO test keeps the deterministic claim under the name it actually pins — `test_a_status_is_answered_while_a_request_holds_the_connection` — and a new `test_a_status_is_answered_while_the_database_is_working` runs your `select count(*) from range(2000000000)` under `--timeout 20` and polls until the session reports busy, which is the case a GIL-holding driver call would break. `--timeout` rather than a kill, so the session is free again for teardown.

**The hang** is fixed at the source: `hsql_subprocess` now takes a `timeout` that goes to `subprocess.run`, and every status call that asserts an invocation *returns* passes one. A regression reports instead of stalling the leg.

---

tests/unit_tests/test_hsql_serve.py
tconbeer: Right — and given the `_comparable` finding, the untested half was the one actively broken. Fixed in `be0f2b0`.

The parametrize now carries `spelling`, `asked` and `has`, and asserts the whole clause including the closing period:

```python
assert (
    f"{spelling} says {asked}, and the session named 'inproc' connected with {has}."
) in res.stderr
```

I also added a fourth case, `--init-path`, so an **adapter-declared** option is exercised on the refusal side rather than only hsql's own — which turned up that a `PathOption` was printing `PosixPath('/p/boot.sql')` at the caller. There is now a `_shown()` that renders a `Path` as its path.

The profile test gets the same treatment: `has` is a parameter and the second half is asserted.

---

src/harlequin/hsql/server.py
tconbeer: Taken verbatim in `be0f2b0` — it's now the one line you wrote. The reasoning stays in `docs/plans/hsql-warm-sessions-design.md`, where this PR's §9 note already had it.

---

src/harlequin/hsql/protocol.py
tconbeer: Six times is indefensible. Trimmed all of them in `be0f2b0`.

- `protocol.py` `STATUS`: your suggested two-liner verbatim.
- `server.py` module docstring: the argument goes, the fact stays — "`--session-status` is the one thing that does not wait: it is its own frame rather than an invocation."
- `status()`: the whole first paragraph goes; the `secret=` sentence is what a reader of that function needs.
- `_answer_status`: cut to the load-bearing constraint, close to your wording — "A served invocation borrows this process's cwd, environment and streams for its whole run, so this one is answered here rather than parsed beside it."
- `STATUS_OPTION`: two lines.
- `client.py`: the inline is now `# not a request, so it takes no turn at the session's connection`.
- `diagnostics.report_status_answered`: rewritten anyway (it now takes the exit code, so a status logs like a request does) and the clause is gone.

The rejected-alternative sentence is deleted from **both** `server.py:24-26` and `cli.py`, as you asked.

`_transaction_mode` is now a summary line and the mechanism it needs (it takes a turn, per the `Turnstile` thread); the session-semantics commentary is gone. `# adapters are third-party code` stays.

---

CHANGELOG.md
tconbeer: Your text taken verbatim in `be0f2b0`. All three points are right — the "now" was the worst of them, since it implies a regression against a `--serve` that has never shipped.

---

tconbeer:
All twelve threads addressed in `be0f2b0` and answered inline. Both blocking findings were real and neither had a test that would have caught it; thank you for the reproductions, which is what made them cheap to fix.

**The two blocking ones**

`TUI_ONLY_KEYS` classifying as connection options was the worst of it — the feature not working on the input it was built for. Fixed in `_is_connection_option` rather than at the call site, so the identity snapshot and the request-side comparison read the same rule. `_answer_status` now has the same net `_run` has: exit 70 with a crash report, rather than a traceback out of the thread and into whichever request's recorder happens to hold the process's stderr.

**Two of the fixes are broader than the finding**, on your suggestion:

- The request side now builds its snapshot through `connection_options_in()` too, so the message prints the values the comparison used. "These two identical things differ" is no longer expressible rather than fixed for `conn_str`.
- `redact_conn_str` masks an item it cannot parse rather than raising. It is the last thing between a value and a terminal and `harlequin.crash` calls it, so it should not be able to raise on the way past.

**Where I did something other than what was suggested**, once: for the `transaction_mode` race I did not hold `_condition` across the adapter call — that puts third-party code inside the lock `enter()`/`leave()` need, trading a race for a possible wedge. It takes a zero-timeout *turn* instead, which is a non-blocking try-acquire, so a busy session still reports null without waiting and an idle one holds the connection only as long as the adapter's own call. `Turnstile.snapshot()` fixes the `busy`/`queued` half as you described.

**Tests.** Both values are now asserted in every refusal; the busy short-circuit is pinned against a connection that counts reads and reports a mode, so deleting it fails; "while a query is running" is two tests — the FIFO one under the name it actually pins, plus a new one that polls while duckdb executes `select count(*) from range(2000000000)` under `--timeout`; and `hsql_subprocess` takes a `timeout`, so a status that started queueing fails rather than hanging the leg. New coverage for the profile server-option refusal, an IDE key in a profile, an adapter's `secret=` declaration on the refusal side, the profile-worded `did not ask for` branch, and a mistyped profile value.

**Docs.** Your CHANGELOG text taken verbatim. §4.10's example now shows null lifetime fields and a `read_only` that a session started with it would actually record — you were right that the sketch showed a document the code cannot produce.

**CI.** `test_ssh_recovery.py::test_a_refresh_that_reconnects_does_not_ask_for_a_second_one` on macOS 3.13: I checked `main`, which is green, so it is not red on base — but it is a TUI/SSH-recovery test this diff touches nothing in, it failed on 1 of 14 legs, and `1805 passed` beside it. This push re-runs the suite, which is the one re-run it gets; if it reproduces on the same leg I will treat it as real rather than re-running again.

Locally after the changes: 1823 passed / 7 skipped on 3.10, py12 subset green, mypy clean over 148 files, five import-linter contracts kept, ruff clean, both generated artifacts regenerated.

---

src/harlequin/hsql/cli.py
```diff
@@ -511,6 +517,15 @@ def build_cli(argv: Sequence[str]) -> click.Command:
             "transaction are gone. Needs --session."
         ),
     )
+    @click.option(
+        "--session-status",
+        is_flag=True,
+        help=(
+            "Ask the session what it is doing -- adapter, connection, uptime, "
+            "requests, whether it is busy -- as JSON, and exit. Answered "
+            "while a query is running. Needs --session."
```
tconbeer: "Poll the server for its status as JSON. Needs --session"

src/harlequin/hsql/cli.py
```diff
@@ -838,6 +863,16 @@ def inner_cli(
             )
 
         # every mode below this connects to the database
+        if session_status:
+            # the client answers this one before click exists, off a frame the
+            # session replies to while it is busy; reaching the parser means
+            # no session did
+            diagnostics.error(
+                "--session-status asks a running session what it is doing, and "
+                "no session answered this invocation. Run "
```
tconbeer: No "ask " and "answer". Be direct (formal?) in language and error messages. What should the user do here? Restart the server? Specify the name of a session?

src/harlequin/hsql/cli.py
```diff
@@ -1285,11 +1327,17 @@ def _serve(
...
+    ssh: str | None,
     queue_timeout: float | None,
 ) -> ExitCode:
     """Connect, and answer for the session called `name` until stopped.
 
-    The server reconnects for `--session-reset`, so it is handed the adapter.
+    The server reconnects for `--session-reset`, so it is handed the adapter;
+    it records what it connected with, so a request that asserts a connection
+    option can be told what it would have changed; and it reports what the
+    adapter declares secret, so `--session-status` prints no credential.
```
tconbeer: Sessions aren't people. They can't be told or answer. They don't have hands. Don't talk about what DOESNT get printed.

src/harlequin/hsql/cli.py
```diff
@@ -1323,16 +1374,30 @@ def _reset_session(ctx: click.Context, served: "Served") -> ExitCode:
...
+    | CONFIG_OPTIONS
+    | set(TUI_ONLY_KEYS)
+)
+"""Every key that reaches a profile and is not an adapter's.
+
+`TUI_ONLY_KEYS` are here because one profile serves both commands: they are
+the IDE's, hsql drops them before it connects, and a session records none of
+them -- so reading one as a connection option would refuse a request over a
+`theme`.
+"""
```
tconbeer: Delete this whole docstring. It's just long winded and confusing

tconbeer:
CI status, and one failure that is not this PR's.

**Windows — mine, fixed.** The `--init-path` case I added while addressing your review on both-values-asserted spelled POSIX path literals, so it failed on every Windows leg. `545f128` moved the asked value out of the assertion; that was still wrong, because `str(Path("/boot.sql"))` is `\boot.sql` on Windows. `4c567f4` takes the separator out entirely: the session records a bare `boot.sql`, and the asked value is asserted by its file name. Two speculative pushes where one validated one would have done — I should have checked the platform-dependence of a `Path` assertion before the first.

**macOS — not this PR's.** `test_ssh_recovery.py::test_a_refresh_that_reconnects_does_not_ask_for_a_second_one`, `ValueError: No Tab with id 'tab-1'`. **The same test with the same error fails on `main`**: run [33833836899](https://github.com/tconbeer/harlequin/actions/runs/33833836899) on `e278c0b`, macOS 3.10 and 3.11. On this branch it has hit macOS 3.13 (`600030a`), then macOS 3.10 + 3.11 (`be0f2b0`), then no leg at all (`545f128`) — a different set each time, which is a timing race rather than a break. The traceback is entirely in Textual's `Tabs` (`_tabs.py:497` `refresh_active` → `validate_active` at `:580`) reached from `EditorCollection.on_mount`, neither of which this diff touches.

No fix exists for me to port, so per the rules I am proposing rather than widening this PR. The race is that `Tabs.add_tab()` activates the first tab through `refresh_active()`, which sets `active` and validates it with `query("#tabs-list > #tab-1")` before the `Tab` mount has landed in the DOM. Two options, and I think the first is the one AGENTS.md points at:

1. **Upstream.** This is Textual's `Tabs`, and "when the real fix belongs upstream, make it upstream" — `add_tab`'s returned awaitable should not be able to complete with `active` set to a tab the DOM cannot yet find. Happy to write that up as an issue against Textual if you want it.
2. **Here, in the meantime.** The repo already runs `--force-flaky --max-runs=3` on Windows for exactly this class of thing; extending it to the macOS legs would stop this blocking PRs without skipping or quarantining anything. Your call — it is outside this PR either way.

I have not re-run any job; the two Windows pushes each triggered a fresh run, and the macOS legs went green on their own in the latest one. If it reappears I will leave it alone rather than re-running, since the `main` failure already settles the attribution.

The PR is watched with a check-in scheduled until it is green and mergeable.

---

src/harlequin/hsql/cli.py
```diff
@@ -1323,16 +1374,30 @@ def _reset_session(ctx: click.Context, served: "Served") -> ExitCode:
...
+
+
 def _is_connection_option(name: str) -> bool:
     """Whether an option is answered when a connection is opened.
```
tconbeer: This docstring is terrible too. Be clear about what this does. Do not talk about what it does not do or does not have.

src/harlequin/hsql/cli.py
```diff
@@ -1382,6 +1447,38 @@ def _refuse_unservable_name(ctx: click.Context, name: str) -> None:
         ctx.exit(code)
 
 
+def connection_options_in(values: Mapping[str, Any]) -> dict[str, Any]:
+    """The connection-time half of what an invocation asked for, normalized.
```
tconbeer: No "asked for"

src/harlequin/hsql/cli.py
```diff
@@ -1382,6 +1447,38 @@ def _refuse_unservable_name(ctx: click.Context, name: str) -> None:
...
+
+    A session records this when it connects and a served request builds the
+    same thing from its own values, so the two can be compared key by key.
```
tconbeer: I don't understand this docstring

src/harlequin/hsql/cli.py
```diff
@@ -1390,66 +1487,151 @@ def _refuse_the_sessions_options(
...
+    key: str,
+    asked: Any,
+    has: Any,
+    known: bool,
+    profile: str | None,
+    options: "Sequence[AbstractOption] | None",
+) -> None:
+    """Exit having named the connection option that differs, and both values.
+
+    Both are masked by `harlequin.redact` before they are printed: a
```
tconbeer: You don't need to explain why.

tconbeer:
CI is green on `4c567f4` — all 15 Test legs, Test Build, and Static Analysis. The Windows failure was mine and is fixed; the three macOS legs passed on the one re-run, which is the attribution the `main` failure already established. Nothing outstanding from my side.

---

tconbeer (commented):
Docstrings are terrible. Fix my specific comments and apply the same critique to the others. 

Generally I'm worried the option parsing is becoming too complex and brittle. We will be adding more options. What then? They will need to be added tot he right semantic groups for hsql and harlequin etc. If we really think the current solution is the best that we can do, then I need you to write up new documentation on CLI options and the process for adding an option (either as a mode or a connection option, or whatever the other categories are).

src/harlequin/hsql/cli.py
tconbeer: Taken, close to verbatim in `efd7773`:

```
Poll the server for its status as JSON, and exit. Reports while a query is
running. Needs --session.
```

I kept the second sentence because "reports while a query is running" is the reason to reach for it over reading the server's log, and a caller can't infer it. Dropped the field list — `--session-status` prints them.

---

src/harlequin/hsql/cli.py
tconbeer: Fixed in `efd7773`. The old message described the mechanism and left the caller nowhere to go. Both refusals now name the remedy:

```
hsql: error: --session-status reports on a running session. Name one with
--session NAME, or set HSQL_SESSION, and start it with 'hsql --serve NAME ...'.

hsql: error: --session-reset reconnects a running session. Name one with
--session NAME, or set HSQL_SESSION.
```

Both answers to your question, in order: name a session, and start one if none is running. `--session-reset` gets only the first, since reaching it means a session was named but the client fell back — starting a server is not what that caller needs to hear.

The "ask/answer" vocabulary is gone from the whole diff, not just here: `asks_for_status` → `requests_status`, `_answer_status` → `_send_status`, `report_status_answered` → `report_status`, and the docstrings that had sessions answering questions now say what the code does.

---

src/harlequin/hsql/cli.py
tconbeer: Fixed in `efd7773`. `_serve`'s docstring was three clauses about what the server would later do with each argument, two of them phrased as the session being told things. It is now one line:

```python
"""Connect, and serve the session called `name` until the server stops."""
```

The "don't talk about what doesn't get printed" note applied to `_refuse_a_differing_option` too, which opened by explaining that the values are masked *so that* nothing prints a password. That is now one sentence stating what happens: "Values are masked by `harlequin.redact` before they print."

---

src/harlequin/hsql/cli.py
tconbeer: Deleted — along with the constant it documented. In `efd7773`, `NOT_AN_ADAPTERS_OPTION` and `_is_connection_option()` are both gone; see the thread on line 1394 for what replaced them.

---

src/harlequin/hsql/cli.py
tconbeer: The docstring was bad because the function was. It was defined by exclusion — "an option the command does not know is an adapter's" — so it could only be described by what it does not do, and the `TUI_ONLY_KEYS` bug you found earlier was that definition failing on the first key nobody declared.

Replaced in `efd7773` with a positive one:

```python
def connection_option_names(
    options: "Sequence[AbstractOption] | None",
) -> frozenset[str]:
    """The keys that describe a connection: hsql's own, plus one adapter's.

    A positive test, so only a key one of the two declares is compared
    against a session.
    """
    return CONNECTION_OPTIONS | frozenset(
        sluggify_option_name(option.name) for option in options or []
    )
```

That deletes `NOT_AN_ADAPTERS_OPTION`, deletes `_is_connection_option`, and deletes the `TUI_ONLY_KEYS` special case with them — an IDE key is now not a connection option because nothing declared it, rather than because it was listed as an exception. The two call sites become set intersections.

---

src/harlequin/hsql/cli.py
tconbeer: Gone in `efd7773`, from the docstring and from the local it named — `asked` is now `requested`, and the parameter on `_refuse_a_differing_option` with it. The user-facing half changed too: the refusal said the session "did not ask for" an option, and now says it "was started without it".

---

src/harlequin/hsql/cli.py
tconbeer: Fair — reading it back, the second paragraph was arguing with an objection nobody raised. Rewritten in `efd7773`:

```python
def connection_options_in(
    values: Mapping[str, Any], connection_keys: "frozenset[str]"
) -> dict[str, Any]:
    """The connection options in `values`, normalized for comparison.

    `--serve` records this; a served request builds it from its own values,
    and the two are compared key by key. It holds the keys a profile or a
    command line named. A key named by neither takes its value from the
    adapter's own default, which core cannot enumerate.
    """
```

First line says what it returns; second says who uses it; third and fourth are the one fact a reader needs and cannot see from the code — that the dict is sparse on purpose, so a missing key means "nobody named it" rather than "it has no value."

---

src/harlequin/hsql/cli.py
tconbeer: Cut in `efd7773`. `_refuse_session_keys_from_a_profile` is now one line:

```python
"""Refuse a config file that sets which process runs an invocation."""
```

The explanation of why only a profile can reach it went with it. Its message also lost the anthropomorphism — `{key} decides which process runs an invocation` rather than "says which process answers" one.

Same treatment on the neighbours: `_refuse_per_request_options`, `_refuse_unservable_name`, `_reset_session`, `_shown`, `_comparable` and `_refuse_the_sessions_options` all lost their why-paragraphs. The five group constants too, which were all phrased as questions being "answered".

---

tconbeer:
Both halves addressed in `efd7773`.

## The docstrings

Eight threads answered inline, and the critique applied across the diff rather than only at the lines you marked. What I took the rules to be, so you can check I read them right: state what the code does; never describe what it does not do or does not have; sessions and servers do not ask, answer, or get told; do not explain why; and an error message ends with what to do about it.

The vocabulary is gone from the code too, not just the prose — `asks_for_status` → `requests_status`, `_answer_status` → `_send_status`, `report_status_answered` → `report_status`, and the local `asked` → `requested`. Every user-facing message this PR adds was rewritten; the two you flagged now read:

```
hsql: error: --session-status reports on a running session. Name one with
--session NAME, or set HSQL_SESSION, and start it with 'hsql --serve NAME ...'.

hsql: error: --adapter is 'sqlite'. The session named 'prod' was started with
'duckdb', and its connection is fixed. Drop it here, or start a session with
it: 'hsql --serve NAME --adapter ...'.
```

## The option groups

You gave me a conditional — simplify, or document — and I think the honest answer is *both*, because one part of the complaint has a real fix and the other does not.

**The part that was fixable was the classification rule, and it is the part that already bit us.** `_is_connection_option()` worked by exclusion: an option in none of the other groups was an adapter's, and therefore connection-time. That is why `theme` was compared against a session, and it would have gone wrong again for any future profile key no command declares. It is now positive:

```python
def connection_option_names(options) -> frozenset[str]:
    """The keys that describe a connection: hsql's own, plus one adapter's."""
    return CONNECTION_OPTIONS | frozenset(
        sluggify_option_name(option.name) for option in options or []
    )
```

That deletes `NOT_AN_ADAPTERS_OPTION`, deletes `_is_connection_option`, and deletes the `TUI_ONLY_KEYS` exception with them. A key nothing declares is now not a connection option *because nothing declared it*, which needs no list to maintain.

**The part I did not fix is the one your question is really about: "they will need to be added to the right semantic groups — what then?"** After the above, adding an `hsql` option is: write the `@click.option`, put its name in one of five frozensets, and if it is a mode, add it to `_one_mode()` and the first-pass predicates. Forgetting the frozenset is already a test failure — `test_every_option_is_in_exactly_one_group` asserts the groups partition the command's declared parameters — so it fails loudly rather than silently. But it is still a list beside the declaration rather than something on it, and that is a real weakness.

The alternative is carrying the group on the `click.Parameter`, so an option cannot be declared without one and the frozensets disappear. That is the right end state and I did not do it here: it touches both commands and every option in them, and this PR is already four commits past its first review. **If you want it, I would rather do it as its own PR** where the diff is only that.

So, the documentation you asked for: **`docs/cli-options.md`**, linked from AGENTS.md's options section and from CONTRIBUTING.md. It covers where options come from (adapter vs. command), what each of the five groups means and how `--serve` and a served request treat it, why `CONFIG_OPTIONS` are judged by what a profile holds rather than by being a profile, and an eight-step checklist for adding one — including the artifacts to regenerate and the test that will fail if the group is missed. It ends with a "Known weakness" section stating the drift problem and the alternative above, so the next person meets it as a documented tradeoff rather than as a surprise.

Full suite green: 1823 passed / 7 skipped on 3.10, mypy clean, five import-linter contracts kept, `docs/generated/hsql-reference.md` regenerated for the new help text.

---

tconbeer:
CI on `efd7773` turned up a second flake, this one in scope. Fixed in `2c17beb`.

`test_a_timeout_stops_the_request_and_leaves_the_session_up` runs `select count(*) from range(2000000000)` under `--timeout 0.1` and asserts empty stdout. On an arm64 macOS runner duckdb won that race and returned `2000000000`, so the test saw a result set where it wanted none. It came in with `--serve` in #1138 and it has been a race since; this is the first runner fast enough to show it.

The query is now 100× longer and lives in one constant shared with the status test that also needs a long one, so neither depends on how fast the runner is. The status test's own `--timeout` drops from 20s to 5s at the same time — it only has to outlive a poll that returns on the first busy status, and the 20s was being paid in full at teardown while the session waited for the query to be cancelled.

The other two macOS legs on that run were `test_ssh_recovery`'s `No Tab with id 'tab-1'` again — the one I established fails on `main` too, and which I've already commented on and spent the re-run for. Not commenting on it again.

Locally after the fix: 1823 passed / 7 skipped, mypy and import-linter clean.

---

CONTRIBUTING.md
```diff
@@ -21,6 +21,10 @@ Code for the [harlequin.sh](https://harlequin.sh/) website is hosted in the [`ha
 
 We aim to create a welcoming and inclusive environment for new contributors, whether this is your first open-source contribution or your 1,000th. Please see the [Contributing Guide](https://harlequin.sh/docs/contributing/index) for detailed instructions on local development and testing, including an end-to-end walkthrough [video](https://www.tella.tv/video/cls3nmne700000gl4bcp91arr/view).
 
+## Adding a CLI Option
```
tconbeer: Remove this from this document

docs/cli-options.md
```diff
@@ -0,0 +1,105 @@
...
+Every option that is not `CONN_STR` is also a **profile key**, under the name
+click gives it (`--read-only` is `read_only`). `config.py` validates a profile
+against the command's parameters plus the adapter's declared options, so a key
+neither declares is an error.
+
+## The five groups
+
+`hsql` runs in two roles — a server (`--serve NAME`) and a client
+(`--session NAME`) — and an option belongs to exactly one group, decided by
+**when the question it answers can be answered**. The groups are frozensets at
```
tconbeer: No questions or answers. Write directly.

docs/cli-options.md
```diff
@@ -0,0 +1,105 @@
...
+
+   ```bash
+   uv run python scripts/write_config_schema.py
+   uv run python scripts/write_cli_reference.py
+   ```
+
+8. **Add a `CHANGELOG.md` entry** under `[Unreleased]` if the option is
+   user-facing.
+
+## Known weakness
```
tconbeer, line 95: Open an issue to fix this. I would think the solution would be a click parameter subclass with additional properties? Generally I would prefer a simpler declaration for each option wrt how and where it can be used.

src/harlequin/hsql/protocol.py
```diff
@@ -51,15 +51,27 @@
 EXIT = 5
 """Server to client, last: the code the client exits with."""
 
+STATUS = 6
+"""Client to server: report the server's status, as its own argv.
+
+Not a request, and so not a turn at the connection: the server reports from
```
tconbeer: Delete this last sentence

docs/cli-options.md
tconbeer: Rewritten in `534a9c7`. The group is now decided by **when its value is read**, the table's column is `Read`, and the three rows that were phrased as questions say what the value is read for. Step 3 of the checklist loses its "Ask when its question can be answered" too — same framing, same fix.

---

docs/cli-options.md
tconbeer: Opened [#1145](https://github.com/tconbeer/harlequin/issues/1145), and the section now points at it rather than describing an alternative nobody has taken — `534a9c7`.

The issue proposes the subclass you describe: the group as a property on a `click.Parameter` subclass, so an option declares how and where it can be used where it is declared, and the partition test reads the parameters instead of the name lists. It names `CLI_ONLY_SESSION_KEYS`, `CLI_ONLY_SSH_KEYS` and `TUI_ONLY_KEYS` as the same shape, since they are the same problem, and flags `AbstractOption` as worth checking in the same pass — "every adapter option is a connection option" is a rule core enforces but nothing declares.

---

PR #1142, opened 2026-09-07
fix: don't break the JSON export on a single quote in the path

MaxFreedomPollard:
Exporting a query to JSON fails with a duckdb Parser Error whenever the destination path holds a single quote, so a folder called `Bob's exports` cannot be exported into. `json`, `jsonl` and `ndjson` are affected; `csv`, `tsv`, `parquet`, `orc` and `feather` write into that same folder fine. The two free-text JSON options have it too: a Date Format of `'%Y'` comes back as `Catalog Error: Type with name Y does not exist!`.

There is no open issue for this. I found it reading `export.py`, so the CHANGELOG entry references none.

`_export_json` (`src/harlequin/export.py:207` on main) is the one writer that reaches duckdb as SQL text rather than through a binding, and it built the whole statement with f-strings: the destination as `to '{dest_path}'`, and the options as `DATEFORMAT '{...}'` and `TIMESTAMPFORMAT '{...}'`. A single quote in any of the three ends the literal it sits in, and the rest of the statement no longer parses.

The statement now binds its values: `copy (select * from data) to ?`, with `COMPRESSION ?`, `DATEFORMAT ?` and `TIMESTAMPFORMAT ?` appended for the options that were set. `ARRAY TRUE` stays inline, since it is a fixed keyword rather than a value.

Two regression tests. `TestDestinationPaths::test_a_quote_in_the_path_is_part_of_the_name` writes into a `Bob's exports` folder for every one of the nine formats, so the assertion is that no writer is special about what a file may be called; on main the six non-JSON formats already pass it and `json`, `jsonl` and `ndjson` fail. `TestJsonOptions::test_a_quote_in_a_format_string_is_part_of_the_format` passes `'%Y'` and `'%H:%M'` and reads the quotes back out of the JSON.

Doubling the quotes (`dest_path.replace("'", "''")`) fixes the same three cases, but it puts a hand-rolled SQL escaper in a module where the other four writers hand duckdb and pyarrow their arguments directly. Binding is what those writers effectively do, and it is why csv and parquet never had this.

I bound `COMPRESSION` as well, so no value is pasted into the statement any more, and checked that it is behavior-neutral first: for each of the four choices the JSON format declares (`auto`, `gzip`, `zstd`, `uncompressed`), writing to `out.json` and to `out.json.gz` gives byte-identical files whether the codec is a bare token or a bound parameter, extension detection under `auto` included.

The `hsql` side is only reachable on a machine whose temp directory holds a quote, since `write_stream()` writes into `tempfile.TemporaryDirectory()` and copies the bytes out. The Data Exporter is where a user picks the path, so that is where this bites.

- [x] No.

- [x] Yes.

Testing performed, on macOS 26.2 arm64 with Python 3.10.15:

- `uv run pytest tests/unit_tests/test_export.py`: 117 passed. With `src/harlequin/export.py` reverted to main and the tests kept, `-k quote` is 5 failed, 7 passed: the three JSON paths and both format strings.
- `uv run pytest -m "not online"`: 1789 passed, 6 skipped, 1 failed. The failure is `tests/adapter_tests/test_sqlite.py::test_load_extension`, and it fails identically on unmodified main here: the committed `hello0.dylib` fixture is x86_64, and dlopen wants arm64 on this machine.
- `uv run mypy`: Success: no issues found in 148 source files.
- `uv run ruff format . && uv run ruff check .`: All checks passed.
- `uv run lint-imports`: Contracts: 5 kept, 0 broken.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file.
 
 - `harlequin` now exits non-zero when it fails: 1 after a crash, and 2 for a config or connection error it used to report as success ([#687](https://github.com/tconbeer/harlequin/issues/687)).
 - Harlequin now refreshes the Data Catalog when an SSH tunnel reconnects, so expanding a node no longer raises a Catalog Error ([#1127](https://github.com/tconbeer/harlequin/issues/1127)).
+- Exporting to JSON no longer fails with a Parser Error when the path holds a single quote (a `Bob's exports` folder, say), or when the Date Format or Timestamp Format option does.
 
 ## [2.13.0] - 2026-09-02
 
--- a/src/harlequin/export.py
+++ b/src/harlequin/export.py
@@ -186,28 +186,26 @@ def _export_json(
 ) -> None:
     import duckdb
 
-    array = ", ARRAY TRUE" if kwargs.get("array") else ""
-    compression = (
-        f", COMPRESSION {kwargs.get('compression')}"
-        if kwargs.get("compression")
-        else ""
-    )
-    date_format = (
-        f", DATEFORMAT '{kwargs.get('''date_format''')}'"
-        if kwargs.get("date_format")
-        else ""
-    )
-    ts_format = (
-        f", TIMESTAMPFORMAT '{kwargs.get('''timestamp_format''')}'"
-        if kwargs.get("timestamp_format")
-        else ""
-    )
+    # the one writer reached as SQL text, so every value in the statement is
+    # bound: a path or a format string holding a single quote would end the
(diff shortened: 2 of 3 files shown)


tconbeer (approved):

tconbeer:
Merged in #1146 as 556f894 — thank you, @MaxFreedomPollard.

This branch lives on your fork and could not be pushed to from the session doing the rebase, so the commit was rebased onto `main` on a branch in this repo and merged from there. Nothing in it changed: the pre- and post-rebase diffs are identical apart from a CHANGELOG hunk offset, and the merge commit credits you as co-author.

Closing in favour of #1146.

---

PR #1147, opened 2026-09-11
Warm sessions 4/6: cancellation

tconbeer:
No issue; this is PR 4 of the six in [`docs/plans/hsql-warm-sessions-design.md`](https://github.com/tconbeer/harlequin/blob/main/docs/plans/hsql-warm-sessions-design.md) §9, on top of #1140.

§4.6, delivered: **`Ctrl-C` on a session's query stops it, rather than orphaning it there.** Today the client exits 130 and the query runs on, holding the session until the database is done with it.

- **The request carries an id** — eight bytes of `os.urandom()`, a sixth section of the `REQUEST` frame — and there is a new `CANCEL` frame that names one.
- **The client's interrupt.** A `KeyboardInterrupt` out of the send-and-relay opens a **second** connection (the first is carrying the response), sends `CANCEL` with its id, copies back whatever the session says, and exits 130 — silently, as it does cold. Still stdlib-only: `os.urandom` is the whole of what this added to the client's graph, and its module budget is untouched at 38 of 40.
- **The server's answer.** `CANCEL` is answered off the server's own bookkeeping, the way `--session-status` is, so it lands while the query it stops is running rather than queueing behind it. The server marks the request and calls `connection.cancel()` from the accepting thread — the one adapter method the IDE has always called from another thread.
- **The attribution is the server's**, following `timeout.py` one level up. A cancelled cursor comes back empty and error-free, exactly like a query that matched nothing, so `_Run.stopped` reads the session's flag between results and the run stops rather than printing the empty result the cancel produced. `_Run` now answers `cancelled` beside `timed_out`, so a stopped run is exit 130 with no rows on stdout.
- **The `IMPLEMENTS_CANCEL = False` path.** Nothing is asked of the database, and the caller is told on their own stderr that the query is still running on the session, which adapter it is, and that `--session-status` says when it finishes. That is the only outcome a cancel reports; every other one is silent, because the caller already knows they pressed Ctrl-C.

Two decisions this PR made on top of the doc:

- **The request id is the client's, not the server's.** The obvious alternative is a server-assigned id, which is what the per-request numbering in the server's log already has — but it would have to reach the client *before* its query started, and the only frame ahead of a request is the handshake. `HELLO` is the frame every release has to be able to read the version out of for §4.9's skew check to work, so its shape cannot grow; the alternative was a fourth frame on the hot path, to save the client an `os.urandom(8)`. So the id travels up with the request instead. Not guessed at for secrecy — the socket is the credential and §4.7 is what protects it — only for uniqueness among the handful of requests a session holds at once.

  A cancel with *no* id was the other alternative, since a session runs one request at a time: "stop whatever is running." Rejected because it is racy in exactly the way that matters — a caller's `Ctrl-C` arriving a moment after their own request finished would kill a second caller's query instead.

- **A request is cancellable from the moment its argv arrives, not from the moment it takes its turn.** This one the doc did not name, and it is the half of the feature that is not about Ctrl-C-ing a running query: a caller who gives up while their request waits behind the one ahead is cancelling something the session is holding, and nothing else would stop it running when its turn came. Without this, `Ctrl-C` on a queued request means the query runs anyway, minutes later, with nobody waiting for it — a `create table` that lands behind the caller's back.

**That makes the load-bearing invariant one about *windows*** — and review found it broken at both ends, which is the substance of the second commit. `HarlequinConnection.cancel()` is connection-wide, so the window in which a request is registered and marked `started` has to be exactly the window in which it holds the turnstile. Widen it at either end and a cancel naming a request that has already finished interrupts whichever query holds the connection now — and marks *that* one's bookkeeping not at all, so the empty, error-free result the interrupt produces is one nobody attributes and the caller reads as an empty table, exit 0. Two things keep the windows equal, and both are needed:

- `_attend` releases a request from the bookkeeping **before** it leaves the turnstile.
- `cancel()` holds the lock a request takes to start **across** the driver call, so nothing can start between the check and the interrupt.

Tradeoffs worth naming. Holding that lock across `connection.cancel()` can delay a request start by the length of an `interrupt()` call; the other three holders of the lock are single dict operations, so this is the only one worth mentioning. A cancel that arrives before the server has parsed the request it names finds nothing and says so, and the query runs on; the window is the microseconds between the client's `sendall` returning and the server's `unpack_request`, and closing it would mean an ack frame on the hot path for a race that needs a `Ctrl-C` inside it. And the run that a cancel stops answers into a socket whose client has already gone, so the exit code and `--stats` payload it produces reach only the server's log — kept honest anyway, because that log is what an operator reads.

Where a session has no connection to interrupt — a reset between its two halves, or a session coming down — the request is still marked, so it stops rather than resuming, and the log says which of the two happened rather than reporting it as an unknown request.

**From review.** Four findings, two of them halves of the same blocker:

- **The windows above**, in both places. Each is pinned by a test that fails on the unfixed code, and the second of them is the one that would have caught this: `test_a_request_is_released_before_the_turnstile_lets_the_next_one_in` drives a real request frame through `_attend` over a `socketpair` and wraps `Turnstile.leave` to record what is registered at the moment it is called. The unit tests could not have — they drive `_hold`/`_take_turn`/`cancel` directly, so they assert the invariant while `_attend` is what maintains it.
- **The client's `send_frame` was outside the `try`**, so an interrupt landing between the request going out and the relay starting exited 130 having sent no cancel — the orphaned query this PR exists to remove. It is inside now; a frame an interrupt truncates is one the server never holds, and the cancel that follows answers `UNKNOWN`.
- **`request_id` is required rather than defaulted**, on both `Request.__init__` and `pack_request`, and `unpack_request` refuses one that is not `REQUEST_ID_BYTES` long. The bookkeeping is keyed by id, so two id-less requests would cross-wire and leave one permanently uncancellable; the wire check is what closes it, since a hand-built client could still send `b""`.
- **Two docstrings carried the plan's alternatives-weighed prose**, which AGENTS.md routes to the PR or `docs/plans/`. Trimmed, and the rest of the diff swept for the same shape. `_release()` gained one instead, because its ordering constraint is a live rule about adjacent code.

- [ ] No.
- [x] Yes; I haven't opened a PR, but the gist of the change is: the session section the "Headless & Agents" topic still needs (PR 6 of the plan) should say that `Ctrl-C` on a served invocation stops the query on the session and exits 130, as it does cold — and that an adapter without query cancellation is the exception, where the caller detaches and the query runs to completion, which hsql says on stderr.

- [x] Yes.

- **The server's bookkeeping** (`test_hsql_serve.py`) — each outcome of a cancel: one that stops a running query (and asks the connection exactly once), one for a request the session no longer holds, one for a request still queued (which then never runs), and one an adapter cannot honor (which asks the connection *nothing* and still marks the request). Plus what the answer carries over a socket: exit 130 always, and a line only when the query outlived the caller.
- **The windows** — a cancel naming a finished request never reaching its successor; `_attend`'s release ordering, over a `socketpair`; and a connection whose `cancel()` starts a thread that tries to `_take_turn` and asserts it is still blocked. Each verified to fail on the code without its fix.
- **End to end, with a real `SIGINT`** — a client interrupted mid-query against a real session exits 130 with empty stdout *and* empty stderr, the session answers the next invocation immediately, and its log says `cancel: stopped the query`. And the queued half: a client that gives up while waiting behind another request exits 130, and the temp table its statement would have created does not exist afterwards.
- **The client, against the stub** — the cancel names the id the request carried; an interrupt during the send still cancels; the session's note reaches the caller byte for byte; a session that went away between the two connections is still exit 130; and an interrupt before there is a request to name cancels nothing.
- **The frames** — the id through a round trip beside the surrogate-escaped argv, an id that could not name a request refused on the wire (empty, short, long), and that two ids are never the same.
- `StubServer` now answers every connection rather than one, since a cancelling client opens a second.

Full suite green on the merged head: 1853 passed / 7 skipped, mypy clean over 148 files, all five import-linter contracts kept, ruff clean. `docs/generated/hsql-reference.md` and `schemas/config-v1.json` are both unchanged — this PR adds no option; a cancel is a signal. `main` is merged in (through #1146), and the textual-fastdatatable 0.19.2 bump brought no snapshot churn.

https://claude.ai/code/session_01CSmAP3tQXAK4idXLU4edWJ


src/harlequin/hsql/server.py
```diff
@@ -416,6 +484,64 @@ def reset(self) -> None:
...
+                return WAITING
+        if not self._implements_cancel:
+            return UNSTOPPABLE
+        connection = self._connection
+        if connection is None:
+            # a reset between its two halves, or a session coming down: the
+            # request is marked either way, so it stops rather than resuming
+            return DETACHED
+        with contextlib.suppress(Exception):  # adapters are third-party code
+            connection.cancel()
```
tconbeer: **Blocker.** The lock is dropped at the end of the `with` above, and `connection.cancel()` is called outside it — so the "started" the cancel checked is not the "started" it acts on. `HarlequinConnection.cancel()` is connection-wide by contract (`adapter.py:96-99`; both bundled adapters call `conn.interrupt()`), so a cancel that arrives just after its own request finished lands on whatever is holding the connection now.

Deterministic, using only the server's own methods in the order `_attend` calls them:

```python
a = server._hold(b"AAAAAAAA"); server._take_turn(a)   # A runs, finishes
b = server._hold(b"BBBBBBBB"); server._take_turn(b)   # leave() precedes _release(), so B starts
server.cancel(b"AAAAAAAA")
# -> "stopped the query", connection.cancel() called once, b.cancelled is False
```

B's query is interrupted in the driver and B's `InFlight` is never marked, so `Served.cancelled` is False and nothing attributes it. A cancelled query comes back empty and error-free — the premise this whole design rests on — so **B prints an empty result set and exits 0**.

`_take_turn` already takes this lock, so widening the `with` to cover the rest of the method body closes the half that lives here: nothing can start while a cancel is between its check and the driver call. The other half is the `leave()`/`_release()` ordering in `_attend`; both are needed.

---

src/harlequin/hsql/server.py
```diff
@@ -545,25 +671,36 @@ def _attend(self, connection: socket.socket) -> None:
...
+                    if self._take_turn(in_flight):
+                        segments, code = self._run(request, in_flight)
+                    else:
+                        # cancelled while it waited for the request ahead of it
+                        segments, code = [], int(ExitCode.INTERRUPT)
+                finally:
+                    self._turnstile.leave()
             finally:
-                self._turnstile.leave()
+                self._release(in_flight)
```
tconbeer: The other half of the blocker: `leave()` runs before `_release()`, so between them a finished request is still in `_in_flight` with `started=True` while the next request has already taken its turn and entered the driver. A cancel naming the finished one then reports `STOPPED` and interrupts its successor's query.

Releasing inside the turnstile — `_release(in_flight)` before `leave()` — means a request is out of the bookkeeping before anything else can start, and that same cancel answers `UNKNOWN`, which says nothing to the caller and still exits 130.

No test can catch this today: the unit tests drive `_hold`/`_take_turn`/`cancel` directly and never construct a started-but-unreleased entry alongside a running successor.

---

src/harlequin/hsql/client.py
```diff
@@ -236,9 +244,40 @@ def _exchange(
             stdin=stdin,
             stdout_isatty=_isatty(sys.stdout),
             stderr_isatty=_isatty(sys.stderr),
+            request_id=request_id,
         ),
     )
-    return _relay(connection)
+    try:
```
tconbeer: `send_frame` is outside the `try`, so a `KeyboardInterrupt` delivered after `sendall` returns but before the `try` is entered goes to `run()`'s handler and exits 130 **without sending a `CANCEL`** — the orphaned query this PR exists to remove.

Widening the `try` to start above `protocol.send_frame(...)` is strictly safe: an interrupt landing mid-`sendall` leaves a truncated frame, the server never holds the request, and the cancel that follows is answered `UNKNOWN` — silent, still 130. The remaining window between `except` and `_cancel` can't be closed, but this one can.

---

src/harlequin/hsql/protocol.py
```diff
@@ -117,13 +140,26 @@ def __init__(
         stdin: "bytes | None",
         stdout_isatty: bool = False,
         stderr_isatty: bool = False,
+        request_id: bytes = b"",
```
tconbeer: Nothing in `src/` builds a request without an id — `client.py:236` is the only caller and `unpack_request` always supplies one — so this default (and the matching one on `pack_request`) buys nothing and costs a real hazard. `_hold` keys `_in_flight` by id and overwrites blindly (`server.py:526`), `_release` pops by id (`server.py:531`): two id-less requests cross-wire, the second evicting the first's entry and the first's release popping the second's, leaving it permanently uncancellable.

Making the parameter required costs four test edits and makes an uncancellable request unconstructable. A length check in `unpack_request`, beside the ones already there, would be belt and braces on the wire.

---

src/harlequin/hsql/protocol.py
```diff
@@ -117,13 +140,26 @@ def __init__(
...
+        self.request_id = request_id
+
+
+def new_request_id() -> bytes:
+    """An id for one request, unique among the ones a session holds at once.
+
+    The client's rather than the server's, because a server-assigned one would
+    have to reach the client before its query started, and the only frame that
+    precedes a request is the handshake -- which every release has to be able
+    to read the version out of, so its shape cannot grow.
```
tconbeer: This is the alternative-weighed paragraph AGENTS.md routes to the PR or `docs/plans/` — and it is near-verbatim the paragraph this PR adds at `docs/plans/hsql-warm-sessions-design.md:926-931`, so it already has its home.

```python
"""An id for one request, unique among the ones a session holds at once.

Assigned by the client, because a `CANCEL` names it and the client is what
sends one.
"""
```

Same shape at `server.py:518-522` (`_hold`'s "Before the turnstile rather than after it: ..."), which is the plan's other new paragraph.

---

tconbeer (commented):
## Scope of this review

Four narrow passes, each excluding the others' lenses:

1. **Server bookkeeping and thread-safety** — `InFlight`, `Server.cancel()`, `_hold`/`_take_turn`/`_release`, `_attend`, `_send_cancel`, and the `_Run.cancelled`/`stopped` plumbing.
2. **Client and wire protocol** — `_cancel()`'s second connection, interrupt timing, the sixth request section, version skew, cold-start budget.
3. **Tests** — whether the assertions pin the claims the description makes.
4. **Prose** — docstrings, comments, CHANGELOG, AGENTS.md, against this repo's own rules.

On this head `ruff check` and `ruff format --check` are clean, and `pytest tests/unit_tests/test_hsql_serve.py tests/unit_tests/test_hsql_session.py` is 224 passed.

## One blocker

**A cancel can interrupt the next request's query, which then reports success with no rows.** `Server.cancel()` drops `_in_flight_lock` before calling `connection.cancel()`, and `_attend` releases the turnstile before the in-flight entry. Detail, a deterministic reproduction, and the two one-line fixes are in the inline comments on `server.py:506-514` and `server.py:700-703` — both are needed.

It matters because `docs/plans/hsql-warm-sessions-design.md:936-941` asserts this invariant in as many words ("never interrupts the database while it is idle — which would land on whichever request holds it instead"), and the failure mode is the worst one available here: a second caller gets an empty result set and exit 0. The window needs a cancel arriving within milliseconds of its own request completing with another request queued — rare interactively, less rare where a wrapper SIGINTs on a timer.

## Worth fixing

- **The request is sent outside the `try` that turns an interrupt into a cancel** (`client.py:237-253`) — inline comment there.
- **`request_id` defaults to `b""`** (`protocol.py:143`, `protocol.py:202`) — inline comment there.
- **Two comments in `cli.py` now describe only half of what `stopped` means.** `_Run.stopped`'s docstring was updated to "the clock ran out, or the caller gave up" (`cli.py:2199`), but the two comments that read the flag still name only the clock: `cli.py:2356` ("the clock ran out between statements") and `cli.py:2363` ("...is not this run's error to report; the deadline is").

## Smaller

- `_cancel()` skips the version check `_exchange` does (`client.py:270-272` vs `client.py:205-216`) and then relays whatever the second connection sends straight to the caller's stdout and stderr. Narrow — the session must have been restarted under a different release between the request and the `Ctrl-C` — but the check is two lines and the relay is unfiltered. The same reconnect also skips `check_runtime_dir`, which the first connection pays for (`client.py:97`).
- `_cancel()` sets no socket timeout, and the server answers it by calling third-party `cancel()` on the accept thread (`server.py:747`). An adapter whose `cancel()` blocks parks a non-interactive caller indefinitely after they have already given up. Interactively a second `Ctrl-C` escapes, which the comment anticipates; a `settimeout()` would bound the other case.
- `Server.cancel()` does not consult `_abandoned` (`server.py:506-514`, vs `Server.connection()` at `server.py:457-462`). After a grace-expired `--timeout` the flag is set but `self._connection` still points at the live object, so a cancel calls into a connection a thread is still inside, and logs `STOPPED` where `DETACHED` is the honest answer.
- `_take_turn` marks every served mode started, not just the query path, so a `Ctrl-C` on a served `--info`, `--catalog` or `--config show` reaches `connection.cancel()` on a connection no query is using. Harmless for the bundled adapters — I checked that an idle `interrupt()` does not poison the next query on either duckdb or sqlite — but the contract does not promise it. The other half is that `--catalog` and `--catalog-search` never build a `_Run`, so "marked, therefore it stops" does not hold for them either.
- `-o PATH` opens the destination directly (`cli.py:2626-2632`), so a cancelled run leaves a truncated file behind. Same shape as the existing timeout path, so in-family rather than new — worth naming only because "empty stdout" is what the tests pin.

## On the tests

The four outcome tests (`test_hsql_serve.py:765-800`) pin what they claim, `connection.cancels` included, and the two end-to-end tests are the strongest thing here: a real `SIGINT`, assertions on *both* streams being empty, and `blocked`/`_until_status` rather than sleeps, so they should hold up on a slow runner. `test_a_client_that_gives_up_while_it_waits_runs_nothing` asserting the temp table's absence afterwards is exactly the right shape.

The gap is the one the blocker falls through. Every unit test drives `_hold`/`_take_turn`/`cancel` directly, and the case none of them constructs is a *started* request still in `_in_flight` while the next one runs — which is what `_attend`'s `leave()`-before-`_release()` ordering produces. `test_a_cancel_for_a_request_the_session_no_longer_holds_stops_nothing` comes closest, but it releases without ever taking a turn. `DETACHED` and `FAILED` are also unexercised, and nothing covers `--stats` on a cancelled run (`status="error"`, `message="cancelled"`).

## Prose

AGENTS.md says a docstring explains what the code does and why, and that the design behind a change — the alternatives weighed — belongs in the PR or `docs/plans/`. This PR put it in both: several new docstrings are near-verbatim the `docs/plans` paragraph it adds in the same commit. Inline on `protocol.py:155-160`; `server.py:518-522` (`_hold`) is the same shape.

Also worth trimming:

- `server.py:84-94` — five lines of design note on a constant group; the neighbour at `protocol.py:95-96` gets one line for two constants.
- "A cancelled query comes back empty and error-free" is now stated in four new places (`server.py:17-21`, `server.py:103-106`, `server.py:261-264`) on top of where the run acts on it (`cli.py:2205`). Keeping it at `_Run.stopped` loses nothing.
- `server.py:741-743` (`_send_cancel`'s second paragraph repeats the constants' note), `cli.py:2194-2197` (`_Run.cancelled`'s contrast with the cold path — every other `_Run` member is a one-liner), `client.py:258-265` (`_cancel`'s second sentence is `protocol.py:60-62`'s, which is the right home for it), `diagnostics.py:166-170` (first sentence is the message body one line below), `protocol.py:98-100`.

The `served` field docstring at `cli.py:2181-2183`, `report_cancel()`, and the new module-docstring paragraph at `client.py:16-18` are right as they are. The CHANGELOG entry is compliant — one sentence, user-facing, no internals, and correctly carries no link. The AGENTS.md edit is accurate. The `docs/plans` edit belongs and is the right home for the rationale duplicated above — but its invariant paragraph is the one the blocker falsifies, so it needs the code fixed or the claim softened.

## Not checked

- I did not review `_answer`, `Recorder`, `_Stream`, the SSH tunnel, the lock file, the accept loop or the shutdown-signal path beyond where they touch cancel.
- I did not exercise the `IMPLEMENTS_CANCEL = False` path against a real adapter that declares it; only the in-process server tests cover it.
- I ran `ruff` and the two hsql test files, not the full suite, and not mypy or `lint-imports`.
- I did not re-derive the identity/refusal rules from #1140, or check a cancel against `--session-reset` beyond reading `_close_connection`.
- I did not measure the interrupt-timing windows; they are reasoned from CPython's signal-delivery semantics, not observed. The `server.py` race is the one I reproduced.
- Snapshot tests, Windows behaviour and the IDE are untouched by this diff and unexamined.

A clean review is necessary, not sufficient — the merge call is yours.

---

src/harlequin/hsql/server.py
tconbeer: Right, and the repro is exact. Fixed in 28ec9fd: the whole method body is now under the `with`, so `_take_turn` — which takes the same lock — cannot let a request start between the `started` check and the driver call.

With the release ordering fixed too, the two together give an invariant rather than two patches: **a request that is registered and `started` is one that holds the turnstile.** Inside the lock, A registered ⟹ A has not released ⟹ A has not left the turnstile ⟹ nobody else is in the driver, and holding the lock across `connection.cancel()` keeps that true until the interrupt is issued. Dropping either one reopens it — dropping this half leaves the narrow interleaving where A finishes, releases, leaves, and B starts, all between the check and the call.

Pinned by `test_nothing_takes_a_turn_between_a_cancel_and_the_interrupt_it_sends`: a connection whose `cancel()` starts a thread that tries to `_take_turn` and asserts it is still blocked. Fails `[False] == [True]` with the narrow lock restored.

The cost is that a request start can be delayed by the length of an `interrupt()` call. Worth it, and the other lock holders are three dict operations.

---

src/harlequin/hsql/server.py
tconbeer: Agreed, and fixed in 28ec9fd — `_release(in_flight)` now runs before `leave()` in that `finally`. The queue-timeout path releases on its own line instead, so there is one `_release` per path rather than an outer `finally` that would double up.

You're right that no test caught it, and that the unit tests could not: they drive `_hold`/`_take_turn`/`cancel` directly, so they assert the invariant while `_attend` is what maintains it. Two tests now:

- `test_a_cancel_never_reaches_the_request_that_took_over_the_connection` — your sequence, with the `_release(first)` that `_attend` now performs in it, asserting `UNKNOWN` / `connection.cancels == 0` / `not second.cancelled`.
- `test_a_request_is_released_before_the_turnstile_lets_the_next_one_in` — pins `_attend` itself rather than the invariant. It drives a real request frame through `_attend` over a `socketpair`, wrapping `Turnstile.leave` to record what is registered at the moment it is called, and asserts that is empty. Fails `[[b'AAAAAAAA']] == [[]]` with the ordering restored.

The second one is the one that would have caught this, so it is worth the socketpair.

---

src/harlequin/hsql/client.py
tconbeer: Fixed in 28ec9fd — the `try` now opens above `send_frame`, and your reasoning about the truncated frame is what makes it safe: the server's `recv_frame` raises, `report_bad_request` drops the connection, the request is never held, and the cancel that follows answers `UNKNOWN` — silent, still 130.

`new_request_id()` stays outside, since an interrupt ahead of it has nothing to name and nothing was sent.

`test_an_interrupt_while_the_request_is_going_out_still_cancels` covers it: `send_frame` raises `KeyboardInterrupt` after the real send of a `REQUEST`, and the stub is asserted to have received a `CANCEL` naming that request's id.

Agreed the window between `except` and `_cancel` can't be closed — an interrupt there exits 130 with the query running, which is the same outcome as the `IMPLEMENTS_CANCEL = False` path minus the note. Narrower than the one this closed and with no cheap fix, so I've left it.

---

src/harlequin/hsql/protocol.py
tconbeer: Done in 28ec9fd, all three parts. `request_id` is required on both `Request.__init__` and `pack_request` (moved above the two `isatty` defaults so it reads as the section it is), and `unpack_request` now checks the length beside its other section checks:

```python
if len(raw_id) != 1 or len(raw_id[0]) != REQUEST_ID_BYTES:
    raise ProtocolError(f"a request carries one {REQUEST_ID_BYTES}-byte id")
```

The wire check is the one that actually closes the hazard you describe, since `_hold`/`_release` only ever see what `unpack_request` produced — a hand-built client could still have sent `b""` with only the signature changed. Required parameters are what keep the hazard from being reachable from inside `src/`.

Five test call sites rather than four: `test_a_request_carries_which_streams_are_terminals` builds one too. The ids the server tests name are now `A_REQUEST` / `ANOTHER_REQUEST` constants, since they have to be 8 bytes and `b"seven"` no longer is. New test `test_a_request_whose_id_could_not_name_it_is_refused`, parametrized over empty, short and long.

---

src/harlequin/hsql/protocol.py
tconbeer: Fair, and the near-verbatim duplication is the tell. Took your wording for `new_request_id()` in 28ec9fd, and trimmed `_hold()` the same way:

```python
"""Register the request, so a cancel naming it finds it.

From the moment its argv arrives rather than from its turn, so that a
request still waiting for the one ahead can be cancelled too.
"""
```

Swept the rest of the diff for the same shape while I was there — `InFlight`, `Served.cancelled`, `_send_cancel` and the client's `_cancel` each lost a paragraph that was arguing rather than describing.

The one docstring that grew is `_release()`, which had none: it now says the ordering constraint it is subject to, because that is a live rule about the code next to it rather than a road not taken, and `_attend`'s `finally` points at it instead of restating it.

---

tconbeer:
**CI is green on the current head** (`519fe66`): all 15 test jobs, all 15 build jobs, Static Analysis and Validate Plugin. Recording the red that came before it, because it is not this PR's and it is worth knowing about.

`MacOs-latest - 3.10` and `- 3.11` failed on the superseded head `28ec9fd`, both on the same test:

```
  - ValueError: No Tab with id 'tab-1'
```

**Not this PR's: it is red on the base branch too.** The same test with the same error failed on `MacOs-latest - 3.11` on `main` at [`58e3e40`](https://github.com/tconbeer/harlequin/actions/runs/34554337216) — this PR's base commit, before any of this code existed. This PR touches `hsql/` and nothing the TUI mounts.

**No fix exists to port**, so nothing was. `main`'s tip is green, but by luck rather than by a fix: nothing in #1141 or #1146 goes near `Tabs`. The one re-run the failure was owed happened naturally — merging `main` in produced a fresh full run on `519fe66`, and all five macOS jobs passed.

**What it is**, from the traceback, for whoever picks it up. `EditorCollection.on_mount` calls `action_new_buffer()` → `tabs.add_tab(...)`. Textual's `Tabs.add_tab` mounts the tab into `#tabs-list`, then `refresh_active()` awaits that mount and sets `self.active = tab_widget.id`; `validate_active` queries `#tabs-list > #tab-1`, gets nothing back, and raises. So the mount was awaited and the tab still is not queryable — `Tabs` is itself mid-mount, because the call comes from an ancestor's `on_mount`. Rare, macOS-only so far, and it crashes the app rather than the assertion, which is why it surfaces as `run_test()` re-raising at `__aexit__`.

Per AGENTS.md's "fix things upstream" rule the real fix is in Textual, since that ordering is `Tabs`' own; the Harlequin-side mitigation would be to stop adding tabs from inside `on_mount`, which is a startup-ordering change well outside a cancellation PR and touches every buffer test. Happy to open an upstream issue with the traceback, or to take the Harlequin-side change as its own PR — say which and I'll do it. Leaving this PR as it is either way.

---

tconbeer:
I worked the five inline threads first and missed that the review body carried findings of its own. Those are in `17539cf`, so the review is now addressed end to end.

**A cancel reaches only a request that asked for the connection.** You were right that `_take_turn` marking every served mode started is not what `cancel()`'s reach should be keyed on. `Served.connection()` is the one place a served invocation takes the connection, so that is where the request is now noted as holding it; a cancel that finds it unheld answers a new `IDLE` and marks the request without interrupting anything. That covers `--info`, `--config show` and `--session-reset`, and keeps `--catalog`/`--catalog-search` cancellable, since they do take the connection.

The half you noted about `--catalog` never building a `_Run` still stands — for those two, "marked, therefore it stops" means the marking is inert and `connection.cancel()` is what does the work. I left that rather than threading `run.stopped` through the catalog walk, which is its own change.

**`cancel()` consults `_abandoned`.** Real bug, and the reasoning was exactly right: after a grace-expired `--timeout` the flag is set and `_connection` still points at the live object, so a cancel called into a connection a worker thread was still inside. `DETACHED` now, before the driver call.

**The second connection makes every check the first one made.** `check_runtime_dir` before anything is sent, and the version check before the cancel goes out — the relay being unfiltered is what decided it: a foreign release writing to the caller's terminal is worse than a cancel that does not land. It also carries a `CANCEL_TIMEOUT`, so a blocking `cancel()` cannot park a caller who has already given up and may not be at a terminal to press Ctrl-C again.

**The two `cli.py` comments** that named only the clock now name both.

**Tests** for `IDLE`, `DETACHED`, and a bug in the cancel path, plus `--stats` on a cancelled run (`status="error"`, `message="cancelled"`, exit 130). The `IDLE` gate is pinned by `test_an_interrupted_query_stops_and_leaves_the_session_up` rather than only by a unit test — I checked that removing the `Served.connection()` hook hangs it, which is the right failure: the query is never interrupted and the session stays busy.

**Prose**: the constants block is one line; `_send_cancel`, `_Run.cancelled`, `_cancel`, `report_cancel_unsupported`, `REQUEST_ID_BYTES` and the `CANCEL` frame each lost the paragraph that was arguing; "a cancelled query comes back empty and error-free" is now stated where the run acts on it and not four times over. `_hold` and `new_request_id` were done in `28ec9fd`.

**Not taken**, both deliberately: `-o PATH` still leaves a truncated file on a cancelled run — in-family with the timeout path, so it belongs with that or not at all. And the `docs/plans` invariant paragraph you flagged is now rewritten around the windows rule rather than softened, since the code holds it.

Full suite 1857 passed / 7 skipped, mypy clean over 148 files, ruff clean, all five import contracts kept.

---

tconbeer:
CI caught a hole the `IDLE` gate from `17539cf` opened — my failure, and a useful one. Fixed in `da1de83`.

`ubuntu-latest - 3.13` and `MacOs-latest - 3.12` failed on `test_an_interrupted_query_stops_and_leaves_the_session_up`, with the session's log reading a cancel that was not `stopped the query` and `request 1: exit 130 in 51ms` — 51ms for a query that counts to 2×10¹¹.

**What it exposed.** `--session-status` reports `busy` from the moment a request takes its turn, which is a moment *before* it has parsed its argv, read config and asked for the connection. So the test's `Ctrl-C` was landing in that window: `holds_connection` was still False, the cancel answered `IDLE`, and nothing was interrupted. The run still stopped and the session stayed usable, because the flag was set — but only because `_execute_all` reads `run.stopped` *after* submitting the first statement. On duckdb that submit is cheap; on an adapter whose `execute()` blocks, a caller's interrupt would not have stopped the database at all.

**The fix.** `_execute_all` now reads `run.stopped` before the first statement as well as between them, so a run that was over before it began submits nothing. That is what makes the `IDLE` answer safe rather than merely quiet: either the cancel interrupted a query, or no query was ever sent. Pinned by `test_a_run_cancelled_before_it_started_submits_nothing`, which fails with `Left contains one more item: 'select 1'` without the guard.

**The test.** Asserting the outcome *by name* was pinning a timing detail — which outcome a cancel reports depends on whether the `Ctrl-C` beat the request to the connection, and both are correct. It now takes one more status round trip before signalling, which puts the interrupt inside the query rather than ahead of it (`cancel: stopped the query` locally, 3/3), and asserts that a cancel was logged rather than which one. The assertion that actually proves the query stopped is the next invocation being answered at once — that is what times out when nothing stops it, and it is what caught the unwired gate when I checked this change was load-bearing.

Full suite 1858 passed / 7 skipped, mypy clean, ruff clean, all five import contracts kept.

---

PR #1130, opened 2026-09-03
M4 1/5: edit the buffer in $EDITOR

tconbeer:
Closes #1102

PR 1 of M4 Release A, per [§3.6 of the M4 technical plan](https://github.com/tconbeer/harlequin/blob/main/docs/plans/m4-hsql-technical-plan.md).

**What** are the key elements of this solution?

`alt+e` in the Query Editor hands the active buffer to the editor named by `$VISUAL` or `$EDITOR`:

```
capture the buffer -> temp file -> suspend the app -> run the editor ->
read the file back -> one undoable edit -> delete the temp file
```

- **`harlequin/external.py`** (new) owns the process; the app owns the widgets. It imports Textual only under `TYPE_CHECKING`, apart from one function-local import of `SuspendNotSupported` where it is caught.
  - `resolve_editor()` — `$VISUAL`, then `$EDITOR`, then refuse with an error naming both. No fallback to `vi` or `notepad`.
  - `run_in_terminal(app, argv)` — `with app.suspend(): subprocess.call(...)`, turning `SuspendNotSupported` and `OSError` into a `HarlequinExternalError`.
  - `launch_external_editor(app, text)` — a `.sql` temp file handed over as the editor's last argument (the `git rebase -i` shape), closed before the child opens it, always unlinked, read back with universal newlines. Returns `ExternalEdit(returncode, text)`, where `text is None` means the edit was discarded.
- **`CodeEditor.action_launch_external_editor`** — an error modal for a `HarlequinExternalError`, a warning notification naming the status when the editor exits non-zero, otherwise one assignment to `text`, so the whole round trip is one `ctrl+z` away.
- **`code_editor.launch_external_editor`** joins `HARLEQUIN_ACTIONS`, bound to `alt+e` in the vscode keymap.

**Why** did you design your solution this way? Did you assess any alternatives? Are there tradeoffs?

- **A non-zero exit discards the edit.** That's the `git commit` / `:cq` convention, and it is the only escape hatch a user has once the editor is open.
- **The exchange is a file, not stdin**, because editors do not read stdin.
- **`run_in_terminal()` is a seam.** `App.suspend()` is main-thread only and raises `SuspendNotSupported` on the headless driver, so the real thing cannot be reached from `run_test()`; the seam is what the round-trip tests replace, and the unsupported-terminal path is tested for real against the headless driver.
[...]

PR #1148, opened 2026-09-12
fix: derive binding priority per key so alt+<letter> can be bound

anandghegde:
What are the key elements of this solution?

bindings.bind() now decides priority per key of passing one flag through for the whole comma-separated list. A new key_needs_priority() marks the two spellings that get delivered to a widget as text. Alt+<character> and alt+shift+<character>. As priority bindings and leaves everything else bubbling. Because BindingsMap.bind() gives every key in a call the priority the keys get partitioned and its called once per group.

That replaces the per-action priority=True on code_editor.launch_external_editor added in #1130. The fix now applies to any action a user binds to an alt key rather than to alt+e on one action.

The root cause: a terminal sends alt+e as ESC e and Textuals parser turns that into a key named alt+e that still carries the character "e" (_xterm_parser._sequence_to_key_events: if len(name) == 1 and alt: name = f"alt+{name}"). Key.is_printable only looks at character. A focused TextArea or Input handles the event as text and inserts the letter. And a bubbling binding on the key never runs.

Why did you design your solution this way? Did you assess any alternatives? Are there tradeoffs?

The issue suggests handling this in bind() and that's where it belongs: the thing that makes the key unreachable is a property of the key, not of the action or of the widget that happens to be focused. Deriving it in bind() means a keymap author binding alt+j to something gets working behavior without anyone touching HARLEQUIN_ACTIONS.

The issue also asks what else carries a character. Going through Textual 8.2.8 the full set of keys delivered with a printable character is:

1. Bare printable keys. E, space, full_stop and the punctuation names

2. Alt+<printable>

3. Alt+shift+<letter> (uppercase arrives as alt+shift+ plus the lowercase letter)

Only (2) and (3) are treated as priority. Ctrl+<letter> isn't affected. It arrives as a control code, which's n't printable. And named keys like alt+left never take the alt+ prefix at all since the parser only adds it when the key name is a single character. Under Kittys protocol these keys arrive with no character so the priority flag is a no-op there.

Does this PR require a change to Harlequins docs?

- [x] No.

- [ ] Yes. I have opened a PR at [tconbeer/harlequin-web](https://github.com/tconbeer/harlequin-web).

- [ ] Yes; I haven't opened a PR but the gist of the change is:...

I checked src/docs/keymaps/config.md, src/docs/bindings.md and src/docs/troubleshooting/key-bindings.md in harlequin-web. None of them document a limitation on keys so there's nothing that needs correcting. Happy to add a note if you'd like the key rules written down somewhere.

Did you. Update tests for this change?

- [x] Yes.

- [ ] No I believe tests aren't necessary.

- [ ] No I need help with testing this change.

- tests/unit_tests/test_bindings.py was a # TODO! placeholder so it now covers which keys need priority the per-key partitioning (binding "alt+e,ctrl+e" in one call gives one priority binding and one bubbling one) and that show/key_display survive the split.

- test_keymap_from_config.py::test_alt_letter_binding_beats_the_focused_editor binds alt+n to code_editor.new_buffer from a config keymap and checks a buffer opens of an "n" being typed. It constructs the event directly because pilot.press() doesn't send a character alongside an alt key and so can't reproduce the bug. This also covers the ancestor-target case that the existing alt+e coverage doesn't. New_buffer lives on EditorCollection, not CodeEditor.

- test_keymap_vscode.py::test_editor_bindings_do_not_beat_the_find_input is the guard for the alternative above.

Full suite green locally on 3.10: 1853 passed 5 skipped 148 snapshots passed, with ruff and mypy clean. The one test I skipped is test_load_extension. The committed hello0.dylib fixture is x86_64. I'm on an arm64 Mac. It fails the way on a clean checkout of main so its unrelated to this change.

- [x] I have added an entry, to CHANGELOG.md under the [Unreleased] section heading. That entry references the issue closed by this PR.

- [x ] I acknowledge Harlequins MIT license. I do not own my contribution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file.
 - `harlequin` now exits non-zero when it fails: 1 after a crash, and 2 for a config or connection error it used to report as success ([#687](https://github.com/tconbeer/harlequin/issues/687)).
 - Harlequin now refreshes the Data Catalog when an SSH tunnel reconnects, so expanding a node no longer raises a Catalog Error ([#1127](https://github.com/tconbeer/harlequin/issues/1127)).
 - Exporting to JSON no longer fails with a Parser Error when the path holds a single quote (a `Bob's exports` folder, say), or when the Date Format or Timestamp Format option does.
+- A keymap can now bind `alt+<letter>` to a Query Editor action; the letter used to be typed into the buffer instead of running the action ([#1131](https://github.com/tconbeer/harlequin/issues/1131)).
 
 ## [2.13.0] - 2026-09-02
 
--- a/src/harlequin/actions.py
+++ b/src/harlequin/actions.py
@@ -107,13 +107,10 @@ HARLEQUIN_ACTIONS = {
     "code_editor.load_buffer": Action(
         target=CodeEditor, action="load", description="Open Query", show=True
     ),
-    # priority: a focused TextArea inserts any key that carries a printable
-    # character, and a terminal's alt+e carries an "e".
     "code_editor.launch_external_editor": Action(
         target=CodeEditor,
         action="launch_external_editor",
         description="Launch External Editor",
-        priority=True,
     ),
     "code_editor.find": Action(
         target=CodeEditor, action="find", description="Find", show=True
--- a/src/harlequin/bindings.py
+++ b/src/harlequin/bindings.py
@@ -9,6 +9,27 @@ if TYPE_CHECKING:
     from textual.widget import Widget
 
 
+def key_needs_priority(key: str) -> bool:
+    """Must a binding on this key be checked before the focused widget sees it?
+
+    A terminal sends alt+e as ESC e, which Textual parses into the key `alt+e`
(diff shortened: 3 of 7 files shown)


tconbeer:
Superseded by #1152, which is this same commit by @anandghegde rebased onto `main` (the only conflict was two neighboring `[Unreleased]` bullets in `CHANGELOG.md`; both are kept). The rebase landed on a branch in this repo rather than here because this PR's head is on a fork I can't push to.

The two macOS CI failures here were `test_ssh_recovery.py::test_a_refresh_that_reconnects_does_not_ask_for_a_second_one` raising `ValueError: No Tab with id 'tab-1'` — the pre-existing crash fixed on `main` by #1151, which the rebase picks up. Everything else is green: full suite on 3.10, `ruff`, `mypy` and `lint-imports`.

Thanks for the fix, @anandghegde — authorship is preserved on the commit.

---

tconbeer:
Merged as 51f6930 via #1152 — CI was green across all 32 checks. Closing this one out; thanks again @anandghegde.

---

Issue #898, opened 2025-12-11
Feature request: environement variable extrapolation in configuration


noirbizarre:
Hi 👋🏼 

Would it be possible to add environment variables extrapolation support in the config file?
Due to security, I avoid having credentials stored in a clear text file and often rely on a password manager (gopass, 1password, bitwarden...).
To have some security as well as some flexibility, I would love to be able to write a profile with this kind of syntax:

```toml
[profiles.my-db]
adapter = "postgres"
host = "{env.MY_DB_HOST}"
user = "{env.MY_DB_USER}"
password = "{env.MY_DB_PASSSWORD}"
dbname = "{env.MY_DB_NAME}"
```

Or using pyproject.toml

```toml
[tool.harlequin.profiles.my-db]

adapter = "postgres"
host = "{env.MY_DB_HOST}"
user = "{env.MY_DB_USER}"
password = "{env.MY_DB_PASSSWORD}"
dbname = "{env.MY_DB_NAME}"
```

It would help with password managers but also when having DB access stored in a `.env` file, which can be automatically loaded.

I don't have any preference for the syntax; it could also be `{env:VAR}` or whatever.

Does it seem acceptable to you? Would you accept a pull request for it?

tconbeer:
If you use the driver's expected var names (e.g. PGHOST, PGUSER, PGPASSWORD), and don't specify those attributes in a profile or at the cmd line, libpg will load them for you. Most adapters have some version of this. Harlequin will also work with a PGPASS file, etc. 

So I'm not opposed to this feature, but I think the problem is already solved for most users. I see there is a gap with custom var names, but there are workarounds for that also (like invoking harlequin with PGPASS=$MY_PASS harlequin). 

Implementation would have to be thoughtful so we don't require escaping characters like brackets; it would have to work with the config app also. 

Thanks for your interest here, and for your broader work in OSS!

tconbeer:
This feature is now planned, with the syntax `${MY_VAR}`, mirroring the typical docker-compose syntax. It should make it into the next 1-2 releases.

PR #1150, opened 2026-09-14
Warm sessions 5/6: lifecycle and state hygiene

tconbeer:
No issue; this is PR 5 of the six in [`docs/plans/hsql-warm-sessions-design.md`](https://github.com/tconbeer/harlequin/blob/main/docs/plans/hsql-warm-sessions-design.md) §9, on top of #1147.

§4.7's two clocks and §5's two state-hygiene pieces — one of which review cut back, see below.

- **`--idle-timeout SECONDS` (default 30m) and `--max-lifetime SECONDS` (default 8h)** on `hsql --serve`, `0` to switch either off. Both join `SERVER_OPTIONS`, so `--serve` takes them and every other invocation refuses them — the partition test picked both up without a change.
- **They come down the way a stop signal does.** The accept loop reads them between accepts; an expiry closes the listener, unlinks the socket, waits at the turnstile for whatever is in flight, and closes the connection. So a lifetime that runs out mid-query does not take the running request's answer with it, and the operator's terminal says which clock ran out and how to switch it off. (It does leave that query unreachable while it finishes — §8 now records why, per review.)
- **`--session-status` answers `idle_timeout_s` and `expires_in_s`** off them, which have been `null` since #1140 put them in the document. `expires_in_s` is the sooner of the two countdowns, floored at zero.
- **A session left in a transaction mode it did not connect in says so**, on the caller's stderr, after every request. **Narrower than the plan advertised** — see the first review finding below.
- **`--serve` warns when a secret is typed on its command line**, naming the flags and pointing at `-P`. Detected through the adapter's declarative `secret=` and `harlequin.redact`'s reading of a DSN, so `CONN_STR` is covered too. A plain filepath is not a secret: the check asks whether `redact_conn_str()` finds a credential *span*, which needs `scheme://user:password@host` or a `password=`-shaped pair, so `./warehouse.db`, `:memory:` and `md:my_db` are all silent.

Three decisions this PR made on top of the doc, all three recorded in §9:

- **A status poll is not use of a session.** The idle clock counts requests, not connections: an agent that polls a session's status on a timer would otherwise keep a warehouse credential open forever by watching a session it had stopped sending queries to.

  Review found the first cut of this got it wrong in the other direction, and the fix decided the mechanism. A client is **accepted before it says what it wants**, and `-f -` reads its script *after* the handshake — so a connected client was invisible to `_expired()`: in no turnstile, and the clock wound only once its REQUEST frame parsed. A slow producer upstream of `-f -` got the socket unlinked under it. The obvious repair, winding the clock on accept, would have broken the property above, because a status poll is an accepted connection too and would reset the clock every time. So the session **counts clients it has accepted and holds the clock while any is open** — pausing rather than resetting, which separates the two cases by duration: a poll holds it for microseconds, a `-f -` for the whole stdin read.

- **The idle clock does not run while a request does; the lifetime clock does.** The idle one is about a session nobody is using, so an hour-long query is not idle. The lifetime one bounds the credential, so it is not exempt for being busy; it stops the accept loop and lets the request in flight finish, which is the path a `SIGINT` already took. `expires_in_s` follows the same split: a busy session is answered by its lifetime alone, because how long it will be idle *after* the current query is not a question the server can answer.

- **The transaction-mode note reports a setting, not open work.** The plan (§5 item 2) proposed this as the mitigation for the sharp edge — a `BEGIN` left open — on the grounds that `HarlequinConnection.transaction_mode` already exists. **That premise does not hold**, as review established: the mode is the adapter's Auto/Manual *setting*, a plain attribute only `toggle_transaction_mode()` moves, and hsql never calls it (callers live in `app.py` alone; DuckDB's adapter has no modes at all). So a session reached through hsql cannot leave the mode it connected in, and a `begin` left open is silent — the exact case it was sold on.

  Rather than reach into the adapter contract, the claim is cut back to what this observes: a session moved **out of band** is reported, and nothing is said about what it may be holding. §5 item 2 now records that its premise fails and that `--session-reset` is the only thing in this design that answers an open transaction; detecting one needs a new contract member, which is §7's "distinguished cancellation signal" shape of change and not this feature's to make. The default the mode is compared against is still the one the session connected in, re-read on every `--session-reset`.

- **Zero is an off switch on these two and not on `--timeout`.** `parse_seconds()` refuses zero, correctly: a run has no deadline until a caller sets one, so there is nothing for zero to turn off there. These two default the other way round, so `_server_seconds()` folds a typed zero to `None` before delegating.

- **The secret warning is a warning.** Refusing needs env interpolation (#898) to have shipped, so that a container with no config file has somewhere else to put a password; §4.1 says to tighten it then. Detection goes through `redact_profile()`/`redact_conn_str()` rather than a flag-name list kept beside the check, and only over what the caller *typed*, since a profile is the remedy the warning points at.

Tradeoffs worth naming. The clocks are read between accepts, at `ACCEPT_POLL_SECONDS`, so an expiry can be up to half a second late; a timer thread would be exact and would be another thread around a connection nothing promises is safe to share. And a `--max-lifetime` expiry mid-query leaves that query unreachable until it finishes, since the socket is unlinked before the drain — pre-existing ordering from the `SIGINT` path, but this is the change that makes it happen unattended; §8 records it.

Byte-equivalence is untouched: the transaction note is the one new thing a served invocation can write that a cold one cannot, and it cannot fire for any in-tree adapter.

- [ ] No.
- [x] Yes; I haven't opened a PR, but the gist of the change is: the session section the "Headless & Agents" topic still needs (PR 6 of the plan) should say that a session stops itself after 30 minutes idle or 8 hours up, that `--idle-timeout 0` is the answer to "my session died between calls" — §8 predicts that as the first bug report — that `--session-status` reports both clocks, and that **a lifetime expiry mid-query leaves that query uninterruptible until it finishes**. It should also say plainly that an open transaction survives the invocation that opened it and that hsql will *not* warn about it, so `--session-reset` is the way back to a known state. The secret warning belongs beside the `-P` recommendation §4.1 asks the docs to lead with.

- [x] Yes.

- **The clocks, wound by hand rather than waited out** (`test_hsql_serve.py`) — a session with neither clock set never expires; one nobody has used runs out of idle time; a held turnstile stops the idle clock and giving the turn back restarts it; the lifetime clock fires on a busy session; the sooner of the two is the one that answers.
- **The status fields** — on an idle session, on a busy one (lifetime only), a countdown that has run out reading `0.0` rather than negative, and both `null` with the clocks off.
- **A client that has not spoken yet** — the count holds the clock, and releasing it lets the clock run out *from where it was* rather than restarting, which is what keeps a poll from being use. Plus an end-to-end test driving a real client that connects, waits past the timeout, then sends its script. Both confirmed failing on the code without the fix, the second with the reviewer's exact broken pipe.
- **A status poll is not use** — `_expired()` still says the session is idle after one.
- **End to end** — a session started with `--idle-timeout 1` exits 0 on its own with the expiry line and no socket left behind; one started with `--idle-timeout 0 --max-lifetime 2` answers a status saying its idle clock is off, then stops on its lifetime.
- **The transaction note** — a departure from the connected mode reported, and reported *again* on the next request; a session in the mode it connected in (and one whose adapter has no modes) silent; a reset re-basing the default; an abandoned connection never asked; an adapter that raises read as one with nothing to say. Plus one driving a real request frame through `_attend` over a `socketpair`. The `Moody` double's docstring now says why no shipping adapter reaches this.
- **The secret warning** — a declared `secret=` option and a password inside `CONN_STR`, the second exiting 3 so that the warning is shown to land *before* the connection; the warning asserted not to contain the secret; a profile's secret not warned about; a clean `--serve` silent.
- **The partition**, reparametrized — all three server options refused on a cold invocation, on a served request, and from a served request's profile.
- `format_duration()` over its cases, since the two defaults print as `30m` and `8h`.

Also here, and unrelated to the feature: `test_toggle_full_screen` pressed `f10` without awaiting the `ctrl+j` before it, so on a loaded runner the Results Viewer had no results yet, could not take focus, and `f10` full-screened the editor instead. It already takes the `wait_for_workers` fixture; it now awaits it after the query too. Seen failing on macOS 3.11 here.

Full suite green: 1893 passed / 7 skipped, mypy clean over 148 files, all five import-linter contracts kept, ruff clean. `docs/generated/hsql-reference.md` and `schemas/config-v1.json` regenerated for the two new options.

- [x] I have added an entry to `CHANGELOG.md`, under the `[Unreleased]` section heading. That entry references the issue closed by this PR. *(No issue exists for warm sessions, so the entries carry no link.)*

https://claude.ai/code/session_01Vyid3ewU2r25P87jYFkuGs


src/harlequin/hsql/server.py
```diff
@@ -379,17 +412,38 @@ def _transaction_mode(self) -> "str | None":
...
+        None where there is nothing to read: an adapter with no modes, a
+        session without a connection, or one whose connection belongs to a
+        thread that outlasted the cancel that was meant to stop it.
+        """
+        connection = self._connection
+        if connection is None or self._abandoned:
+            return None
+        return _transaction_mode_of(connection)
+
+    def _note_transaction_mode(self, segments: "list[tuple[int, bytearray]]") -> None:
```
tconbeer, line 436: This can't fire for any adapter that ships, and the gap is in what it's advertised as, not in the adapter contract.

`HarlequinConnection.transaction_mode` is documented at `adapter.py:203` as "the user-facing label of the currently-active transaction mode" — the Auto/Manual **setting**. It's a plain instance attribute (`harlequin_sqlite/adapter.py:309`) mutated only by `toggle_transaction_mode()`, which hsql never calls, and no SQL statement touches it. So a session can never leave the mode it connected in, and the only thing that can trigger this note is the `Moody` double at `test_hsql_serve.py:2235`.

Confirmed end to end on 3.12 + sqlite, the one in-tree adapter that has modes on that build:

```
$ hsql --session tx3 --session-status | grep transaction_mode
"transaction_mode":"Auto"
$ hsql --session tx3 -c "begin; insert into t values (1);"     # exit 0, no note
$ hsql --session tx3 -c "select count(*) from t;"              # exit 0, no note
 count(*)
----------
        1          # the uncommitted row: the transaction really is open across invocations
```

The transaction is open, every later request runs inside it, and hsql says nothing — the exact case the feature is sold on.

I'd leave the adapter contract alone and cut the claim back to what this actually observes. Concretely: keep `_note_transaction_mode()` as "the session's transaction *mode* is not the one it connected in," and drop the open-transaction framing from the docstring here (`a BEGIN left open is a session holding locks…`), from `diagnostics.report_transaction_mode()`'s text, and from the CHANGELOG entry. That leaves an honest, if narrow, feature — an operator who toggles a mode out of band still gets told — and no promise of a safety net that isn't there.

Worth noting the PR body reasons about a different gap ("the contract offers no way to ask an adapter which of its modes *is* its default"). The default-mode workaround is sound; it's the premise above it that doesn't hold.

---

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 - `hsql --serve NAME [CONN_STR]` holds a database connection open as a named session, and `hsql --session NAME -c ...` (or `HSQL_SESSION=NAME`) sends queries to it without paying start-up or connection cost again. Temp tables, settings and open transactions persist between invocations; `--session-reset` reconnects. POSIX only.
 - `hsql --session NAME --session-status` reports what a session is doing as JSON — adapter, connection, uptime, requests, busy or idle — and answers even while a query is running. A request that names a different database, adapter or connection option than the session connected with is refused.
+- A session now stops itself once it has gone 30 minutes with no request, or eight hours after it connected: `--idle-timeout SECONDS` and `--max-lifetime SECONDS` on `hsql --serve` set both, and `0` switches either off. `--session-status` reports them as `idle_timeout_s` and `expires_in_s`.
+- A session that is left in a transaction mode it did not connect in now says so on stderr after every request, so an open transaction holding locks is visible rather than something the next caller finds out about.
```
tconbeer: The second clause — "so an open transaction holding locks is visible rather than something the next caller finds out about" — is the part that doesn't ship. See the comment on `server.py:431`: what's compared is the adapter's Auto/Manual mode setting, which nothing in the hsql path can change, so a `begin` left open is silent.

Suggest trimming to the first clause, e.g.:

> - A session that is left in a transaction mode it did not connect in now says so on stderr after every request.

A user reading the current entry will expect `begin;` without a `commit` to be flagged, and it won't be.

---

src/harlequin/hsql/diagnostics.py
```diff
@@ -140,6 +140,72 @@ def report_session_ready(name: str, adapter: str, *, stream: TextIO) -> None:
...
+        else f"has been up for {format_duration(seconds)}"
+    )
+    note(
+        f"session {name!r} {ran_out}, so it is stopping ({flag}). "
+        f"Pass `{flag} 0` for a session that runs until something stops it.",
+        stream=stream,
+    )
+
+
+def report_transaction_mode(name: str, mode: str, *, stream: TextIO) -> None:
```
tconbeer, line 164: Same over-claim as the CHANGELOG entry (see `server.py:431`). The message tells the caller "What it holds open is held until the session is told otherwise" and prescribes `-c commit` / `-c rollback`, but the condition that produced it is a mode-label change, which says nothing about whether a transaction is open — and the case where one *is* open never reaches here.

As written this can only mislead: the invocation that sees it may have nothing open, and the invocation that does have something open never sees it. Worth rewording to report the mode change itself, and pointing at `--session-reset` as the reliable way back to a known state.

---

src/harlequin/hsql/server.py
```diff
@@ -694,6 +778,10 @@ def _attend(self, connection: socket.socket) -> None:
                 diagnostics.report_bad_request(str(e), stream=self._stderr)
                 return
 
+            # a request is what the idle clock counts; a status poll or a
+            # cancel is bookkeeping about a session rather than use of one,
+            # and a caller watching a session must not be what keeps it up
+            self._active_at = time.monotonic()
```
tconbeer, line 815: Because the clock is only wound here — after the REQUEST frame is parsed — a client that has connected but not yet sent one is invisible to `_expired()`: it isn't in the turnstile either, so neither half of the idle check can see it. The unclocked read above is deliberate ("a `-f -` sends the bytes a human is still typing"), but nothing keeps the session alive for that thread, and `client.py:233` reads stdin *after* connecting and shaking hands.

Reproduced end to end:

```
$ hsql --serve repro -a duckdb :memory: --idle-timeout 5 --max-lifetime 0 &
$ ( sleep 8; echo "select 1;" ) | hsql --session repro -f -
hsql: error: the session named 'repro' did not answer: [Errno 32] Broken pipe
CLIENT EXIT=3
```

The socket is unlinked under the connected client and the daemon `_attend` thread dies with the process.

The exposure isn't the microsecond race the PR body describes — it's the whole duration of the stdin read. At the 30-minute default it needs an idle session plus a slow producer: an interactive `-f -`, a slow upstream pipe, or an agent that opens the connection and then composes its query. The failure is also one of the more confusing ones to land on, since the session was alive when the client reached it.

Cheapest fix is to wind the clock on accept as well as here; alternatively have `_expired()` count accepted-but-unsubmitted connections, so a session with a client mid-handshake is never idle. The first keeps the property you want — a status poll still isn't use of a session, since `_send_status` returns without a request ever arriving.

---

src/harlequin/hsql/server.py
```diff
@@ -608,9 +665,36 @@ def _unlock(self) -> None:
...
     def _accept(self, listener: socket.socket) -> None:
         listener.settimeout(ACCEPT_POLL_SECONDS)
         while not self._stopping.is_set():
+            expired = self._expired()
+            if expired is not None:
+                flag, seconds = expired
+                diagnostics.report_session_expired(
+                    self.name, flag, seconds, stream=self._stderr
+                )
+                return
```
tconbeer, line 711: Minor, and pre-existing rather than introduced here — but this is the change that makes it reachable unattended.

Returning from `_accept` runs `serve()`'s `finally`, which closes the listener and unlinks the path *before* `_turnstile.enter(None)` drains. The in-flight request is still answered, as the PR says, but for as long as it runs nothing can reach it: `client._cancel()` opens a second connection to a path that's gone and swallows the ENOENT, so `Ctrl-C` exits 130 while the query runs on, and `--session-status` is equally unreachable.

Under `SIGINT` an operator is present and knows what they just did. Under `--max-lifetime` this now happens by itself at eight hours, to whoever happens to be mid-query, with no one at the server's terminal.

Not asking for a change in this PR — draining before the unlink would need care around a client racing in on the still-live socket. But it's worth a line in the §8 notes, and it argues for the docs saying that a lifetime expiry mid-query leaves the query uninterruptible until it finishes.

---

tconbeer (commented):
Reviewed at `cc994e5`.

**Verified locally:** `test_hsql_serve.py`, `test_hsql_session.py` and `test_hsql_equivalence.py` — 320 passed. `ruff format --check` and `ruff check` clean. mypy reports 1142 errors here, but 1122 of the same class (unresolved `textual`/`duckdb`/`tomli`) reproduce at `main`, so that's this container, not the diff. `lint-imports` isn't installed here.

What holds up: the option partition is extended in one place and the cold/served/serve refusal triangle is symmetric; value handling is right at the edges (`0` switches a clock off, negatives and non-numerics are refused at parse, fractional seconds work); the secret warning sits before connect and exits 3 so it lands ahead of a connection error; `config-v1.json` and `hsql-reference.md` are regenerated. The two-clocks design reads well, and `_expired()`'s split between them is the right call.

Three things, one of which I think blocks:

1. **The transaction-mode note can't fire for any adapter that ships**, and the CHANGELOG and the stderr message both promise open-transaction detection on the strength of it. `transaction_mode` is the Auto/Manual setting, not whether a transaction is open, and nothing in the hsql path can change it. Confirmed on 3.12 + sqlite: `begin; insert` left a transaction open across invocations and nothing was printed. My suggestion is to narrow what's advertised rather than reach into the adapter contract — details inline on `server.py:431`, with the two follow-on edits on `CHANGELOG.md` and `diagnostics.py`.

2. **The idle clock can kill a client that has connected but not yet sent its request** (`server.py:784`). `_active_at` is wound only after the REQUEST frame is parsed, and such a client isn't in the turnstile either, so `_expired()` can't see it. Reproduced: `( sleep 8; echo "select 1;" ) | hsql --session repro -f -` against `--idle-timeout 5` dies with a broken pipe and exit 3. The window is the whole stdin read, not the microsecond race the PR body describes.

3. **Minor / no change requested:** a `--max-lifetime` expiry unlinks the socket before draining, so the in-flight query is uninterruptible for the rest of its run (`server.py:697`). Pre-existing ordering from the `SIGINT` path, but this PR makes it happen unattended at eight hours.

---

src/harlequin/hsql/server.py
tconbeer: Confirmed, and you're right that the premise is the part that fails. Fixed in 86ed0f6.

I verified it independently before changing anything: `transaction_mode` returns `self._transaction_mode` (`harlequin_sqlite` line 309), a plain attribute written only in `toggle_transaction_mode()`, and `grep -rn toggle_transaction_mode src/harlequin/` finds callers only in `app.py` — none under `hsql/`. DuckDB's adapter has zero occurrences of `transaction_mode` at all. So in the hsql path the label is immutable, exactly as you say.

Taken as recommended — narrow the claim, leave the contract alone:

- `_note_transaction_mode()`'s docstring now says the mode is the Auto/Manual setting, that nothing hsql runs changes it, and that this therefore reports a session **moved out of band** and says nothing about what it may be holding.
- `report_transaction_mode()` reworded (thread below).
- CHANGELOG trimmed to your suggested wording.

Two places in the plan carry it too, since the premise is the plan's rather than the PR's:

- **§5 item 2** now records that its own premise does not hold — the worst failure mode is *not* made visible, `--session-reset` (item 3) is the only thing in the design that answers an open transaction, and detecting one needs a new adapter-contract member, which is §7's "distinguished cancellation signal" shape of change.
- **§9's PR 5 entry** notes that the default-mode workaround was sound but the premise above it was not.

On your last paragraph — that's the honest summary of my error. I reasoned carefully about how to pick a *default* to compare against and never checked whether the value could move at all. The test double made it look like a working feature.

---

src/harlequin/hsql/server.py
tconbeer: Real bug, reproduced verbatim before fixing it — same broken pipe, same exit 3. Fixed in 86ed0f6.

I took your **second** option rather than the first, because the first doesn't hold the property after all. Winding `_active_at` on accept *resets* the clock, and a status poll is an accepted connection — so a caller polling every 10s against the 30m default would reset it every 10s and the session would never idle out. `_send_status` returning without a request doesn't help: the damage is done at accept, before the frame is read.

Counting connections *pauses* the clock instead, which distinguishes the two cases by duration rather than by kind:

- a status poll holds it for the microseconds it takes to answer, then the clock resumes from where it was — the session still idles out on schedule;
- a `-f -` client holds it for the whole stdin read, which is the case you found.

So `Server` keeps an `_attending` count (incremented in `_accept` before the thread is spawned, so there is no window; released in `_attend`'s `finally`), and `_expired()` treats a session with any client connected as not idle.

Verified both properties end to end, not just the fix:

```
# the reproduction, now
$ ( sleep 8; echo "select 1;" ) | hsql --session repro -f -     # --idle-timeout 5
 1
CLIENT EXIT=0
note: request 1: exit 0 in 214ms
note: session 'repro' has had no request for 5s, so it is stopping   # still idles out after

# and a poll still isn't use: --idle-timeout 5, polled every 1s for 10s
note: session 'poll' has had no request for 5s, so it is stopping (--idle-timeout).
note: session 'poll' stopped after 0 requests.
```

Two tests, both confirmed failing on the unfixed code (the second with your exact broken pipe): a unit test that the count holds the clock and that it resumes rather than restarts, and an end-to-end one that drives a real client which connects, waits past the timeout, then sends its script.

You're also right that the PR body was wrong about the exposure — it described a microsecond race between accept and the turnstile. The window is the whole stdin read. §9 now records that, and says the clock is held rather than reset and why.

---

src/harlequin/hsql/diagnostics.py
tconbeer: Agreed — and your framing of why it's worse than a harmless over-claim is the part that decided it: the invocation that sees the note may have nothing open, and the one that does never sees it. A message that is wrong in both directions is worse than no message.

Reworded to report the setting and nothing else, and to point at `--session-reset`:

> note: session 'prod' is in transaction mode 'Manual', which is not the one it connected in. Every request runs under it until the session is told otherwise; `hsql --session prod --session-reset` reconnects it in the mode it started in.

`-c commit` / `-c rollback` are gone, since this condition gives no reason to think there is anything to commit. The docstring says the same: the mode is the adapter's own Auto/Manual setting rather than a statement about open work.

---

CHANGELOG.md
tconbeer: Taken verbatim in 86ed0f6 — the entry is now exactly your suggested line:

> - A session that is left in a transaction mode it did not connect in now says so on stderr after every request.

The second clause is gone. Reasoning on the thread at `server.py`.

---

src/harlequin/hsql/server.py
tconbeer: Noted as asked — no code change, and §8 now carries it (86ed0f6):

> **A lifetime expiry mid-query leaves that query unreachable.** The expiry path is the stop signal's, which closes the listener and unlinks the socket *before* it drains: the request in flight is still answered, but for the rest of its run no cancel and no `--session-status` can reach it, because both open a second connection to a path that is gone. Under `SIGINT` an operator is present and knows what they just did; under `--max-lifetime` it happens by itself, at eight hours, to whoever is mid-query. Draining before the unlink would need care around a client racing in on the still-live socket, so PR 5 left the ordering alone — but the docs should say that a lifetime expiry mid-query leaves the query uninterruptible until it finishes.

§9's PR 5 entry also now says the lifetime path lets the in-flight request finish "at the cost §8 now records," so the tradeoff isn't stated only in the optimistic direction.

I've added the docs line to the PR-6 gist in the PR body as well, so it doesn't get lost between here and the "Headless & Agents" topic.

Agreed on leaving the ordering: unlinking after the drain opens a window where a client connects to a socket that is about to stop answering, which trades a visible failure for a hanging one.

---

PR #1155, opened 2026-09-14
M4 7/14: hsql reads the query log

tconbeer:
Closes the headless half of [#429](https://github.com/tconbeer/harlequin/issues/429). PR 7 of [the M4 plan](https://github.com/tconbeer/harlequin/blob/main/docs/plans/m4-hsql-technical-plan.md#release-b--query-history-and-search-214), on top of #1135.

- **`hsql --history`** lists what both commands have run, newest first; **`hsql --history-search TERM`** lists the ones whose SQL mentions TERM. Two modes beside `--catalog` and `--catalog-search`, refused together by `_one_mode()` like every other pair.
- **`query_log.recent()`** is the read: one indexed `select … where connection = ? [and sql like ?] order by id desc limit ?`. `READ_COLUMNS` names what comes back. It creates and migrates nothing — a store that is not there yet is a history of nothing, and one that is there and cannot be read raises for the caller to report as exit 2.
- **`query.typed_rows_to_result()`**, a sibling of `rows_to_result()` for rows that did come from a database. Nothing new in the output layer: every `--format`, the psql switches, `-o PATH`, `--display-rows` and `-x` work on the listing the day it lands.
- **`elapsed_ms` is rounded to microseconds on the way into the store**, where it is written once rather than in the listing that prints it every time it is read.

```console
$ hsql --history --limit 2
 run_at                           | program   | profile | adapter | status | rows | elapsed_ms | sql
----------------------------------+-----------+---------+---------+--------+------+------------+------------------------------
 2026-09-14T20:44:45.563989+00:00 | hsql      | prod    | postgres| ok     | 1    | 0.012      | select count(*) from orders
 2026-09-14T20:44:45.342956+00:00 | harlequin | prod    | postgres| ok     | 42   | 0.89       | select * from line_items
(2 of >2 rows)
```

- **These are queries, not listings.** `--catalog` has to synthesize a result set out of a walk, which is what `rows_to_result()` exists for. `--history` runs SQL against a SQLite database, so its rows go where a cursor's rows go — `create_backend(rows, column_names=…)`, the same call `fetch()` makes. That is what makes `--json` emit `"rows":42,"elapsed_ms":12.5` rather than strings an agent has to parse back. The price is that `text_columns()` no longer short-circuits on an all-string table, so a text layout casts through duckdb: +112ms to import it, 1.8ms to cast 500 rows.
- **`--limit` is honored here rather than refused as it is under `--catalog`**: the store has a row count, so a hard limit on it is exact. It fetches one row more than it keeps, the same probe a query's limit fetches, which is what makes the `(2 of >2 rows)` above true. Defaults to 500, `-1` for all.
- **The only database they open is the store.** `-P`, `-a` or a `CONN_STR` narrows to one connection, and the invocation's adapter is constructed solely to be asked for its `connection_id` — never connected with. A path that exits 3 for a query is an empty listing here. With none of the three named, every connection is in the answer, because "I typed `hsql --history` and got someone else's idea of the default database" is the surprising one; the `profile` and `adapter` columns tell them apart.
- **The branch sits after `_open_tunnel()`, like `--catalog`.** A tunnel is part of a connection's identity — `cache_material()` reads what `ssh -G` resolved, so it cannot be reproduced without opening one — and a tunneled profile has to key the same way a run does or it narrows to nothing. It is a no-op for anyone without `ssh_host` set, but it does mean `hsql -P tunneled --history` opens the tunnel to read a local file. The alternative was silently wrong narrowing; happy to revisit if the tunnel is the worse surprise.
- **`sql` is folded to one line in every format**, in the mode rather than the layout: `layout.py` pads by terminal cells and has no concept of a cell spanning rows, and `--format table` and `--format csv` agree cell for cell, so a column that is verbatim in one and folded in the other is not on offer. Folding loses the formatting, not the query.
- **A search term's `_` and `%` are escaped.** `_` matches any character in a `like` and is in half the table names there are; `line_items` must not also find `lineXitems`. Case is left to SQLite's ASCII-insensitive `like`, which is what someone hunting for a table name wants, and is pinned by a test.
- **`sqlite3` is imported inside `_report_history()`**, not at module scope: it is 4ms on every invocation that reads no store.
- `docs/cli-options.md`'s step 4 said every mode belongs in the `needs_profile` / `needs_adapter` predicates, which is wrong for a mode configured like a run — corrected, naming `--catalog` and `--history`.

- [x] Yes; I haven't opened a PR, but the gist of the change is: the query history page — where it lives, what it holds, how to search it from either front end, and that it is a SQLite file you can open in Harlequin itself — plus `--history` in "Headless & Agents". That is PR 10 of the M4 plan, with the rest of Release B's docs. The Agent Skill that ships in this wheel is updated here: `SKILL.md` §3 teaches both modes, `references/queries.md` gains a "Reading the query history" section, and `allowed-tools` pre-approves them beside the other modes that only read.

- [x] Yes. 23 in `test_hsql.py` for the two modes (narrowing by profile and by CONN_STR with two connections interleaved, `--limit` from the newest end and its `>`, the folded `sql` matching between `table` and `csv`, `--json` keeping its types, the underscore, the case, a store that is not there, a store that is not a database, and that the modes neither connect nor record themselves), 10 in `test_query_log.py` for `recent()` and the rounding, and 3 in `test_query.py` for `typed_rows_to_result()`.

https://claude.ai/code/session_01Qqx91xCxzfj2bNkxXKEtiz

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qqx91xCxzfj2bNkxXKEtiz)_


src/harlequin/hsql/modes/history.py
```diff
@@ -0,0 +1,96 @@
...
+
+def folded(sql: str) -> str:
+    """One query on one line, which is the only shape a listing has.
+
+    `layout.py` pads by terminal cells and has no concept of a cell spanning
+    rows, and `--format table` and `--format csv` agree cell for cell -- so a
+    column that is verbatim in one and folded in the other is not on offer.
+    Folding loses the formatting and not the query: it still runs.
+    """
+    return " ".join(sql.split())
```
tconbeer: `" ".join(sql.split())` doesn't preserve the query — it changes what the SQL means, in two ways, and the docstring above ("Folding loses the formatting and not the query: it still runs") and `references/queries.md` ("The SQL is folded onto one line in every format; it still runs") both promise the opposite.

**A `--` line comment swallows the rest of the statement.** Logged SQL:

```sql
select 1 as a, -- the first column
2 as b
```

lists (in every format, `--json` and `--csv` included) as:

```
select 1 as a, -- the first column 2 as b
```

Everything after `--` is now commented out, so the query an agent copies back is `select 1 as a,` — a syntax error. `split()` has no idea `--` runs to end-of-line.

**Whitespace inside string literals is rewritten.** `select 'hello  world'` lists as `select 'hello world'`, and a literal containing a newline or a tab is silently changed too. That one still runs, and returns a different value.

This isn't cosmetic here, because the new `SKILL.md` §3 tells agents to read the history "before writing a query from scratch," so the listed text is meant to be reused.

Folding that keeps the query intact needs to know where comments and literals are. `harlequin.statements` already drives tree-sitter-sql; a fold that skips (or strips) `comment` nodes and leaves string-literal nodes byte-for-byte would hold up. Failing that, the honest fix is to narrow the claim: drop "it still runs" from both the docstring and `queries.md`, and say the `sql` column is a searchable rendering rather than a runnable one.

---

src/harlequin/hsql/cli.py
```diff
@@ -1003,6 +1035,58 @@ def inner_cli(
...
+                history_limit = parse_row_count(raw_limit, key="--limit")
+            except HarlequinConfigError as e:
+                diagnostics.report_error(e)
+                ctx.exit(ExitCode.USAGE)
+            # -P, -a or a CONN_STR names one database and narrows the store to
+            # it; with none of them the answer is every connection this machine
+            # has used, because "I typed `hsql --history` and got someone
+            # else's idea of the default database" is the surprising one.
+            named_a_database = (
+                profile is not None or "adapter" in explicitly_set or bool(conn_str)
```
tconbeer: The three arms of this test don't ask the same question, so a config file the caller never mentioned can narrow the listing.

`profile` is the `-P` parameter and `"adapter" in explicitly_set` is CLI-only, but `conn_str` was popped off `values` at ~line 780 — the *merged* dict. `merge_profile_with_cli()` deliberately lets a profile's `conn_str` through when the CLI's is empty (it's an argument, so click always reports it as typed), so `bool(conn_str)` is true for a value that came out of a config file.

Concretely, with this in a discovered config file:

```toml
default_profile = "prod"

[profiles.prod]
adapter = "postgres"
conn_str = ["postgresql://..."]
```

a bare `hsql --history` — in that directory, with no `-P`, no `-a`, no CONN_STR — silently narrows to `prod`'s connection, and the caller gets exactly the "someone else's idea of the default database" the comment two lines up says this is here to prevent. `--help` ("`-P`, `-a` or a CONN_STR narrows it to one database") and `references/queries.md` say the same thing. Worse, the answer depends on the directory you run from: `cd` out of the project and the same command lists every connection.

The asymmetry is its own bug even setting the discovery case aside: a `default_profile` that sets `conn_str` narrows, while one that sets only `adapter` (plus `host`/`port`/`dbname`) does not — two profiles that name the same single database, treated differently.

Whichever rule is intended, all three arms should read it off the same source. For "typed on this command line":

```suggestion
            named_a_database = (
                profile is not None
                or "adapter" in explicitly_set
                or "conn_str" in explicitly_set
            )
```

(`explicitly_set` is over `kwargs`, and `conn_str` is in it whenever click saw the argument; the empty-tuple case is what `merge_profile_with_cli` filters, so pair this with a `bool(conn_str)` guard if an empty `CONN_STR` needs to not count.) If instead the intent is "this invocation resolves to one database, however it was configured," then `adapter` should be read off the merged value too — but that reverses the default for every user who has a `default_profile`, and the `--help` text needs to change with it.

---

src/harlequin/hsql/skill/references/queries.md
```diff
@@ -29,6 +30,32 @@ hsql --info -a postgres    # look at .adapters.postgres.capabilities
...
+Eight columns: `run_at` (UTC), `program` (`hsql` or `harlequin`), `profile`, `adapter`,
+`status` (`ok`, `error` or `canceled`), `rows`, `elapsed_ms`, and `sql`. The SQL is
+folded onto one line in every format; it still runs.
+
+`-P`, `-a` or a CONN_STR narrows the listing to that one database. With none of them,
+every database you have used is in it, and the `profile` and `adapter` columns are what
+tell them apart. `--limit` defaults to `500` and means "the most recent N".
+
+Neither mode connects to a database, so both work when the warehouse is down or you
+have no credentials for it. `--no-write-history` keeps one run out of the record.
```
tconbeer: "Neither mode connects to a database, so both work when the warehouse is down or you have no credentials for it" is not true for a tunneled profile, and the PR description already says why: the branch sits *after* `_open_tunnel()`.

In `cli.py`, `tunnel = _open_tunnel(ctx, ssh_config)` is line 1008 and `if history or history_search is not None:` is line 1038. `_open_tunnel()` is not a no-op when the profile sets `ssh_host` — it shells out to `ssh`, and on failure calls `diagnostics.report_error(e); ctx.exit(ExitCode.CONNECTION)`. So `hsql -P tunneled --history` exits **3** without ever reading the store, which is precisely the "warehouse is down" case this sentence promises works. It can also block on an `ssh` passphrase prompt before printing a row.

This one matters more than a normal doc slip: it ships in the Agent Skill, and `troubleshooting.md` routes by exit code — an agent told the mode connects to nothing has no reason to read a 3 from `--history` as anything but a bug.

Either qualify the claim:

```suggestion
Neither mode connects to the database itself, so both work when the warehouse is down or
you have no credentials for it — but a profile with `ssh_host` still opens its tunnel
first, so add `-a duckdb` (or drop `-P`) to read the history when the tunnel is down.
`--no-write-history` keeps one run out of the record.
```

or move the `history` branch above `_open_tunnel()` and drop the narrowing-by-connection-id for a tunneled profile, which is the tradeoff the PR description already weighs.

---

tconbeer (commented):
Reviewed the diff for correctness. Three findings inline, plus one that falls outside a diff hunk so I could not anchor it:

**`src/harlequin/hsql/cli.py:964`** (context, not in a hunk) — the comment `# every mode below this connects to the database` is now false. The new `--history` / `--history-search` branch sits 74 lines below it and connects to nothing; the callback's own `connects` predicate at line ~720 already says so. Every other non-connecting mode (`--skill`, `--info`, `--spec`, `--config`) is above this line, so the comment reads as the section boundary it no longer is. Worth either moving the branch above the line or amending the comment to name the exception — it's also what makes the tunnel surprise in `queries.md` easy to miss.

The three inline findings, most severe first:

1. `modes/history.py:45` — `folded()` merges a `--` line comment into the following code and rewrites whitespace inside string literals, contradicting "it still runs" in both the docstring and the new skill reference.
2. `cli.py:1054` — `named_a_database` mixes a merged value (`conn_str`) with typed-only checks, so a discovered `default_profile` silently narrows `hsql --history` to one database.
3. `skill/references/queries.md:56` — "Neither mode connects to a database" is false for a profile with `ssh_host`.

Everything else in the diff held up: `recent()`'s escaping (`\` before `%`/`_`, escape char first), the `-1`-as-no-limit shape, the `version < 1` guard, the `limit + 1` probe and its `fetched_row_count` excluding the probe row, and `_rounded()` at both write sites.

---

src/harlequin/hsql/modes/history.py
tconbeer: You're right on both counts, and the `--` case is the bad one: the listed query is silently truncated at the first line comment, and the skill tells an agent to reuse it. Fixed by folding with the grammar rather than with `split()`, in 7fc03be.

`statements.fold()` is new, beside `split()` and for the same reason — only a parse knows where a comment ends and a literal begins. It captures `(comment)`, `(marginalia)`, `(literal)` and `(dollar_quote)`, drops the comments, keeps everything else byte for byte, and collapses whitespace only in the gaps between them. A dropped comment leaves a space behind only where the tokens either side would otherwise run together (`select 1--why\n+2` → `select 1 +2`), and a span inside one already taken is skipped, so a comment inside a dollar-quoted body is body content.

```
'select 1 as a, -- the first column\n2 as b'  -> 'select 1 as a, 2 as b'
"select 'hello  world' as x,\n  2"            -> "select 'hello  world' as x, 2"
'select $$a  b$$ as x,\n  2'                  -> 'select $$a  b$$ as x, 2'
```

Two things I could not make true, both now stated rather than promised away:

- **A literal that spans lines is joined.** `'multi\nline'` becomes `'multi line'`. One line per row is what the layouts need, and standard SQL has no portable escape to write that newline with — `--format table` and `--format csv` already disagree cell for cell on any query returning a multi-line value, and a logged query is the one case where that is the norm rather than the exception. Only vertical whitespace is touched inside a literal; horizontal spacing survives.
- **Comments are dropped, so the text a `like` matched is not always the text printed.** Searching for a term that appears only in a comment returns a row that does not visibly contain it. Both are in `fold()`'s docstring, and `queries.md` now says "comments are dropped so a `--` one cannot comment out the rest of the line, and a literal keeps its own spacing".

Cost: a statement already on one line of single spaces is returned as it stands with no parse, which is every statement typed at a shell. The rest pay ~15µs each — 7.3ms for 500 rows, against the ~112ms the typed columns already cost in duckdb.

16 fold cases in `test_statements.py`, each also asserted to be one line and to be a fixed point; one that runs a folded statement against DuckDB and checks the values come back; and two in `test_hsql.py` that run a commented query, list it, and then run what the listing printed.

---

src/harlequin/hsql/cli.py
tconbeer: Confirmed and fixed in 7fc03be — I read the merged value by accident, and the directory-dependence is the part that makes it indefensible.

Taking the "typed on this command line" reading, since that is what `--help` and `queries.md` already say, and what `_refuse_the_sessions_options` does with a `-P` for the same reason ("a profile discovered in the caller's directory describes that directory, not this session"):

```python
named_a_database = (
    profile is not None
    or "adapter" in explicitly_set
    or bool(kwargs.get("conn_str"))
)
```

`kwargs` rather than `explicitly_set` for the third arm, exactly as you flagged: `conn_str` is an argument, so it is in `explicitly_set` on every invocation, and `kwargs["conn_str"]` is the CLI's own tuple before the merge — so the empty-tuple case falls out without a second guard. That also fixes the asymmetry: a `default_profile` now narrows nothing, whether it sets `conn_str` or only `adapter`.

A typed `-P prod` still narrows, and still resolves through the merge, so `-a` typed beside a discovered profile's `conn_str` keys the database that invocation would actually have connected to.

`test_a_discovered_profile_does_not_narrow_the_history` writes a `default_profile` naming one of two databases and asserts all three rows come back.

---

src/harlequin/hsql/skill/references/queries.md
tconbeer: Took the second option — moved the branch above `_open_tunnel()` — rather than qualifying the sentence, in 7fc03be.

What decided it: the sentence is the mode's whole point, and `ExitCode.CONNECTION` out of a mode whose module docstring says it "reads a database without connecting to one" is the kind of thing an agent cannot reason about. The tunnel is also a side effect on the user's machine — a child `ssh`, a bound local port, possibly a passphrase prompt — paid to read a local SQLite file.

So the narrowing stands down instead, and says why:

```console
$ hsql -P tunneled --history -tA
note: this connection is reached through bastion.example.invalid, and naming which database
that is means opening the tunnel, so the history below is every connection rather than that one
2026-09-14T21:31:23.250424+00:00|hsql|NULL|duckdb|ok|1|0.638|select 'hello  world' as x
2026-09-14T21:31:22.987981+00:00|hsql|NULL|duckdb|ok|1|1.026|select 1 as a, 2 as b
```

0.33s, exit 0, no `ssh`. Not silently narrowing to an id that matches no row was the other thing to avoid — that failure mode reads as "this database has run nothing", which is worse than either.

Two things fall out. The branch is back above `# every mode below this connects to the database`, so your fourth finding goes with it — that comment is true again and I did not have to amend it. And `_refuse_undeclared_read_only` / `_refuse_undeclared_timeout` no longer run for these modes, which is right: both are about a connection this one never makes.

`queries.md` now reads "Neither mode connects to a database or opens a tunnel", and the narrowing paragraph names the tunneled case alongside the `default_profile` one. `test_a_tunneled_profile_is_not_narrowed_and_says_so` pins the exit code, the full listing, and the note.

---

tconbeer:
All four findings fixed in 7fc03be; replies on each thread with the details.

- **`folded()` broke the query it printed.** Replaced by `statements.fold()`, which drives the same grammar `split()` does: comments dropped, literals kept byte for byte, whitespace collapsed only in the gaps. A statement already on one line of single spaces is returned untouched with no parse. Two residual limits are now stated rather than promised away — a literal spanning lines is joined (one line per row is what the layouts need, and SQL has no portable escape for that newline), and dropped comments mean the text a `like` matched is not always the text printed.
- **`named_a_database` mixed merged and typed values.** All three arms now read the command line, `conn_str` off `kwargs` rather than `explicitly_set` so the always-present argument does not count. A `default_profile` narrows nothing, whichever keys it sets.
- **The tunnel.** Took your second option: the branch moved above `_open_tunnel()`, so these modes start no `ssh` and cannot exit 3. A tunneled profile is reported unnarrowed with a note on stderr saying why, rather than narrowed to an id no row carries.
- **`# every mode below this connects to the database`** is true again — the move put the branch back above it, so there was nothing to amend.

Also updated `AGENTS.md`'s execution-core section to name `fold()` beside `split()` and say why it needs the grammar.

Tests: 16 fold cases in `test_statements.py`, each also asserted to be one line and a fixed point, plus one that runs a folded statement against DuckDB; and four in `test_hsql.py` — a commented query listed and then re-run from the listing, a literal's spacing surviving, a `default_profile` not narrowing, and a tunneled profile reported with its note. Full suite 2073 passed / 7 skipped, plus mypy and `lint-imports`.

---

tconbeer:
Rebased onto `main` (now at 6937734, 2.14.0). Three commits, force-pushed; the review fixes above are unchanged, and the only content decisions the rebase forced were two conflicts:

**`CHANGELOG.md`** — 2.14.0 shipped the query log, so `[Unreleased]` was empty. The `--history` entry moved there under a fresh `### Features`; `[2.14.0]` is byte-for-byte what #1156 published. So these modes land in 2.15, not the 2.14 the M4 plan's Release B assumed.

**`SKILL.md`** — #1154 added `sessions.md` and rewrote the body, which left it at 4092 bytes against `MAX_BODY_BYTES = 4096`. I took your five-reference sentence as written and spent a while trying to fit a `--history` line into §3 on top of it; every version of that meant shaving a clause off one of your rules (the `--catalog-search` capability caveat, `--limit`'s "unless told otherwise"), which is a worse skill in exchange for a round number. So the body is now yours verbatim except one word in the reference index:

```
`queries.md` (running SQL, catalog, history), `config.md` (config files and
```

and the discovery surface carries the rest instead — `description:` gained "read the history of queries run against it" (frontmatter is outside the body budget and is what decides whether the skill fires at all), `allowed-tools` pre-approves both modes, and `references/queries.md` has the full section. Body is 4089 bytes, 7 under.

The tradeoff: an agent that never opens a reference no longer gets the "read the history before writing a query from scratch" nudge in the always-read text, which is the half of B7 that fires unprompted. If you want it back, the honest way is +64 on the cap rather than a trim elsewhere — your call, and I did not make it unilaterally.

Checks after the rebase: 2074 passed / 7 skipped, the py12 run, mypy, `lint-imports`, and both generated artifacts regenerate to no diff.

---

tconbeer:
Dropped the fold and fixed the layout instead, in 960554e. The listing now reports the SQL the store holds, byte for byte.

**`layout.py` does what psql does.** The defect was never history's: `hsql -c "select 'a
b'"` breaks the table today for any query — the row runs into the next one and the column rule is padded to the length of the whole value. The two aligned layouts now span the cell across physical lines, mark each continuing one with `+` in the separator, blank the columns beside it, and measure a column by its widest line rather than its longest value. Unaligned writes the value as it stands, which is psql's answer there too; markdown already escaped a newline to `<br>` and is unchanged.

Asserted against psql 16's own output for the same data, which it now matches byte for byte inside the table:

```
 n | status | sql
---+--------+---------------
 1 | ok     | select       +
   |        |   o.id,      +
   |        |   o.total    +
   |        | from orders o
 2 | err    | select 1
(2 rows)
```

and in `-x`, and for the awkward case of a value ending in a newline.

**What that bought.** `statements.fold()` is gone — 89 lines and 19 tests — and with it all three caveats, because all three were consequences of folding: comments are no longer dropped, a literal spanning lines is no longer joined, and the text a `like` matched is the text printed. `--history` hands back exactly what ran, so what a listing prints runs.

The one thing I checked before committing to it: `test_the_layouts_and_the_files_agree_cell_for_cell` compares `text_columns()` values against the **parsed** csv values, so it is an invariant about the data, not the rendering — a continuation wrap doesn't touch it, and the fold was changing the value in every format when the invariant only ever asked that the value be the same everywhere. The golden corpus holds no multi-line string, so no snapshot churn; the new cases are exact-string tests in `test_layout.py`, where layout behavior is tested.

Residual, and it is psql's too: `-tA` prints those newlines as they are, so one query is more than one line there. `references/queries.md` now says so and points at `--jsonl`.

Two changelog entries under `[Unreleased]` — the modes under Features, the wrap under Bug Fixes, since it fixes every query's output and not just this listing.

2024 passed / 7 skipped, mypy, `lint-imports`.

---

Issue #485, opened 2024-03-24
Pressing `F8` when already viewing the history screen crashes the app


mattgiles:
**Before Proceeding, please acknowledge**:
- [x] I have consulted the [Troubleshooting guide](https://harlequin.sh/docs/troubleshooting/index).
- [x] I have searched Issues and Discussions in this repo.
- [x] Feature requests should be initiated as Discussions. This is a bug report.

**Describe the bug**
Pressing `F8` when already viewing `history` crashes the app.

**To Reproduce**
1. Launch harlequin
2. Press `F8` to view history screen
3. Press `F8` again

**Expected behavior**
One of: (a) nothing, or (b) exit history screen, as when `esc` is pressed.

**Actual behavior**
The app crashes with a stack trace (hard to paste rich-rendered stack traces here) that concludes:

```
DuplicateIds: Tried to insert a widget with ID 'history_screen', but a widget HistoryScreen(id='history_screen') already exists with that ID in this list of children. The children of a widget must have unique IDs.
```

**Contributing**
Are you interested in contributing a fix?
- [ ] Yes
- [x] Maybe
- [ ] No

**Additional context**
*Please provide as much as you can, but don't waste your time if it's definitely not relevant or you don't know.*

What is the output of `harlequin --version`?

```
$ rye run harlequin --version
harlequin, version 1.16.0

Installed Adapters:
  - duckdb, version 1.16.0
  - sqlite, version 1.16.0
```

What database adapter are you using with Harlequin? (Default is `duckdb`)

duckdb

What other options are you using when invoking Harlequin? (If you are using a profile, please add relevant items from your profile here).

Can you tell us more about your system?
- Shell: zsh
- Terminal: iTerm2
- OS of the shell: macOS
- OS of the terminal (if different from the shell):
- Default locale / language for your OS/Terminal/Shell: en_US.UTF-8

> [!TIP]
> For example, for my system, these are:
> - Bash
> - Windows Terminal
> - Ubuntu 22.04 / WSL2
> - Windows 11
> - en_US/UTF-8

How did you install Harlequin?
- [ ] pipx
- [ ] conda/mamba
- [ ] pip (global install)
- [ ] pip + venv/virtualenv or pipenv
- [ ] Poetry
- [x] uv/Rye
- [ ] other: _____________

tconbeer:
Thanks for the report!

PR #1160, opened 2026-09-15
M4 8/14: the History screen reads the query log

tconbeer:
Part of [#429](https://github.com/tconbeer/harlequin/issues/429) — PR 8 of M4. It does not close it: the filter input in the History screen is PR 9.

The History screen was a pickled deque inside the catalog cache, written in exactly one place: `action_quit`. A session that crashed lost its queries, and nothing `hsql` ran ever appeared in it. The screen now reads the SQLite store both commands already write, so `F8` shows an agent's queries beside the human's, and a session that never quit cleanly keeps both.

- `History.recent(connection=…, limit=500, search=None)` is one indexed read of `harlequin.query_log`, narrowed to this connection, run on a worker thread and posted back as `QueryHistoryLoaded`. The screen can no longer be "not yet loaded", so that error modal is gone.
- `QueryExecution` carries the store's `status`, so a canceled query no longer renders as an error, and a statement whose rows were never counted — DDL, or a session that died mid-fetch — renders without a row count or a duration. Its renderer is now one method the screen's `HistoryOption` calls, rather than two copies of the same layout.
- `CatalogCache.history` is removed and `CACHE_VERSION` goes to 3. A version-2 cache's records for one connection are moved into the store once, with their own timestamps and `program = 'harlequin'`, by `query_log.adopt()`.

The reasoning is in [`docs/plans/m4-hsql-technical-plan.md`](docs/plans/m4-hsql-technical-plan.md) §3.2 and §3.4. The short version:

- **The read is on a worker, not in the action.** All database work in this app happens on threads that never touch widgets, and the store is a database. It waits only `UI_BUSY_TIMEOUT_MS` for a lock another process holds, so a worker is not held up by someone else's write.
- **The move out of the pickle is one transaction, keyed on a marker the store holds.** `query_log.adopt()` writes the rows and a row in `migrated_connections` under one `begin immediate`. The first shape of this asked "does the store hold a row for this connection?" instead, which was wrong three ways — review reproduced all three: a single `hsql` run against a database made the guard true forever and stranded the pickle, two sessions starting at once each moved the whole thing, and a move that failed part-way left rows that looked like a finished one. A marker makes "already adopted" a recorded fact, and the transaction makes a failure retryable. Once a connection is adopted its version-2 pickle is chmodded `0600` — it holds every statement it recorded, and nothing redacted them. It is not deleted: the contents are in the store either way, and removing a user's file is the harder step to take back.
- **Timestamps are stored in UTC and rendered in the reader's zone**, which is where they ran the query. That is also what lets the snapshot test pin a row: it anchors on a local wall-clock time, so a record made anywhere renders the same.
- **`QueryError.query_text` and `ResultsFetched.errors`' SQL are left in place** although the history was their only reader. They describe the event for the error-modal path, and trimming message payloads is outside what this change is about.
- **DDL statements now show no duration on the screen**, and a result's duration is its own fetch rather than wall-clock from submission. That is what `hsql` already records; for an adapter whose `execute()` does the server-side work it under-reports, and it is the one behavior change here I have left open rather than settled.

- [ ] No.
- [x] Yes; I haven't opened a PR, but the gist of the change is: the query history — where it lives, what it holds, and that the History screen and `hsql --history` now read the same store — is the release's docs PR (PR 10 of M4), which covers this and `hsql --history` together.

- [x] Yes.

`tests/unit_tests/test_history.py` covers the row→record mapping, the connection filter, search, local-time rendering, canceled vs. error, a run with no row count, and an unreadable timestamp costing only its own row; plus the move — it happens once across processes, still happens for a connection `hsql` wrote to first, and no-ops on a pickle this version cannot unpickle. `tests/unit_tests/test_query_log.py` pins `adopt()` itself: three processes race it and produce one set of rows, and a failure mid-batch leaves the store untouched and retryable. Functional tests assert that an `hsql` row shows up in the screen, that another connection's does not, that the screen opens before anything has been run, and that a pickled history arrives through the move.

The migration tests load `tests/data/catalog-cache-2.pickle`, written by a released Harlequin 2.14.0 via `scripts/write_legacy_cache.py`, so they are pinned to the on-disk shape rather than to today's classes.

https://claude.ai/code/session_01CMq3y911Y5GzzfumhN8NEa


tconbeer:
CI on `acc1619`: 14 of 15 test jobs green. **`Windows-latest - 3.12`** failed on one test:

```
FAILED tests/functional_tests/test_crash.py::test_a_crash_prints_a_panel_and_not_a_traceback
  - AssertionError: assert 'Harlequin crashed.' in ''
```

I do not believe it is this PR's, and I have not pushed a fix. Three things say so:

- **The test cannot reach what this PR changes.** It is not marked `use_cache`, so `no_use_catalog_cache` has already replaced `harlequin.app.update_catalog_cache` with a no-op for it — and that call, plus a docstring, is the whole of this PR's footprint on the crash path. The assertion is on the panel Textual prints from `_exit_renderables` at shutdown, which this diff does not touch.
- **The same job passed on the pre-rebase head** (`abdd3fc`, [job](https://github.com/tconbeer/harlequin/actions/runs/34923972931/job/104237925030)) four minutes earlier. The rebase added only main's Feather-version fix in `export.py`, so the crash-path code was identical in both runs.
- **The base branch is green**, and every other Windows job on this head — 3.10, 3.11, 3.13, 3.14 — passed, as did the full suite locally (2049 passed).

There is no fix elsewhere to port. Re-running the failed job once to confirm; if it fails again, the failure is real and I will root-cause it rather than call it a flake.

If it does recur, my first suspect is the test's own timing rather than the crash handler: `panel_text()` reads `capsys` once, immediately after `run_test()` re-raises, while the panel is written when Textual prints its exit renderables during shutdown. Accumulating `capsys.readouterr().err` until it holds the panel or a short deadline passes would close that window. That is a change to `test_crash.py`, which this PR otherwise leaves alone, so I would rather propose it than widen the diff on my own.

---

src/harlequin/catalog_cache.py
```diff
@@ -59,36 +62,81 @@ def update_catalog_cache(
...
+    how many records moved.
+    """
+    if not connection_hash:
+        return 0
+    cache_file = _get_cache_file(HISTORY_CACHE_VERSION)
+    if not cache_file.exists():
+        return 0
+    try:
+        if recent(connection=connection_hash, limit=1):
+            return 0
```
tconbeer: **One `hsql` run before the first upgraded `harlequin` run discards the whole pickled history, permanently.**

"The store holds a row for this connection" is not the same question as "the move already happened". `hsql` writes the store under an identical key — `harlequin/cli.py:706` and `hsql/cli.py:1498` both call `connection_id(...)` with the same arguments — and `hsql` never migrates (`migrate_pickled_history` is referenced only from `app.py:1386`). So one `hsql -e ...` against a database, any time between the upgrade and the next time that database is opened in the IDE, makes this guard truthy forever.

Reproduced on this branch — a 3-record v2 pickle, then a single `hsql` write, then the migration:

```
migrated: 0
F8 now shows: ['select 1']
pickle still on disk: True
```

Nothing is destroyed on disk, but the user's pre-upgrade history never reaches F8 again and nothing says why. Given that the two commits immediately before this one are what taught `hsql` to write the store, "an agent ran something first" is the expected ordering here, not the exotic one.

Suggest keying the decision on a durable per-connection marker (a sentinel row, or a `migrated` table) rather than on whether *any* writer got there first.

---

src/harlequin/catalog_cache.py
```diff
@@ -59,36 +62,81 @@ def update_catalog_cache(
...
+                rows=None if failed else record.result_row_count,
+                elapsed_ms=record.elapsed * 1000,
+                # recorded in local time, and the store keeps UTC
+                run_at=record.executed_at.astimezone(timezone.utc),
+            )
+            if written is not None:
+                migrated += 1
+    finally:
+        log.close()
+    return migrated
```
tconbeer: **Two instances starting at once each migrate the whole pickle, and a part-done migration is never finished.**

The guard above runs on its own sqlite connection that `recent()` opens and closes; each `log.write()` here is then its own committed transaction (`QueryLog._run` commits per insert). Nothing spans the two, so:

*Duplication.* Three processes against one 200-record pickle, run on this branch:

```
each process migrated: ['200', '200', '200']
rows in store: 600 (pickle held 200)
```

Two panes restoring `harlequin f1.db` at login is enough — the write window is ~34ms for 500 records (measured), and the duplicates are permanent and count against `RETENTION_ROWS`.

*Partial migration.* Kill the process, or have `log.write()` hit a store error part-way (`_run` catches `sqlite3.Error`, disables the log, and returns `None` — the loop keeps calling a dead log), and the rows that landed make the next start's guard return 0. The remainder never moves, even though the pickle is still there.

*Silence.* `log.failure` is dropped at line 121. `_report_query_log_failure()` reads `self.query_log.failure`, a different object, so a migration that half-failed tells the user nothing.

One `begin immediate` around the guard and the inserts fixes all three — and makes "complete" a meaningful enough state to then `chmod 0600` or unlink the v2 pickle, which is worth doing: it holds every historical statement *verbatim* (the v2 writer never redacted), at 0644 in a 0755 dir, while this PR is careful to put the same text into a 0600 store in a 0700 dir.

---

src/harlequin/catalog_cache.py
```diff
@@ -59,36 +62,81 @@ def update_catalog_cache(
...
+        return 0
+    cache_file = _get_cache_file(HISTORY_CACHE_VERSION)
+    if not cache_file.exists():
+        return 0
+    try:
+        if recent(connection=connection_hash, limit=1):
+            return 0
+    except sqlite3.Error:
+        return 0
+    cache = _load_cache(cache_file)
```
tconbeer, line 80: `_load_cache`'s `except` tuple (line ~139) catches `UnpicklingError, ValueError, IndexError, FileNotFoundError, AssertionError, EOFError` — not `AttributeError` or `ModuleNotFoundError`. That was survivable when it only read a file this same version wrote; reading a *version-2* file makes `harlequin.history.History` and `harlequin.history.QueryExecution` an on-disk format contract, because a real v2 pickle names them by module path.

Simulating a future rename on this branch:

```
RAISED AttributeError: Can't get attribute 'History' on <module 'harlequin.history'>
```

That escapes the worker. `exit_on_error=False` means the app survives, but the user gets "Harlequin could not load its cache." on every launch and their history never migrates.

Worth guarding here, and worth a note on `History`/`QueryExecution` that they are now load-bearing for an old pickle — because `History` reads as a deletable one-field wrapper now that `append`/`blank` are gone, and deleting it is exactly the break.

---

src/harlequin/app.py
```diff
@@ -1408,6 +1380,22 @@ def _load_catalog_cache(self) -> None:
         cache = get_catalog_cache()
         if cache is not None:
             self.post_message(CatalogCacheLoaded(cache=cache))
+        if self.query_log.enabled:
+            # before this session runs anything: the move only happens while
+            # the store holds nothing for this connection
+            migrate_pickled_history(self.connection_hash)
```
tconbeer: The migration shares nothing with catalog-cache loading except the file it reads, and hanging it here costs three things:

- A migration failure reports as **"Harlequin could not load its cache."** (line 251), while the accurate new string — "Harlequin could not read your query history." — is attached to the *other* worker.
- The v2 pickle is never deleted, so every launch forever re-stats it and runs `recent(connection=…, limit=1)`, an sqlite open plus indexed read, before finding nothing to do.
- `tests/functional_tests/conftest.py:45` (`no_use_catalog_cache`) patches `get_catalog_cache` and `update_catalog_cache` but not `migrate_pickled_history`, so the migration now runs in every functional test and the `use_cache` marker no longer means "this test touches the cache path."

`_load_query_history` just below is the cohesive seam: migrate-then-read, under the notification that already says the right thing, paying per-F8 instead of per-launch.

---

tests/conftest.py
```diff
@@ -103,6 +109,48 @@ def query_log_path(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> Path:
...
+        # the field the class no longer declares, which is what makes this a v2
+        cache.history = {  # type: ignore[attr-defined]
+            connection_hash: History(queries=queries)
+        }
+        catalog_cache_dir.mkdir(parents=True, exist_ok=True)
+        cache_file = catalog_cache_dir / f"catalog-cache-{HISTORY_CACHE_VERSION}.pickle"
+        cache_file.write_bytes(pickle.dumps(cache))
+
+    return _write
+
```
tconbeer, line 149: This builds the "legacy" cache out of the **live** `History`/`QueryExecution` classes, so the migration tests round-trip today's classes against today's classes. That is precisely the constraint they exist to protect and the one thing they cannot see break: rename or move either class and real v2 pickles stop loading (see the `_load_cache` comment), while `test_a_pickled_history_moves_into_the_store` keeps passing.

A single committed version-2 pickle under `tests/data/` (the `data_dir` fixture is already there), copied into `catalog_cache_dir`, pins the actual on-disk shape an old Harlequin wrote — and lets this factory, plus the `pickle`/`deque`/`History`/`QueryExecution` imports it drags into the root conftest, come out entirely.

Separately, `del record.__dict__["status"]` at line 140 changes nothing the test exercises — the migration reads only `query_text`, `executed_at`, `result_row_count` and `elapsed` — and it silently stops modelling v2 the day `status` becomes a `field(default_factory=…)`.

---

src/harlequin/history.py
```diff
@@ -50,24 +78,67 @@ def __rich__(self) -> RenderableType:
...
+    except ValueError:
+        return None
+    elapsed_ms = values["elapsed_ms"]
+    return QueryExecution(
+        query_text=str(values["sql"]),
+        executed_at=executed_at,
+        result_row_count=values["rows"],
+        elapsed=None if elapsed_ms is None else float(elapsed_ms) / 1000,
+        status=values["status"],
+    )
```
tconbeer, line 155: `_record` owns store-row → `QueryExecution` (UTC→local, `elapsed_ms`/1000, `status`), and `migrate_pickled_history` owns the exact inverse (local→UTC, `elapsed`*1000, negative-row-count→`"error"`). The two halves of one mapping sit in different modules and neither mentions the other, so a change to either has to be mirrored across a file boundary — and `catalog_cache.py` picks up `sqlite3` + `QueryLog` + `recent` to hold its half, which makes the pickle-cache module a writer of another subsystem's store.

`tests/unit_tests/test_history.py` already says this out loud: it imports `migrate_pickled_history` from `harlequin.catalog_cache` while everything else in the file comes from `harlequin.history`.

Shape that keeps both halves together: leave a narrow reader in `catalog_cache.py` (`load_legacy_history(connection_hash) -> Iterable[QueryExecution] | None`, using its own `_get_cache_file`/`_load_cache`) and move the writing half here, next to `_record`. This module already imports `query_log`; `catalog_cache.py` then needs neither `sqlite3` nor `QueryLog`.

---

src/harlequin/history.py
```diff
@@ -50,24 +78,67 @@ def __rich__(self) -> RenderableType:
...
     @classmethod
-    def blank(cls) -> "History":
-        return cls(queries=deque([], maxlen=500))
+    def recent(
+        cls,
+        *,
+        connection: str | None = None,
+        n: int | None = DEFAULT_ROWS,
+        search: str | None = None,
+        path: Path | None = None,
```
tconbeer: Two small ones on this signature:

- `n` is a single-letter name on a public keyword-only parameter that crosses a function boundary and gets spelled by callers (`History.recent(n=2)`), and it forwards to the same concept spelled `limit` in `query_log.recent`. AGENTS.md's naming rule points at `limit` (or `n_rows`) here.
- `path` has no caller anywhere — production or test — and the same knob is already reachable through `query_log.recent(path=…)`. Worth dropping. (`search` is uncalled in production too, but the plan doc's screen filter lands next, so that one reads as a deliberate lead.)

---

src/harlequin/history.py
```diff
@@ -1,36 +1,64 @@
+"""The queries the History screen shows, read from the query log.
+
+One indexed read of `harlequin.query_log` per screen, so a human scrolling
+their history sees what `hsql` ran against the same database as well as what
+they ran themselves, and a session that never quit cleanly keeps both.
+"""
```
tconbeer: Collecting the AGENTS.md comment-rule violations in one place, since they share a cause — several comments narrate this PR's own refactor, which reads to the next person as a live distinction once the diff is gone.

**"Never reference code that was removed or refactored":**

- Here, lines 3–5 — "sees what `hsql` ran … and a session that never quit cleanly keeps both" is the commit message, selling the change by implicit contrast with the pickle-on-quit path. `"""The queries the History screen shows, read from `harlequin.query_log`."""` is the whole of it.
- `catalog_cache.py:99` — "the field this class no longer declares". The load-bearing half survives as `# a version-2 pickle carries a history dict keyed by connection hash`.
- `tests/conftest.py:143` — same phrase, plus "which is what makes this a v2".
- `catalog_cache.py:26` — "**The last** version that held the query history, **which the store now holds**". Naming the version whose pickle holds a history is legitimate; both of those clauses are only meaningful relative to this change.
- `test_history_screen.py:348` — `"""A history that is read when it is asked for is never 'not yet loaded'."""` quotes the title of the modal this PR deleted.

**"Never record a decision or a tradeoff":**

- `history.py:108-109` — "a missed screen beats a worker stalled behind someone else's write" is the "we chose A over B" shape verbatim.
- `history.py:127-128` — "should cost that row rather than the screen". The first line already says it returns None; and "a row nothing can read" is vaguer than the real condition (an unparseable `run_at`).
- `catalog_cache.py:81-84` — three clauses arguing which record is authoritative, over the "at most a sentence or two on why" budget.

Legitimate and worth keeping: `catalog_cache.py:107-108` (a v2 record's negative row count) and the field list in `tests/conftest.py:127-129` — those describe an on-disk format the code has to decode.

The CHANGELOG entry is compliant as written.

---

tests/functional_tests/test_history_screen.py
```diff
@@ -273,5 +283,91 @@ async def test_a_session_that_records_nothing_still_runs_queries(
...
+
+async def open_history(
+    pilot: Pilot,
+    app: Harlequin,
+    wait_for_workers: Callable[[Harlequin], Awaitable[None]],
+) -> HistoryScreen:
+    """Press the key that opens the History screen, and return the screen."""
+    await pilot.press("f8")
+    await wait_for_workers(app)
+    await pilot.pause()
```
tconbeer: This helper is right — F8 is asynchronous now, so the screen only exists once `_load_query_history` posts `QueryHistoryLoaded`. But the two pre-existing tests above were not brought along, and both are weaker for it:

**`test_history_screen`, lines 70-71** presses `f8` then `down` with no `wait_for_workers` between them. With `harlequin.history.recent` slowed to 200ms, `app.screen` is still `ScreenBase` right after `press("down")` — the `down` goes to the main screen — and when the History screen does arrive `highlighted == 0`, the newest row. `enter` then inserts `sel;`, not `select 14;`, which is what the snapshot on line 75 is named for. It passes on timing today; on a loaded runner that is a snapshot failure, or a silently wrong baseline if someone regenerates under load.

**`test_history_screen_crash`, lines 82-100** no longer pins [#485](https://github.com/tconbeer/harlequin/issues/485). All three `f8` presses now land before any screen exists, so the "push a history screen while one is already open" path is never taken — and because `_load_query_history` is `exclusive=True`, presses 2 and 3 just cancel the previous worker. The test asserts nothing at all, so it passes even if no history screen is ever mounted. The guard that replaced it lives at `app.py:1137-1138` and is untested.

Both want `await open_history(...)` (or a `wait_for_workers` between the presses), and the crash test wants an assertion.

---

src/harlequin/history.py
```diff
@@ -50,24 +78,67 @@ def __rich__(self) -> RenderableType:
...
+        # they ran the query in
+        executed_at = datetime.fromisoformat(str(values["run_at"])).astimezone()
+    except ValueError:
+        return None
+    elapsed_ms = values["elapsed_ms"]
+    return QueryExecution(
+        query_text=str(values["sql"]),
+        executed_at=executed_at,
+        result_row_count=values["rows"],
+        elapsed=None if elapsed_ms is None else float(elapsed_ms) / 1000,
```
tconbeer: The `try` above is narrower than the docstring promises: only `fromisoformat` is inside it, so `float(elapsed_ms)` here is unguarded. Set `elapsed_ms = 'abc'` on one row out of band and `History.recent()` raises `ValueError: could not convert string to float`, which escapes `_load_query_history` — a warning toast and **no screen at all**, rather than the one row being dropped as documented.

Low, since `_rounded()` guarantees float-or-None on both writers' paths, but it is the same "a file on the user's disk shouldn't cost the screen" argument the docstring already makes. `.astimezone()` is outside the `try` too, and can raise `OverflowError`/`OSError` on an extreme timestamp.

---

tconbeer (commented):
## Scope of this review

Four narrow passes over `abdd3fc`: **correctness/logic**, **reuse & architectural fit**, **security & data integrity**, and **comments/docstrings/changelog against AGENTS.md**. Verified locally at `abdd3fc`: `mypy` clean (152 files), `ruff` clean, `lint-imports` 5/5 kept, `pytest tests/unit_tests` 1840 passed, the two history test files 27 passed, no snapshot churn.

The design is right, and the capture set is complete — I diffed the four removed `append_to_history` call sites against the `query_log` writes in `_execute_query`/`_fetch_data` and every one is covered (errors via `app.py:1432`, DDL via `:1447`, results via the `update` at `:1577`, fetch errors via `:1584`). Dropping `reversed(...)` in the screen is correct, and `QueryExecution.render` genuinely single-sources the row rendering.

**The migration is where the risk is.** Three findings there, two reproduced on this branch:

1. **The guard mis-fires when `hsql` wrote first, and the user's entire pickled history becomes unreachable.** Both commands key the store identically, only the IDE migrates, so one `hsql -e '...'` between the upgrade and the next IDE launch is enough. Given that the two commits before this one are what taught `hsql` to write the store, this is the expected ordering, not the exotic one.
2. **Concurrent starts duplicate the history.** Three processes against a 200-record pickle left 600 rows.
3. **A part-way migration is never retried and never reported** — per-row commits, the guard then sees what landed, and `log.failure` is discarded.

All three come from one root cause — the guard and the inserts aren't one transaction, and "any row exists" is standing in for "the move already happened" — and one fix addresses all of them: a durable per-connection marker written inside the same `begin immediate` as the inserts.

Then a medium: `_load_cache`'s `except` tuple doesn't catch `AttributeError`/`ModuleNotFoundError`, and reading the v2 pickle makes `harlequin.history.History`/`QueryExecution` an on-disk format contract. Worth knowing because `History` now reads as a deletable one-field wrapper, and the test fixture builds the "legacy" pickle from the live classes, so the suite would not report that break.

Two test findings on pre-existing tests the async F8 changed out from under: `test_history_screen` races the worker, and `test_history_screen_crash` no longer pins #485.

The rest — the split record↔store mapping, the `_load_catalog_cache` seam, `n`/`path`, and the AGENTS.md comment-rule violations — are inline and non-blocking.

One behavior change worth a deliberate decision rather than a finding: DDL statements lose their duration on the screen (they log `rows=None, elapsed_ms=None`), and a result's duration is now its own fetch rather than wall-clock from submission. For DuckDB those agree — I measured execute at ~0ms against fetch at 49ms on a 30M-row aggregate, since execute is lazy — but for an adapter whose `execute()` does the server-side work, the screen will under-report. `hsql` already behaves this way, so it's consistent; it's just newly visible in the IDE.

## Not checked

- **I did not run the full functional suite** on 3.12/3.14, Windows or macOS — only `tests/unit_tests` (1840) plus the two history files on Linux/3.10. CI is green on the jobs that have finished.
- **I did not test the migration against a genuine version-2 pickle** written by a released Harlequin. Every reproduction above synthesized one from the current classes — which is the same limitation I'm flagging in the test fixture, so treat "the migration reads real v2 files" as unverified by both of us.
- **I did not verify the `connection_id` keying end to end** against a real adapter connection — I read the two call sites and confirmed they pass the same arguments, but did not run `harlequin` and `hsql` against one database and compare the stored keys. The whole first finding rests on that equality.
- **I did not exercise the screen against a large store** (the 500-row read, or a store near `RETENTION_ROWS`), so the "one indexed read" performance claim is unverified at scale.
- **I did not review `query_log.py` itself**, which merged two commits ago — only the new call sites into it.
- **I did not check the S3 tree or editor-cache consequences** of the `CACHE_VERSION` 2→3 bump beyond confirming `databases` has been empty since #1087 introduced v2.

A clean review isn't the merge gate — this one isn't clean anyway. The migration findings are yours to weigh: findings 1 and 3 are silent data-unreachability, which is the kind that gets reported months later as "my history vanished."

---

src/harlequin/catalog_cache.py
tconbeer: Fixed in 772f1b9, keyed on a durable marker as you suggest. `query_log.adopt()` writes the rows and a row in a new `migrated_connections` table in one transaction, so "has this connection been adopted" is now its own recorded fact rather than an inference from whether anyone wrote first.

`test_a_connection_hsql_wrote_first_still_gets_its_pickle` is your reproduction as a test: an `hsql` row for the connection, then the move, then all three queries in the screen. `test_a_connection_that_has_run_since_is_still_adopted` pins the same thing one layer down.

---

src/harlequin/catalog_cache.py
tconbeer: All three fixed in 772f1b9 by the single `begin immediate` you describe. The rows and the marker go in together through `query_log.adopt()`, so:

- Duplication: `test_only_one_of_several_processes_adopts` runs your three-process case — one `claimed`, two `skipped`, 200 rows.
- Partial: `test_a_failed_adoption_leaves_nothing_behind` makes a write fail mid-batch, asserts the store is empty, then adopts successfully — the rollback is what makes the retry possible.
- Silence: the failure now reaches the user. `migrate_pickled_history` lets `sqlite3.Error`/`OSError` out, `_load_query_history` catches it onto `QueryHistoryLoaded.warning`, and the handler notifies — the screen still opens on the rows that are there.

And I took the permissions half: `restrict_legacy_cache()` chmods the v2 pickle 0600 once it has been adopted. I stopped short of unlinking it — the contents are in the store either way, but deleting a user's file is a harder step to take back, and the marker means it is never read again.

---

src/harlequin/catalog_cache.py
tconbeer: Both done in 772f1b9. `_load_cache` now also catches `AttributeError` and `ImportError` (which covers `ModuleNotFoundError`), and `History`/`QueryExecution` each carry a line saying a version-2 cache pickles them by module path, so renaming or moving either stops those files loading.

`test_a_cache_this_version_cannot_unpickle_is_not_an_error` is your simulated rename as a test — it rewrites `QueryExecution` inside the committed pickle's bytes and asserts the move is a no-op rather than a raise.

---

src/harlequin/app.py
tconbeer: Moved to `_load_query_history` in 772f1b9 — migrate, then read, under the notification that says the right thing. Start-up pays nothing, and the per-F8 cost is one set lookup after the first press (`harlequin.history._migrated`), so the pickle is read at most once per session even though the file stays on disk.

On the third bullet: with the migration off the launch path the `use_cache` marker means what it says again. I left `migrate_pickled_history` unpatched in `no_use_catalog_cache` on purpose — the new `catalog_cache_dir` fixture in the root conftest points `user_cache_dir` at tmp for every test, so nothing can reach a developer's real cache, and the history tests need the real function.

---

src/harlequin/history.py
tconbeer: Done in 772f1b9, in the shape you sketched. `catalog_cache.load_legacy_history()` is the narrow reader and imports neither `sqlite3` nor anything from `query_log` beyond the `get_connection_hash` re-export it already had; `_stored()` now sits directly under `_record()` here, and says it is the inverse.

`test_history.py` imports `migrate_pickled_history` from `harlequin.history` with everything else in the file now.

---

src/harlequin/history.py
tconbeer: All nine cut or rewritten in 772f1b9. The module docstring is now exactly the line you wrote; `catalog_cache.py`'s comment is the "carries a history dict keyed by connection hash" half; `HISTORY_CACHE_VERSION` is "The version whose pickle holds a query history."; the `recent()` and `_record()` comments say what the code does without arguing for it; the conftest phrasing went with the factory; and the screen test's docstring no longer quotes a modal that is gone.

The two you called legitimate are kept — the negative-row-count line moved into `_stored()` with the code it explains.

---

tests/functional_tests/test_history_screen.py
tconbeer: Both fixed in 772f1b9.

`test_history_screen` now goes through `open_history()` before pressing `down`, so the keypress cannot reach the main screen. The two snapshots are unchanged, which says the baseline was the right one — it was just being reached by luck.

The crash test is now `test_a_second_request_does_not_stack_history_screens` and actually exercises #485: open the screen, press `f8` again with it up, and — for the half a keypress cannot reach, since the action returns early — post a second `QueryHistoryLoaded` directly. It then asserts the app is alive, the screen is the same object, and there is exactly one `HistoryScreen` on the stack, which is what pins `app.py`'s handler guard.

---

tests/conftest.py
tconbeer: Done in 772f1b9. `tests/data/catalog-cache-2.pickle` is committed, written by an actual 2.14.0 wheel:

```
uv run --isolated --no-project --with 'harlequin==2.14.0' python scripts/write_legacy_cache.py
```

The generator is committed too, and refuses to run against a tree whose `CACHE_VERSION` is not 2 — which is how I caught `uv run` from inside the project handing it the working tree instead of the release.

The factory, `del record.__dict__["status"]`, and the `pickle`/`deque`/`History`/`QueryExecution` imports are all out of the root conftest; what remains is a fixture that copies the file and a `LEGACY_HISTORY` dict of what it holds, for the assertions. `test_a_cache_this_version_cannot_unpickle_is_not_an_error` then breaks a class name inside those committed bytes, which is the failure mode the old factory could not see. `tests/data/*.pickle` is marked `binary` in `.gitattributes` so a Windows checkout cannot rewrite it.

---

src/harlequin/history.py
tconbeer: Both done in 772f1b9 — `limit`, matching `query_log.recent`, and `path` is gone. `search` stays for the reason you give.

---

src/harlequin/history.py
tconbeer: Fixed in 772f1b9: the `try` now covers `fromisoformat`, `.astimezone()` and the `float()`, and catches `ValueError, TypeError, OverflowError, OSError`. `test_a_row_with_an_unreadable_timestamp_costs_only_that_row` asserts the documented behaviour, and the docstring no longer claims more than the code does.

---

tconbeer:
Thanks — the three migration findings were real, and the first one is the kind I'd have hated to ship. All ten threads are addressed in 772f1b9 and replied to individually; the short version:

**The migration is one transaction now.** `query_log.adopt()` writes the rows and a `migrated_connections` marker under one `begin immediate`, so "already adopted" is a recorded fact instead of an inference from whether any writer got there first. That is one fix for all three: `hsql`-wrote-first no longer discards the pickle, three processes produce 200 rows rather than 600, and a failure rolls back and is retried instead of leaving a stub that looks finished. A failure also now reaches the user, through `QueryHistoryLoaded.warning`. Your reproductions are tests: `test_a_connection_hsql_wrote_first_still_gets_its_pickle`, `test_only_one_of_several_processes_adopts`, `test_a_failed_adoption_leaves_nothing_behind`.

**And the rest:** the move runs with the read (`_load_query_history`), the record mapping is back in one module, `_load_cache` survives a rename, `limit` replaces `n`, `_record` guards the whole row, the comments are cut, and both pre-existing screen tests were brought along — the #485 one now asserts something.

**The test fixture is a real 2.14.0 pickle.** `tests/data/catalog-cache-2.pickle`, written by the released wheel via `scripts/write_legacy_cache.py`. That closes your "unverified by both of us" item: the migration is now exercised against bytes a released Harlequin produced, not against today's classes. The generator refuses to run against a tree whose `CACHE_VERSION` isn't 2, which is how I caught `uv run` inside the project handing it the working tree.

**The one thing I did not decide: DDL duration.** You're right that DDL now shows no duration, and that a result's duration is its own fetch rather than wall-clock from submission. I've left both as they are — it matches what `hsql` records, and the alternative means completing DDL rows from the message handler after the batch, which puts a store write on the UI thread. But it is a user-visible change to a screen people read, so if you'd rather the IDE keep reporting wall-clock I'll do it in this PR rather than leave it to be discovered.

Local: 2060 passed, 7 skipped; mypy clean; `lint-imports` 5/5. Same caveat as yours — Linux/3.10 only, CI has the rest.

---

PR #1161, opened 2026-09-15
M4 9/14: the History screen, reconsidered

### LINKED ISSUE #850: Re-think History Screen UI/UX
I think it would be nice to show all keybindings like in other screens.

Also I think History should be a modalScreen (it should not have the keybindings of underlying screen)

If this change is not wanted we can close the issue, but I have prepared a pr for this, so check it out :)

tconbeer:
Closes [#850](https://github.com/tconbeer/harlequin/issues/850), and finishes the IDE half of [#429](https://github.com/tconbeer/harlequin/issues/429) — PR 9 of M4. PR 8 taught the History screen to read the query log; this one lets you filter it and reworks the screen around that.

**Three filter controls, hidden until asked for.** `ctrl+f` shows a search box plus dropdowns for which command ran the query and how it went, and puts the cursor in the box; pressing it again puts them away and clears all three. Typing over the list opens them too, carrying the key you typed. All three narrow together, and the list's border subtitle carries the count.

```
╭─ Filter ─────────────────────────────────╮╭─ Program ──────────╮╭─ Outcome ──────────╮
│  line_items                              ││ hsql             ▼ ││ Any outcome      ▼ │
╰──────────────────────────────────────────╯╰────────────────────╯╰────────────────────╯
```

**The search box is a query over the store, not a narrowing of what is listed** — the same `order by id desc limit` `hsql --history-search` runs, on a debounced worker, so it reaches queries older than the 500 the screen holds. `query_log.recent()` gains `program` and `status` (one indexed equality each), so `hsql --history` can offer the same two whenever it wants them.

**Each pane owns its own keys.** The screen opens on the list, and `tab` cycles whatever is shown — list ↔ preview with the controls away, and through all five when they are out:

| | list | search box | dropdowns | preview |
| --- | --- | --- | --- | --- |
| arrows, pgup/pgdn | move the highlight | nothing | open / pick | scroll the preview |
| `enter` | take the query, and leave | take the term to the list | open, then pick | take the query, and leave |
| `esc` | leave | empty the term, then put the controls away | close the menu | leave |
| anything else | opens the controls, with the key you typed | typed | type-to-search | — |

**`HistoryScreen` is a `ModalScreen`**, so the app's bindings stop reaching through a screen whose job is to pick one query. `ctrl+q` still quits, because App *priority* bindings are dispatched over the full chain.

**A row of the list is ellipsised rather than wrapped**, so one line of a query is one line of the row. A 2,900-character one-liner was filling the pane by itself; `render()` truncated by logical lines and a single long one still wrapped to ~50.

**The preview is readable, not editable** — `read_only=True, show_cursor=False`, so it takes focus and scrolls both ways from the keyboard with no cursor and no cursor-line highlight. The screen's bindings are in a `Footer`, read from the live keymap, and the preview pane renders its title for the first time (the border was on the inner `TextArea` while `border_title` was set on the `TextEditor`).

The reasoning behind the feature is in [`docs/plans/m4-hsql-technical-plan.md`](docs/plans/m4-hsql-technical-plan.md) §1.2 and §3.4. What the interaction work settled, and the parts worth knowing before testing it:

- **Hiding the controls clears them.** A filter still narrowing the list from behind a hidden control is a list nobody can explain. The cost is that `ctrl+f` twice loses your search — the alternative was keeping the state and showing it somewhere, which is more chrome for a rarer case.
- **Typing over the list still opens them.** Otherwise "start typing to filter" would have quietly disappeared behind a hidden box.
- **The debounce coalesces rather than restarts**, matching `CodeEditor.schedule_symbol_scan`, so results come while you type. Two consequences review found: a read the next keystroke overtook still arrives, so `HistoryFiltered` carries the filters it started with and a stale one is dropped; and an edit that comes to nothing inside the window still reads, so the screen holds `applied_filters` and a read asking again for what is already listed changes nothing rather than dropping you back to the top of the list. A dropdown reads immediately — one choice is not a stream of them.
- **`enter` is a priority binding, and `check_action` declines it while a dropdown is open.** The filter box binds `enter` to its own submit and the preview stops it before a binding is looked for, so the screen has to own it; but an open menu needs it too, and a priority binding that declined by doing nothing would swallow the key. Declining in `check_action` lets it fall through to the widget with focus. `esc` needs none of this since textual-textarea 0.18.4.
- **The screen's CSS names `HistoryList`, not `OptionList`.** Textual builds a `Select`'s menu out of an `OptionList`, so the rules written for this screen's own list were landing on the dropdowns as well — and the `height: 1fr` that makes the list fill its pane gave the menus a content height of zero. They held their options and rendered as empty boxes. The guard asserts the height a menu is given rather than the options it holds, since it had those all along.
- **`ModalScreen` here is for binding precedence, not for the dim backdrop** every other modal in the repo uses — this screen fills the viewport. That is now a comment in the CSS rather than a silent divergence.
- **Two upstream bugs came out of this**, both fixed and pinned: [textual-textarea#346](https://github.com/tconbeer/textual-textarea/issues/346), where `read_only` did not prevent paste, cut, undo, toggle-comment or delete-line (macOS CI caught it as `ctrl+v` pasting the runner's clipboard into the preview), and [#345](https://github.com/tconbeer/textual-textarea/issues/345), where `TextEditor` did not forward `show_cursor`, swallowed `escape`, and left its file and search bindings live on a read-only editor. 0.18.4 carries both, which is why `QueryPreview` is now a docstring and one class kwarg.

- [ ] No.
- [x] Yes; I haven't opened a PR, but the gist of the change is: the History screen has filter controls behind `ctrl+f` — a search over your whole query history, plus program and outcome — and the screen's keys are now listed in its footer. This belongs in the release's docs PR (PR 10 of M4), alongside `hsql --history` and `--history-search`.

- [x] Yes.

Twenty-six new tests, asserted on behavior rather than on the snapshot, since the snapshot changes for reasons none of them cares about. Every cell of the table above is pinned, plus: the controls start hidden and putting them away stops filtering; each dropdown has room for its options, narrows on its own, and keeps `enter` and `escape` while open; all three filters narrow together; an app-level binding pushes no screen while `ctrl+q` still quits; the filter finds a query a 501-row store had pushed out of the listing; a typed `_` matches an underscore rather than `like`'s any-character; and an edit that comes to nothing keeps your place. `test_the_preview_ignores_every_key_it_binds` presses every key the text area binds, derived at run time, so a component-library bump that gives a read-only editor a way to edit itself fails and names the key.

I checked the ones worth checking are real guards rather than vacuous: reverting `ModalScreen` → `Screen` fails the binding test on `f1`, removing the ellipsis fails the row-height test, removing the unchanged-filters early return fails the highlight test, restoring the `OptionList` selectors fails the dropdown test with `assert 0 >= 3`, and on 0.18.3 the key sweep failed with all ten editing keys named.

Snapshots regenerated on 3.10, with a third added for the filtered state. Rebased on `main`. Full suite green (2100 passed, 7 skipped), plus the py12 run, mypy, ruff and `lint-imports`.

https://claude.ai/code/session_01MrjoHmowjacPGUWMfjCyu3


src/harlequin/components/history_screen.py
```diff
@@ -42,51 +49,163 @@ def visualize(self) -> object:
...
+                    title="Query History",
+                    severity="warning",
+                )
+            return
+        self._filter_failed = False
+        self.history = message.history
+        self.list.clear_options()
+        self.list.add_options(self._options())
+        self._show_count()
+        self._highlight_first()
```
tconbeer, line 288: **The rebuild is unconditional, so every accepted read throws away the user's place in the list.**

The debounce is a throttle with a trailing read: `schedule_filter` arms a one-shot timer on the first `Input.Changed`, and `_read_filtered` reads whatever the value is when it fires. So an edit that is a net no-op inside the window still produces a read whose `search` equals the current value — it passes the staleness check at :189 and lands here, where `clear_options()` / `add_options()` / `_highlight_first()` force `highlighted = 0`.

Reproduced with a pilot probe:

```
type "orders", wait for the read, press down
  → highlighted: 1, preview: 'select 3 from orders'
press "x", then backspace (same 200ms window)
  → value: 'orders' (unchanged)
  → highlighted: 0, preview: 'select 4 from orders'
```

The typo-and-correction case is the narrow one. The commoner one is any edit that doesn't change the result set — `order` → `orders` where every match already contains it — which also scrolls the user back to the top mid-hunt.

Two fix shapes: track the term `self.history` was read for and return early when `message.search` matches it, or capture `self.list.highlighted_option.value` before the rebuild and restore that option's index after.

---

src/harlequin/components/history_screen.py
```diff
@@ -42,51 +49,163 @@ def visualize(self) -> object:
...
+    )
+    def read_history(self, search: str) -> None:
+        """The whole store filtered by `search`, not the rows already on screen."""
+        try:
+            history: History | None = History.recent(
+                connection=self.connection, search=search or None
+            )
+        except sqlite3.Error:
+            history = None
+        self.post_message(self.HistoryFiltered(history=history, search=search))
```
tconbeer: **Only `sqlite3.Error` is caught, and the same failure that is a toast on the initial read is an error modal here.**

`History.recent` also touches the filesystem — `default_path()` and `store.exists()` — so an `OSError` escapes this `except`. When it does, no `HistoryFiltered` is posted at all: `_filter_failed` stays `False`, no toast fires, and the list keeps showing the previous term's results underneath the new filter text.

The app's generic worker handler does catch it, but as an **"Unexpected Error" modal**, because `read_history` isn't in `_PARTIAL_FAILURE_WORKER_NOTIFICATIONS` (`app.py:252-257`) — and that dict is documented as "loud by default: a worker added later is an error modal until someone decides its failures are benign." Meanwhile `_load_query_history` catches `(sqlite3.Error, OSError)` *and* is in the map, so the identical unreadable store is a warning toast on the way in and a modal on the first keystroke.

Either widen to `except (sqlite3.Error, OSError):`, or drop the try/except and add `read_history` to `_PARTIAL_FAILURE_WORKER_NOTIFICATIONS` — which would also let `exit_on_error=False` do the job it is currently inert for, and match `CodeEditor`'s convention for the same situation.

---

tests/functional_tests/test_history_screen.py
```diff
@@ -399,3 +422,154 @@ async def test_a_session_that_records_nothing_moves_no_pickle(
...
+    log = QueryLog(program="hsql", connection="foo")
+    log.write("select * from lineXitems")
+    log.close()
+
+    async with app.run_test() as pilot:
+        while app.editor is None:
+            await pilot.pause()
+        screen = await open_history(pilot, app, wait_for_workers)
+        await type_filter(pilot, app, "line_items", wait_for_workers)
+        assert len(screen.history) == 0
```
tconbeer, line 788: **This one passes vacuously — it has no positive control.**

The whole test rests on an unstated coupling: `QueryLog(connection="foo")` has to match the `app` fixture's `connection_hash="foo"` (`tests/conftest.py:236`) for the record to be visible at all. I changed the connection here to `"not-this-app"` — the row is then invisible to the screen regardless of `like` escaping, and `assert len(screen.history) == 0` **still passed**.

So the assertion can't distinguish "the escape worked" from "the record was never there". One line fixes it:

```python
screen = await open_history(pilot, app, wait_for_workers)
assert len(screen.history) == 1
await type_filter(pilot, app, "line_items", wait_for_workers)
assert len(screen.history) == 0
```

(The escaping itself is genuinely correct — dropping `"_"` from `query_log._like_literal` does fail this test as written. It's only the premise that's unpinned.)

---

tests/functional_tests/test_history_screen.py
```diff
@@ -399,3 +422,154 @@ async def test_a_session_that_records_nothing_moves_no_pickle(
...
+    async with app.run_test() as pilot:
+        while app.editor is None:
+            await pilot.pause()
+        screen = await open_history(pilot, app, wait_for_workers)
+        assert not app.full_screen
+
+        for key in ("f1", "f6", "f10", "f12", "ctrl+r"):
+            await pilot.press(key)
+            await pilot.pause()
+            assert app.screen is screen, f"{key} pushed a screen"
```
tconbeer: **Two of these five keys assert nothing.**

`f6` (`focus_data_catalog`) and `ctrl+r` (`refresh_catalog`) push no screen even when they *do* reach the app, so `assert app.screen is screen` holds either way. I reverted `ModalScreen` → `Screen` with the loop reduced to `("f6", "ctrl+r")` and the test **passed**.

What actually pins the change is `f1`/`f12` (they push screens) and the trailing `assert not app.full_screen` for `f10`. Either give the other two an observable assertion — `screen.focused` unchanged after `f6` would do it — or drop them, so the loop doesn't read as five keys' worth of coverage.

Worth knowing while you're here: `ctrl+q` still quits from the screen, because App *priority* bindings are dispatched before `_modal_binding_chain` is consulted. That's the right outcome, and it's currently neither tested nor stated.

---

tests/functional_tests/test_history_screen.py
```diff
@@ -54,6 +56,19 @@ async def open_history(
...
+async def type_filter(
+    pilot: Pilot,
+    app: Harlequin,
+    term: str,
+    wait_for_workers: Callable[[Harlequin], Awaitable[None]],
+) -> None:
+    """Type into the filter and wait for the read it debounces to land."""
+    await pilot.press(*term)
+    await pilot.pause(FILTER_INTERVAL * 2)
+    await wait_for_workers(app)
```
tconbeer: **The wait is a timed hope, not a deterministic one.**

`pilot.pause(delay)` is a plain `asyncio.sleep`, so the whole margin is the 200ms between the debounce timer firing and `wait_for_workers` being called. If the timer hasn't fired yet, `wait_for_workers` sees no workers, returns immediately, and the caller asserts against stale history.

The failure mode is a false red rather than a false green — every caller asserts on post-filter content — so this is flake risk, not a coverage hole. But it's the thing here most likely to go intermittent on a loaded Windows runner, which is exactly where the suite already needs `--force-flaky`. A bounded poll on `screen.history` (or on a term-applied marker, if you add one for the highlight fix) removes the guess entirely.

The same inline sequence is repeated at :104-108 and :537-540 — worth routing those through this helper either way.

---

src/harlequin/components/history_screen.py
```diff
@@ -42,51 +49,163 @@ def visualize(self) -> object:
...
         "history-screen--error-label",
     }
 
+    BINDINGS = [
+        Binding("up", "cursor_up", "Up", show=False),
+        Binding("down", "cursor_down", "Down", show=False),
+        Binding("pageup", "page_up", "Page Up", show=False),
+        Binding("pagedown", "page_down", "Page Down", show=False),
+    ]
+    """The list's keys, so it is still driveable while the filter has focus."""
```
tconbeer: **These four keys bypass the keymap registry — they can't be rebound and don't show in `harlequin --keys`.**

This is the first class-level `BINDINGS` anywhere in `src/harlequin/components/`; every other navigation key in the app is a `HARLEQUIN_ACTIONS` entry bound through the vscode keymap — `data_catalog.cursor_up/down` (`actions.py:229-231`), `code_editor.cursor_up/down` (`actions.py:125-128`). AGENTS.md is explicit: *"A new user-bindable behavior needs an entry in `HARLEQUIN_ACTIONS` — otherwise it can't be bound or shown in the help screen."*

Suggest `history_screen.cursor_up` / `cursor_down` / `page_up` / `page_down` with `target=HistoryScreen`, plus entries in `VSCODE_HISTORY_SCREEN_BINDINGS`. The forwarding methods below stay as they are — binding on `HistoryList` wouldn't work, since the focused `Input` is its sibling rather than its ancestor, so the keys never bubble through it.

While you're in here: `pageup`/`pagedown` are never pressed by any test, and `up`/`down` through the filter — the "still driveable" claim this docstring makes — are pinned only by the snapshot. Removing both bindings fails `test_history_screen` and nothing else. One `assert screen.list.highlighted == 1` after a `down` would pin the claim behaviorally.

---

src/harlequin/components/history_screen.py
```diff
@@ -42,51 +49,163 @@ def visualize(self) -> object:
...
-class HistoryScreen(Screen[str]):
+class QueryPreview(TextEditor, can_focus=False):
+    """The highlighted query, for reading.
+
+    Nothing in it can take focus, so a cursor never lands in a query that the
+    pane will not let the user change.
+    """
+
+    def on_mount(self) -> None:
+        assert self.text_input is not None
```
tconbeer: **This is the "fix it upstream" case AGENTS.md names by title.**

`textual-textarea` is on the owned list, and this is a subclass that exists only to reach into a child's `can_focus` after mount because `TextEditor.__init__` has no way to say "not focusable" — alongside `read_only` and `use_system_clipboard`, which is exactly where it would belong. A `focusable: bool = True` argument there makes this one line here.

Both halves are load-bearing, for what it's worth, so this isn't belt-and-braces: `can_focus_children=False` is inherited from `TextEditor`, and `Screen.focus_chain` only descends when `allow_focus_children()`, so the keyboard side is covered by the class kwarg alone (measured chain on this screen is exactly `['Input', 'HistoryList']`). But `TextEditor.on_click` calls `self.text_input.focus()` unconditionally and `Screen.set_focus` gates on `widget.focusable` rather than on ancestors — so `text_input.can_focus = False` is what stops the *mouse*. That's two upstream facts this class is pinned to, which is the argument for putting the fix where they live.

If the release-and-pin-bump cost isn't worth it in this PR, that's fair — but per AGENTS.md, write it up as an issue there rather than letting the reach-into-a-child become the permanent shape.

---

src/harlequin/app.tcss
```diff
@@ -577,8 +577,30 @@ ExportScreen Button:hover,
...
+    height: 3;
+    width: 100%;
+}
+
+HistoryScreen QueryPreview {
+    border: round $border-color-nofocus;
+    border-title-color: $border-title-color-nofocus;
     width: 50%;
     height: 100%;
 }
```
tconbeer, line 612: Four CSS notes on this block, none load-bearing:

**`#history_panes { height: 1fr }` (584-586) is a no-op.** `Horizontal`'s `DEFAULT_CSS` is already `width: 1fr; height: 1fr` (`textual/containers.py:162-169`), and nothing in `app.tcss` or `global.tcss` overrides it for this subtree. Same for `height: 100%` on `#history_filter_pane` — `Vertical` defaults to `height: 1fr`.

**`width: 50%` twice should be `1fr` twice** (588-591 and 601-606). Percentages floor independently, so an odd terminal width leaves a dead column — 121 cols gives 60 + 60. The repo's own pane split already does this: `DataCatalog { width: 1fr }` against `#main_panel { width: 3fr }`. The `50%` predates this PR, but both rules are being rewritten anyway.

**The `border` / `border-title-color` / `background` trio repeats across all three blocks**, and that trio is already a grouped comma selector at `app.tcss:8-16` — whose `:focus` counterpart at 22-28 this PR just joined. Adding these three selectors there and leaving only sizing here keeps the nofocus and focus halves symmetric, which is what they were before.

**`HistoryScreen { background: $background }` (580-582) makes this the only modal in the repo that repaints the backdrop.** `ModalScreen.DEFAULT_CSS` is `background: $background 60%` — dim the app underneath — and `TextModal`, `HelpScreen`, `ConfirmModal`, `DebugInfoScreen` and `ExportScreen` all keep that dim and put `background: $background` on an inset container (`#modal_outer`, `#export_outer`, `ConfirmModal Vertical#outer`). This one fills the viewport opaquely instead, which is what the pre-PR plain `Screen` gave it. Defensible if `ModalScreen` was chosen purely for binding precedence and not for appearance — but as it stands the divergence is silent, and it'd be worth one line in the class saying so.

---

src/harlequin/actions.py
```diff
@@ -334,9 +334,17 @@ class Action:
...
+        # the filter input binds enter to its own submit
+        priority=True,
     ),
     "history_screen.cancel": Action(
-        target=HistoryScreen, action="cancel", description="Cancel"
+        target=HistoryScreen,
+        action="cancel",
+        description="Clear Filter / Cancel",
+        show=True,
     ),
```
tconbeer, line 356: `"Clear Filter / Cancel"` is the only description in `HARLEQUIN_ACTIONS` with a slash, and the only one naming two behaviors. Every other is a 1–3 word Title Case noun phrase: "Select Query", "Run Query", "Format Query", "Refresh Data Catalog", "Launch External Editor", "Previous Tab", "View Cell".

Width isn't the problem — the footer renders at ~43 of 120 cells. The inconsistency is. You already flagged this in the PR description as "a long label for one key"; I'd take the single noun phrase, or set the description dynamically the way `_show_count` already updates the list's border subtitle, rather than baking an either/or into a static string.

---

src/harlequin/components/history_screen.py
```diff
@@ -42,51 +49,163 @@ def visualize(self) -> object:
...
+        self.read_history(self.filter_input.value)
+
+    @work(
+        thread=True,
+        exclusive=True,
+        exit_on_error=False,
+        group="history_filters",
+    )
+    def read_history(self, search: str) -> None:
+        """The whole store filtered by `search`, not the rows already on screen."""
```
tconbeer: Nit, per AGENTS.md's "explain what the code does and why — nothing else": the `, not the rows already on screen` half defines the function by contrast with a design that isn't in the tree. A reader lands here and has to work out which alternative is being ruled out.

```suggestion
        """The whole store, filtered by `search`."""
```

Same shape at :153 — `# OptionList's two paging actions are the only unannotated ones`. The justification for the `type: ignore` is right to have; the "the only ones" claim is about upstream, isn't checkable from here, and goes stale on a Textual bump. `# OptionList's paging actions are unannotated` says everything the ignore needs.

---

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ All notable changes to this project will be documented in this file.
 ### Features
 
 - Harlequin's Query History (`F8`) now shows the queries hsql ran against the same database, and keeps them when a session ends without a clean quit ([#429](https://github.com/tconbeer/harlequin/issues/429)).
+- Harlequin's Query History (`F8`) now has a filter: type to narrow the list to queries that mention what you typed, anywhere in your history rather than only in the queries on screen ([#429](https://github.com/tconbeer/harlequin/issues/429)).
+- Harlequin's Query History screen now lists its keys in a footer, and its preview pane no longer takes a cursor ([#850](https://github.com/tconbeer/harlequin/issues/850)).
```
tconbeer: **The #850 entry omits what #850 actually asked for.** The issue is two sentences: *"I think it would be nice to show all keybindings like in other screens. Also I think History should be a modalScreen (it should not have the keybindings of underlying screen)."* The footer is there; the modal behavior — the app's keys no longer acting while History is open, which is the most user-visible change in this PR — isn't mentioned anywhere in the changelog.

```suggestion
- Harlequin's Query History (`F8`) now has a filter: type to narrow the list to queries that mention what you typed, searching your whole history ([#429](https://github.com/tconbeer/harlequin/issues/429)).
- Harlequin's Query History screen now lists its keys in a footer, ignores the app's keys while it is open, and its preview pane no longer takes a cursor ([#850](https://github.com/tconbeer/harlequin/issues/850)).
```

The #429 line is the same edit for a different reason: "rather than only in the queries on screen" is a distinction only someone who has read the diff has a name for, and the rule is to cut every word that doesn't help a user decide whether they care.

Unrelated to the entry itself — [#851](https://github.com/tconbeer/harlequin/pull/851) is a closed community PR against this same issue, and GitHub lists both as closing it. Your call whether that's worth a nod in the release notes.

---

tconbeer (commented):
Reviewed through four lenses, dispatched as separate passes: correctness/logic, reuse & simplification, test quality, and comment/changelog conventions. Everything below was reproduced with a probe or mutation-tested against the tree — nothing here is inferred from reading alone.

**Baseline I verified independently** on `ba099e4` in a clean worktree: 2066 passed / 7 skipped, 149 snapshots, `mypy` clean over 152 files, `ruff format`/`check` clean, `lint-imports` 5/5 contracts kept. CI on the head is green across all 32 jobs. The PR's claims hold. It is `behind` main and needs a merge before it can land.

### Worth fixing before merge

1. **The highlight resets on every accepted read** (`history_screen.py:203-207`) — including reads whose results are identical, so a user who has arrowed down to a query loses their place on the next keystroke. Reproduced.
2. **`read_history` catches only `sqlite3.Error`** (`:178-184`); an `OSError` from the same call escapes and surfaces as an "Unexpected Error" modal, while the identical failure on the initial load is a warning toast.
3. **`test_the_filter_matches_no_wildcards` passes vacuously** — no positive control; it still passes when the record is made invisible to the screen.
4. **Two of the five keys in the app-binding test assert nothing** — `f6` and `ctrl+r` pass against the pre-PR `Screen` too.

Everything else is judgment: the four navigation keys bypassing `HARLEQUIN_ACTIONS`, `QueryPreview` as an upstream fix, the CSS nits, the slashed action description, the changelog's #850 entry, and the debounce's flake margin in `type_filter`.

### Checked, no finding

Worth recording, since these are the things the design notes rest on and they all hold:

- **No keystroke can be lost, and no timer leaks.** Every change either arms the timer or finds one armed, and `_read_filtered` always reads the current value, so the last edit is always covered. `MessagePump._close_messages` stops timers on screen removal and `post_message` returns `False` on a closed pump, so a worker finishing after dismissal is a no-op.
- **Out-of-order reads converge.** Probed with an artificially slowed read forcing the interleave `start:a, start:ab, end:ab, start:a, end:a` — final state correct. `action_cancel`'s `value = ""` goes through `watch_value` → `Changed`, so clearing schedules a read too.
- **`ctrl+q` still quits from the modal.** App *priority* bindings are dispatched before `_modal_binding_chain` is consulted, so `quit` survives the `Screen` → `ModalScreen` switch while `f1`/`f10`/`f12` correctly don't.
- **No stale preview.** `clear_options()` sets `highlighted = None` first, so `_highlight_first`'s `= 0` is always a reactive change and `OptionHighlighted` always fires.
- **No binding accumulation** across re-opens — `bind()` mutates an instance-level `_bindings` copy.
- **`priority=True` doesn't leak.** The priority pass iterates the current screen's chain only, so `enter` on the catalog and results viewer is untouched. And it is genuinely required: `Input` binds `enter → submit`. It doesn't strip the screen's `up`/`down`/`pageup`/`pagedown` either — `Input.check_consume_key` only claims printable characters.
- **`dismiss` vs `pop_screen`** both behave; `pop_screen` pops the result callback without invoking it, unchanged from before.
- **Both `QueryPreview.on_mount` and `TextEditor.on_mount` run** (no shadowing — Textual dispatches per class across the MRO), and `Compose` precedes `Mount`, so the `assert self.text_input is not None` holds.
- **`_like_literal` escaping works** — `line_items` doesn't match `lineXitems`, `beta_` doesn't match `alpha_beta`.

### Not checked

- PR 8's `query_log.py` / `history.py` internals, except where this PR calls into them (`recent`, `_like_literal`) — outside this diff.
- The py12-only snapshot pass and the `online` tests; relied on CI for both.
- The two regenerated SVGs pixel-for-pixel. I read their extracted text and confirmed the footer, both border titles, the filter value and the match count render, and that `[New buffer with select 14]` is byte-unchanged — which independently confirms the clear-filter → `down` → `enter` path still lands on the same query.
- Performance of the filter against a large store. Note the `LIKE '%…%'` predicate can't use an index despite `query_log.recent`'s "one indexed read" docstring, and `exclusive=True` can't stop a thread mid-`execute` — a probe observed three overlapping reads from fast typing. Correctness is unaffected by both; it's throughput only.

A clean review is necessary, not sufficient — the call on whether the screen's intent is preserved is yours.

---

src/harlequin/components/history_screen.py
tconbeer: Fixed in cbedf3e. Real bug, and your repro is exactly it.

Took the first shape: the screen now holds `search`, the term `history` was read for, and `show_filtered_history` returns early when an accepted read carries that same term. A genuine change still resets to the top, which is what a new result set should do.

Pinned by `test_an_edit_that_changes_nothing_keeps_your_place`, which watches `read_history` rather than waiting on the term settling — waiting on that would be waiting on what is already true, and the first version of the test passed without the fix for exactly that reason.

---

src/harlequin/components/history_screen.py
tconbeer: Fixed in cbedf3e — widened to `except (sqlite3.Error, OSError)`, matching `_load_query_history`.

Took the narrower of your two options deliberately: the same unreadable store should report the same way on the way in and on the first keystroke, and `_load_query_history` already catches both and warns. Dropping the try/except and adding `read_history` to `_PARTIAL_FAILURE_WORKER_NOTIFICATIONS` would get there too, but it would put the message for one failure in two places and lose the once-per-streak guard.

---

tests/functional_tests/test_history_screen.py
tconbeer: Fixed in cbedf3e — `assert len(screen.history) == 1, "the record the filter has to miss is absent"` before the filter, so the coupling to the `app` fixture's `connection_hash` is now asserted rather than assumed.

---

tests/functional_tests/test_history_screen.py
tconbeer: Fixed in cbedf3e — dropped `f6` and `ctrl+r`. The loop is `("f1", "f10", "f12")` now, so every key in it asserts something: `f1` and `f12` on the screen stack, `f10` on `app.full_screen`.

On `ctrl+q`: you're right, and it's tested now — `test_quit_still_reaches_the_app_from_the_screen`, added before your review landed. Worth recording why, since it surprised me twice: `App._check_bindings` uses `reversed(self.screen._binding_chain)` for the priority pass and `_modal_binding_chain` only for the non-priority one, so a modal stops ordinary app bindings and priority ones go through regardless.

That is also what the screen's own two keys now rely on. `TextAreaPlus._handle_escape()` stops the event unconditionally, so a non-priority `escape` on the screen never fires while the preview has focus; both `history_screen.select_query` and `history_screen.cancel` are priority so the screen owns them wherever focus is. Raised the `_handle_escape` half upstream as [textual-textarea#345](https://github.com/tconbeer/textual-textarea/issues/345).

---

tests/functional_tests/test_history_screen.py
tconbeer: Fixed in cbedf3e. The highlight fix gave this its marker for free: the screen now holds `search`, the term `history` was read for, so `settle_filter()` polls that rather than sleeping, and the two inline repeats at :104-108 and :537-540 go through it.

One thing the fix taught me, which is worth flagging because it bit me immediately: polling on the applied term is only deterministic when the term *changes*. For a read that repeats the current term — the highlight case above — the condition is already true and the poll returns before the read starts. That test watches `read_history` instead.

---

src/harlequin/components/history_screen.py
tconbeer: Gone in cbedf3e — the class-level `BINDINGS` are removed entirely, along with the four forwarding methods, so there is no unbindable key left to register.

Ted asked for a different interaction model in the meantime, and it happens to answer this: each pane now owns its own keys rather than the screen proxying them. The list's arrows are `OptionList`'s own, the preview's are `TextArea`'s, and in the filter they do nothing on purpose — a text box is not a second set of keys for a list. So the "still driveable" claim this docstring made is no longer a claim the code makes.

What the screen does own is `tab`, and it owns it by overriding `focus_next`/`focus_previous` rather than by binding: Textual's own `Screen.BINDINGS` maps `tab` to `app.focus_next`, which resolves ahead of the keymap's `focus_next` and calls `self.screen.focus_next()` directly. Overriding the actions alone did nothing, which took a debug probe to see.

The new rules are pinned behaviorally — `test_the_screen_opens_on_the_list`, `test_tab_moves_between_the_list_and_the_preview`, `test_typing_over_the_list_starts_a_search`, `test_the_arrow_keys_do_not_reach_the_list_from_the_filter`, and the preview's own scroll tests.

---

src/harlequin/components/history_screen.py
tconbeer: Written up upstream as [textual-textarea#345](https://github.com/tconbeer/textual-textarea/issues/345), and the comment here points at it.

The shape changed under this in the meantime — Ted wanted the preview focusable so a long query can be scrolled from the keyboard — so `can_focus=False` is gone and both of the upstream facts you measured are moot. What replaced it is `read_only=True` plus Textual 8.2.8's `show_cursor=False`, which draws no cursor, drops the cursor-line highlight, and makes the arrow keys scroll like any other container.

But your point survives the change intact: `TextEditor.__init__` forwards `read_only` and not `show_cursor`, so this is still a subclass reaching into `self.text_input` after mount for something that belongs beside `read_only` in the constructor. That is the issue. It also carries the two neighbours this PR ran into: `_handle_escape()` stopping escape unconditionally (which is why both of the screen's keys are priority bindings now), and `TextEditor.BINDINGS` being live on a read-only editor — `ctrl+s` opened a save-path input in the preview's footer until `inherit_bindings=False` went on.

---

src/harlequin/app.tcss
tconbeer: All four taken, in cbedf3e. The block is down to sizing plus one comment:

```css
HistoryScreen #history_filter_pane { width: 1fr; }
HistoryScreen #history_filter      { height: 3; }
HistoryScreen QueryPreview         { width: 1fr; }
HistoryScreen OptionList           { background-tint: $background; padding: 0; }
```

- `#history_panes { height: 1fr }` and `#history_filter_pane { height: 100% }` dropped as the no-ops they were.
- `50%` → `1fr` twice. The dead column at odd widths is real; I had reproduced the 60 + 60 at 121 without recognising what it was.
- The `border`/`border-title-color`/`background` trio moved up into the grouped selector at :8-16, so the three new selectors sit beside their `:focus` counterparts at :22-28 rather than repeating the trio three times.
- The opaque background now says why: the screen fills the viewport, and `ModalScreen` is here for binding precedence, not for the dim backdrop. Your read was right — that is exactly the reason, and it was silent.

---

src/harlequin/actions.py
tconbeer: Back to `"Cancel"` in cbedf3e.

The either/or went away with the interaction rework rather than being papered over: escape's first meaning is leaving the screen, and emptying the filter is now the special case that applies only while the filter itself has focus. So the single noun phrase is the accurate one, not just the consistent one.

I left the dynamic-description idea alone. `_show_count` can update a border subtitle because that is a reactive on a widget; a binding's description is a frozen dataclass inside `_bindings`, so following focus would mean rebuilding the map on every focus change. Worth it if the footer starts feeling wrong in use, but not for this.

---

src/harlequin/components/history_screen.py
tconbeer: Both taken. The docstring is your suggestion verbatim, in cbedf3e.

The `type: ignore` comment is gone with its lines — `action_page_up`/`action_page_down` were the screen's forwarding methods, which the interaction rework removed along with the class-level `BINDINGS` they served.

---

CHANGELOG.md
tconbeer: Taken, in cbedf3e — your edit with two additions the rework earned:

```
- Harlequin's Query History (`F8`) now has a filter: start typing to narrow the list to queries that mention what you typed, searching your whole history ([#429](https://github.com/tconbeer/harlequin/issues/429)).
- Harlequin's Query History screen now lists its keys in a footer, ignores the app's keys while it is open, and shows no cursor in its preview pane, which scrolls with the arrow keys. A very long query no longer fills the list with one entry ([#850](https://github.com/tconbeer/harlequin/issues/850)).
```

"start typing" because that is now how you reach the filter — it no longer opens with focus. The preview scroll and the row height are new since your review: a 2,900-character query was filling the whole pane with one entry.

You're right that the modal behavior was the omission, and it was the most user-visible line in the PR going unmentioned.

On #851 — that's your call and I've left it alone. It is a closed community PR against the same issue and GitHub lists both as closing it, so a nod in the release notes seems fair to me, but it is not something to decide from inside this diff.

---

CHANGELOG.md
```diff
@@ -6,15 +6,20 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
-- Adds a `chdb` extra for installing the chDB adapter (`pip install 'harlequin[chdb]'`) and querying local ClickHouse databases ([#1144](https://github.com/tconbeer/harlequin/issues/1144)).
 - Harlequin's Query History (`F8`) now shows the queries hsql ran against the same database, and keeps them when a session ends without a clean quit ([#429](https://github.com/tconbeer/harlequin/issues/429)).
+- Harlequin's Query History (`F8`) can now be filtered: press `ctrl+f`, or just start typing, for a box that searches the text of every query in your history, and dropdowns that narrow the list to one program (Harlequin or hsql) or one outcome ([#429](https://github.com/tconbeer/harlequin/issues/429)).
+- Harlequin's Query History screen now lists its keys in a footer, ignores the app's keys while it is open, and shows no cursor in its preview pane, which scrolls with the arrow keys. A very long query no longer fills the list with one entry ([#850](https://github.com/tconbeer/harlequin/issues/850)).
```
tconbeer: Remove this changelog entry.

CHANGELOG.md
```diff
@@ -6,15 +6,20 @@ All notable changes to this project will be documented in this file.
...
 - `hsql --history` lists the queries Harlequin and hsql have run, newest first, and `hsql --history-search TERM` lists the ones whose SQL mentions TERM. Both take every format and layout flag a query does; `-P`, `-a` or a connection string narrows them to one database ([#429](https://github.com/tconbeer/harlequin/issues/429)).
 
 ### Bug Fixes
 
 - A value containing a newline no longer breaks `hsql`'s `table` and `vertical` output: it now spans as many lines as it has, with `+` marking each one that continues, as psql does.
 - The Data Exporter now writes the Feather File Version you chose, and its Compression and Chunk Size options now have an effect. Every export from the dialog wrote a legacy Version 1 file, which cannot hold compression.
 
+### Dependencies
+
+- Bumps `textual-textarea` to 0.18.4, which fixes a bug where a read-only editor — like the Query History preview — could still be changed from the keyboard.
```
tconbeer: Remove this one too

src/harlequin/components/history_screen.py
```diff
@@ -42,50 +64,228 @@ def visualize(self) -> object:
...
         history: History,
+        connection: str | None = None,
         theme: str = "harlequin",
         name: str | None = None,
         id: str | None = None,  # noqa: A002
         classes: str | None = None,
     ) -> None:
         super().__init__(name, id, classes)
         self.history = history
+        self.applied = Filters()
```
tconbeer: Call this applied_filters, not applied

src/harlequin/components/history_screen.py
```diff
@@ -42,50 +64,228 @@ def visualize(self) -> object:
...
+
+        def __init__(self, history: History | None, filters: Filters) -> None:
+            super().__init__()
+            self.history = history
+            self.filters = filters
+
     def __init__(
         self,
         history: History,
+        connection: str | None = None,
```
tconbeer: Is this just the connection hash? If so, call it that: connection_hash

CHANGELOG.md
tconbeer: Removed in b6b309c.

---

CHANGELOG.md
tconbeer: Removed in b6b309c, and the `### Dependencies` heading with it — that entry was its only one.

---

src/harlequin/components/history_screen.py
tconbeer: Renamed in b6b309c — `applied_filters` here and at both reads of it, plus `settle_filter()` in the tests.

---

src/harlequin/components/history_screen.py
tconbeer: It is — `app.py` passes `self.connection_hash` straight in. Renamed to `connection_hash` in b6b309c, on both the parameter and the attribute, so it matches what the app calls it.

The one place it stays `connection` is the call into the store, `History.recent(connection=self.connection_hash, ...)`, since that is the name of the column being filtered on.

---

PR #1162, opened 2026-09-16
Move the Windows tzdata check and download off the start-up path

### LINKED ISSUE #1134: Move the Windows tzdata check and download off the start-up path

tconbeer:
Closes #1134

- `windows_timezone.py` splits into `locate_tzdata() -> bool`, the cheap `pc.assume_timezone` probe that also points Arrow at Harlequin's own data dir if the system has none, and `download_tzdata()`, which fetches the IANA tarball and CLDR's `windowsZones.xml`.
- `Harlequin` takes `no_download_tzdata` and runs `_ensure_tzdata()`, a thread worker started from `on_mount` on Windows only. It posts `TzDataDownloadStarted` before downloading, which becomes a warning notification; a failed download lands in `_handle_worker_error` as a warning toast carrying the error's own text.
- `_fetch_data` waits on a `threading.Event` the worker sets in a `finally`, **bounded** by `TZDATA_WAIT_SECONDS` and proceeding on expiry. The event is pre-set everywhere the worker does not run, so nothing waits on any other platform.
- The download is bounded three ways: `SOCKET_TIMEOUT_SECONDS` (10s) per socket operation, `DOWNLOAD_DEADLINE_SECONDS` (30s) as an overall budget checked between 64 KiB chunks, and `MAX_DOWNLOAD_BYTES` (16 MB) on both the response and the archive's unpacked total. It also takes a `stop` Event that `action_quit` sets first thing, so quitting abandons a transfer rather than joining it.
- Extraction passes `filter="data"` where the runtime has it (3.10.12+), so a hostile archive cannot write outside the data directory.
- `cli.py` no longer calls the check inline and no longer has an exit-2 branch for it; it just passes `no_download_tzdata` through to the app.

Nothing on the way to the first paint needs the answer — the database is for pyarrow's `timestamptz` handling, which is not reached until a result set with a tz-aware timestamp arrives in the Results Viewer — so the check belongs behind the app rather than in front of it.

The failure is a warning toast rather than an error modal because a session with no timezone database is perfectly usable until a `timestamptz` shows up.

**The wait is unconditional but bounded**, and that is a real tradeoff: while a download is in flight, *every* query waits, up to `TZDATA_WAIT_SECONDS`. Making it conditional on `timestamptz` is not available — column types are only known after the fetch, `HarlequinCursor.columns()` returns adapter-specific short labels, and attempt-then-retry is unsound on a drained cursor. The toast says what actually happens. A fetch that proceeds on expiry is no worse off than one on a machine with no database at all.

**Quit is bounded, not instant.** A thread parked inside `response.read()` still waits out one socket timeout (≤10s) before it sees `stop`. An interruptible read or a daemon thread would make Ctrl-Q honest at the keypress; a daemon thread risks exiting mid-`extractall` and leaving a half-written directory that `locate_tzdata()` would then accept. Left as ≤10s — see the open thread on `_ensure_tzdata`.

`--no-download-tzdata` keeps its meaning: it skips the worker, exactly as it skipped the inline call. Its help text is corrected here, since the flag also skips the *lookup*, not only the download. `hsql` does not call any of this and still does not: its output path never converts a tz-aware Arrow column to Python (the text layouts read a duckdb VARCHAR cast; the viewer's `to_pylist()`/`column_content_widths` are what need the database).

- [ ] No.
- [x] Yes; I haven't opened a PR, but the gist of the change is: the `--no-download-tzdata` reference says it "may cause undesired behavior," which is now both vaguer and wronger than what the flag does — it skips the lookup as well as the download, a failed or skipped check no longer stops Harlequin from starting, and it only means a `timestamptz` in the Results Viewer may fail. The troubleshooting page for Windows timezones could say the check now runs in the background.

- [x] Yes.

`tests/unit_tests/test_windows_timezone.py` asserts **both** requests carry a timeout (mutation-checked: removing the second one's fails it), and covers the transfer deadline against a trickling peer, the size cap, and the stop flag. `tests/functional_tests/test_windows_timezone.py` fakes Windows through `sys.platform` and covers four behaviors: the app is up while a download is in flight and a query waits; the wait gives up on a download that never finishes; quit sets the stop flag; a failed download is a warning toast with no modal and does not stop a query. `tests/unit_tests/test_cli.py` covers the flag and the profile key reaching the app.

https://claude.ai/code/session_01UHfyDLHx6dV49rLKLAxiFj


src/harlequin/app.py
```diff
@@ -1572,6 +1628,10 @@ def _fetch_data(
         submitted_at: float,
         limit: RowLimit,
     ) -> None:
+        # Arrow cannot build a timestamptz column without a timezone
+        # database, and on Windows one may still be downloading. Already set
+        # everywhere else, so nothing waits here.
+        self._tzdata_ready.wait()
```
tconbeer: **Blocking.** This wait is unconditional, unbounded, and it parks a pool thread. Three separate user-visible failures come out of this one line.

**It is unconditional.** It's the first statement in `_fetch_data`, before the fetch loop and before anything inspects column types — and `_fetch_data` is the app's only fetch path. So on Windows, while the download is in flight, *every* query blocks; `select 1` waits exactly as long as a `timestamptz` query does. The test in this PR demonstrates it: it blocks on `select 1 as foo`. #1134 asked only what "a query that loads `timestamptz`" should do, and the toast at app.py:649-652 promises the narrow behavior.

**It is unbounded.** `DOWNLOAD_TIMEOUT_SECONDS` reaches `urlopen` as a *per-socket-operation* timeout, not a transfer deadline, so it does not bound `response.read()`. Against a server trickling one byte at a time, `urlopen(timeout=2.0)` + `read()` completed after 12.02s with no exception — 6x the nominal timeout. Scaled up, a peer emitting a byte every 29s keeps `download_tzdata()` alive forever, and `Event.wait()` here has no timeout. `getaddrinfo` also runs before the socket timeout is armed, so a wedged resolver is another unbounded path. Net: the 30s constant does not bound this wait.

**It parks a pool thread that never comes back.** `exclusive=True` cancels the previous worker's asyncio *task*, but its thread is already inside `wait()` and never returns its executor slot. Textual thread workers run on the asyncio **default** `ThreadPoolExecutor` (`worker.py:326`, `run_in_executor(None, ...)`), which is `min(32, cpu_count + 4)` — 8 on a 4-core box. `_connect`, `update_schema_data`, the catalog loader, the completer builders and **`_cancel_query`** all share it. With the executor pinned to 4 for legibility, 6 submissions left 4 threads permanently parked against 1 live worker, and an unrelated worker submitted afterwards never ran at all. So pressing Ctrl-Enter a few times during a stalled download — the natural reaction, since nothing is happening — wedges the app *and* removes the user's ability to cancel out of it.

This is worse on Windows, the only platform where `wants_tzdata` is true: `textual._win_sleep.sleep` also does `run_in_executor(None, ...)` and holds a thread for the whole sleep, so the 60s checkpoint interval, the completer-merge timer and the cursor blink give the pool nonzero baseline occupancy before the first query.

A bounded `wait(timeout=…)` plus a real transfer deadline (an overall budget around `read()`, not just `urlopen`) addresses all three, and proceeding on expiry rather than blocking is no worse than today's behavior when the database is simply absent.

---

src/harlequin/app.py
```diff
@@ -1559,6 +1594,27 @@ def _push_error_modal(self, title: str, header: str, error: BaseException) -> No
...
+        On a worker because nothing before the first result set needs the
+        answer; `_fetch_data` is what waits for it.
+        """
+        try:
+            if find_tzdata():
+                return
+            self.post_message(TzDataDownloadStarted())
+            download_tzdata()
+        finally:
+            self._tzdata_ready.set()
```
tconbeer, line 1630: **Blocking, and a regression:** quitting while this worker is downloading hangs Harlequin until the download finishes, and then needs a second Ctrl-C to escape.

Textual's shutdown calls `self.workers.cancel_all()`, which cancels the asyncio task but cannot touch a running thread, and `App.run()` ends in `asyncio.run(...)`, whose `finally` is `loop.run_until_complete(loop.shutdown_default_executor())`. So the interpreter waits for this thread. Measured directly — a thread worker sleeping 15s with `exit()` called at 0.5s:

```
run() returned after 15.03s (worker sleeps 15s, quit at 0.5s)
```

Same on 3.12. And once a SIGINT escapes that, the process hangs a second time in `threading._shutdown` → `concurrent/futures/thread.py:_python_exit` → `t.join()`; only a second Ctrl-C got out. By then `driver.close()` has already run, so what the user sees is a restored terminal with no shell prompt — it looks like a hang, not a shutdown.

Before this PR `check_and_install_tzdata()` ran on the main thread in `cli.py` *before* `tui.run()`, with the terminal still cooked: one Ctrl-C raised `KeyboardInterrupt` in the download and the process exited immediately. That's the behavior being given up. Combined with the unbounded read noted on `_fetch_data`, "hangs until the download completes" can mean indefinitely.

Worth deciding explicitly what quit should do here — a bounded transfer deadline caps it, and a daemon thread or an interruptible read would make Ctrl-Q honest.

---

src/harlequin/app.py
```diff
@@ -628,6 +644,16 @@ def screen_callback(dismiss_value: bool | None) -> None:
             ConfirmModal(prompt=message.instructions), callback=screen_callback
         )
 
+    @on(TzDataDownloadStarted)
+    def notify_tzdata_download(self, message: TzDataDownloadStarted) -> None:
+        self.notify(
+            "Harlequin is downloading a timezone database, which it needs for "
+            "timestamptz values. It will only do this once, and a result that "
+            "holds a timestamptz waits for it.",
```
tconbeer: **Blocking:** this tells the user something that isn't true. "a result that **holds a timestamptz** waits for it" — but `_fetch_data` waits on `_tzdata_ready` as its first statement, before the fetch loop and before any column type is known, so *every* result waits. This PR's own functional test blocks on `select 1 as foo`.

It's the worst place for that inaccuracy: a user who reads this toast and runs a non-`timestamptz` query to keep working during the download gets an unexplained hang, having just been told that wouldn't happen.

The download-started path also lost the docs link that the old `pretty_print_warning` carried — the failure message still has it, but this is now the only place a Windows user learns the check is happening at all. Suggestion:

```python
"Harlequin is downloading a timezone database, which it needs for "
"timestamptz values. It will only do this once, and query results "
"wait until it finishes.\n"
"https://harlequin.sh/docs/troubleshooting/timezone-windows",
```

(Plain URL rather than `[link]…[/]` since, unlike the failure toast, this call doesn't pass `markup=False`. Worth making the two consistent either way.)

---

tests/unit_tests/test_windows_timezone.py
```diff
@@ -0,0 +1,36 @@
...
+    timeouts: list[Any] = []
+
+    def fake_urlopen(url: str, *args: Any, **kwargs: Any) -> Any:
+        timeouts.append(kwargs.get("timeout"))
+        raise URLError("no network")
+
+    monkeypatch.setattr("harlequin.windows_timezone.urlopen", fake_urlopen)
+    with pytest.raises(HarlequinTzDataError):
+        download_tzdata()
+    assert timeouts == [DOWNLOAD_TIMEOUT_SECONDS]
```
tconbeer: **Blocking (cheap fix):** this tests one request, not every request. `fake_urlopen` raises `URLError` on its **first** call, so `download_tzdata()` aborts at the tarball request and the `windowsZones.xml` request is never reached — `timeouts` is a one-element list by construction, and `assert timeouts == [DOWNLOAD_TIMEOUT_SECONDS]` can't be anything else.

Confirmed by mutation: I deleted `timeout=DOWNLOAD_TIMEOUT_SECONDS` from the second `urlopen` and the file still reported `2 passed in 0.02s`.

The untested one is the request that follows a successful tar download, i.e. the one a user is most likely to be sitting on when the network goes bad — so the gap is exactly where the coverage was worth having. Letting the fake succeed on the first call and fail on the second (or return a tiny valid gzip for the tarball) pins both:

```python
def fake_urlopen(url: str, *args: Any, **kwargs: Any) -> Any:
    timeouts.append(kwargs.get("timeout"))
    raise URLError("no network")
```
→ have it count calls and only raise on the second, so the assertion becomes `[DOWNLOAD_TIMEOUT_SECONDS, DOWNLOAD_TIMEOUT_SECONDS]`.

---

src/harlequin/windows_timezone.py
```diff
@@ -30,41 +35,42 @@ def check_and_install_tzdata() -> None:
...
+    try:
+        response = urlopen(
+            "https://www.iana.org/time-zones/repository/tzdata-latest.tar.gz",
+            timeout=DOWNLOAD_TIMEOUT_SECONDS,
+        )
+        with TemporaryDirectory() as tmpdir:
+            tar_path = Path(tmpdir) / "tzdata.tar.gz"
+            with tar_path.open("wb") as f:
+                f.write(response.read())
+            tarfile.open(tar_path).extractall(HARLEQUIN_TZ_DATA_PATH)
```
tconbeer: `extractall()` with no `filter=` writes whatever the archive says, so a member named `../../…` or with an absolute path lands outside `HARLEQUIN_TZ_DATA_PATH`, with the user's full privileges. Whoever serves the tarball — a compromised host or mirror — can write to arbitrary paths: a `.py` on `sys.path`, the user's own `.harlequin.toml`, a Startup-folder entry.

**This is pre-existing** — byte-for-byte what `main` has, only reindented here — so it's not a defect this PR introduces, and I'd normally leave it. Raising it anyway for three reasons specific to this change:

1. The fix is one keyword in a function this PR is already splitting in two.
2. Moving the extraction onto a background worker makes it *less* visible: it used to print a warning to stderr before the TUI took the terminal, and now it happens silently while the user works.
3. On 3.14 the default becomes `filter='data'` regardless, so pinning it explicitly makes behavior uniform across 3.10–3.14 instead of version-dependent.

```python
tarfile.open(tar_path).extractall(HARLEQUIN_TZ_DATA_PATH, filter="data")
```

One thing to weigh: `filter=` was backported in **3.10.12**, and `requires-python = ">=3.10"` admits 3.10.0–3.10.11, where passing it raises `TypeError`. Your call whether that's a real floor.

I could not verify the real IANA tarball extracts cleanly under `data` — my sandbox blocks `iana.org` — but it's a flat set of regular files with no symlinks or absolute paths, all of which `data` permits, and the Windows 3.14 CI job exercises this path via the `conftest.py` fixture.

Same origin, lower severity, also pre-existing: `response.read()` has no size cap and neither does the decompression, so the same compromised host can serve a gzip bomb. The real payload is ~0.5 MB, so a cap in the low megabytes costs nothing.

---

src/harlequin/cli.py
```diff
@@ -616,15 +614,10 @@ def inner_cli(
         for key in hsql_profile_keys() - harlequin_options - declared_by_adapter:
             config.pop(key, None)
 
-        # detect and install (if necessary) a tzdatabase on Windows
+        # the app looks for a tzdatabase on a worker; nothing before the
+        # first result set needs one.
```
tconbeer: This comment documents code that is no longer in this file — it explains where the download *went*, not the `config.pop` it sits above. Per AGENTS.md, "Never reference code that was removed or refactored. A comment saying 'X rather than the old Y' reads to the next person as a live distinction; the diff is where a change is visible, and it stops being visible the moment the PR merges." The same sentence ("nothing before the first result set needs one") now appears three times across this diff, and the `_ensure_tzdata` docstring is the right home for it.

It also displaces the comment that actually explains this line, leaving the reader with a note about a different module on top and the relevant one underneath. Dropping it is the whole fix:

```python
# popped on every platform: the remaining config is the adapter's
no_download_tzdata = bool(config.pop("no_download_tzdata", False))
```

---

Separately, and not inline-able because the line is unchanged: **`--no-download-tzdata`'s help text at cli.py:544-547 is now wrong.** `wants_tzdata` gates `_ensure_tzdata()` entirely, so the flag skips `find_tzdata()` too — Arrow never gets pointed at a database Harlequin downloaded on a previous run. The help still says only "Prevent Harlequin from downloading an IANA timezone database", while this PR's own new strings say "opts out of **looking** for one" (app.py:328) and "without **checking** for one" (windows_timezone.py:70). This is the only description of the flag in the repo, so it's worth fixing here rather than leaving three descriptions that disagree. "May cause undesired behavior" is also vaguer than what the error path now tells the user.

---

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ All notable changes to this project will be documented in this file.
 - Harlequin's Query History (`F8`) now shows the queries hsql ran against the same database, and keeps them when a session ends without a clean quit ([#429](https://github.com/tconbeer/harlequin/issues/429)).
 - `hsql --history` lists the queries Harlequin and hsql have run, newest first, and `hsql --history-search TERM` lists the ones whose SQL mentions TERM. Both take every format and layout flag a query does; `-P`, `-a` or a connection string narrows them to one database ([#429](https://github.com/tconbeer/harlequin/issues/429)).
 
+### Performance
+
+- On Windows, Harlequin no longer waits for its timezone database check (and a possible download) before starting: the app opens right away and the check runs in the background. A download that fails is now a warning instead of an error that quits Harlequin ([#1134](https://github.com/tconbeer/harlequin/issues/1134)).
```
tconbeer: `Performance` covers the first half of this entry but not the second: "a warning instead of an error that quits Harlequin" is a change to a failure path, which belongs under `Bug Fixes`.

The first sentence also says the same thing three times — "no longer waits … before starting", "the app opens right away", "the check runs in the background" — and that third clause is the internals half your changelog rule cuts ("No implementation details. Harlequin users do not care how it works inside").

```markdown
### Performance

- On Windows, Harlequin no longer waits for its timezone database check (and a possible download) before starting up ([#1134](https://github.com/tconbeer/harlequin/issues/1134)).

### Bug Fixes

- On Windows, a timezone database that cannot be downloaded is now a warning; it no longer quits Harlequin ([#1134](https://github.com/tconbeer/harlequin/issues/1134)).
```

Issue reference format is right as-is.

---

src/harlequin/app.py
```diff
@@ -318,6 +324,13 @@ def __init__(
         # already started, by the command that built this app: `ssh` prompts for
         # a passphrase on the terminal Textual is about to take.
         self.ssh_tunnel = ssh_tunnel
+        # only Windows ships without a timezone database Arrow can read, and
+        # --no-download-tzdata is how a session opts out of looking for one.
+        self.wants_tzdata = sys.platform == "win32" and not no_download_tzdata
+        self._tzdata_ready = threading.Event()
+        """Set once Arrow has a timezone database, or is not getting one."""
+        if not self.wants_tzdata:
+            self._tzdata_ready.set()
```
tconbeer: Nit: `wants_tzdata` and `_tzdata_ready` encode one fact twice. `wants_tzdata` is read exactly once outside `__init__` (`on_mount`, line 490), and it is precisely "`_tzdata_ready` is not already set". Collapsing to the Event alone drops an attribute, the `if not …: set()` branch, and a public name:

```python
self._tzdata_ready = threading.Event()
"""Set once Arrow has a timezone database, or is not getting one."""
if no_download_tzdata or sys.platform != "win32":
    self._tzdata_ready.set()

# on_mount:
if not self._tzdata_ready.is_set():
    self._ensure_tzdata()
```

Two names for one fact can drift — and the new functional tests already show the failure mode, setting both by hand because setting one leaves the other lying.

Also: `wants_tzdata` is the only non-underscored attribute in `__init__` that exists purely to sequence a worker (`ssh_tunnel`, `show_files` etc. are all values a widget or handler consumes). `Harlequin` is what out-of-tree code instantiates, so an accidentally-public name is one someone can start depending on. Either fold it into the Event or call it `_wants_tzdata`.

(The `"""…"""` after the assignment is fine as-is — that attribute-docstring convention is well established in this file.)

---

src/harlequin/windows_timezone.py
```diff
@@ -9,18 +11,21 @@
...
+def find_tzdata() -> bool:
     """
     On Windows, Arrow expects to find a timezone database in the User's
-    Downloads folder. We check to see if it can find one there,
-    and if not, we override the tz db search path to a Harlequin-specific
-    location. If it still can't find one, we download the DB to that
-    harlequin-specific location.
+    Downloads folder. We check to see if it can find one there, and if not, we
+    override the tz db search path to a Harlequin-specific location. Returns
+    whether Arrow found a database in either place.
```
tconbeer: Nit: `find_tzdata` is named and typed like a pure probe but repoints Arrow's global tz search path as a side effect, and the `bool` return gives no hint that it did. Both current callers immediately follow it with `download_tzdata()`, so the split works for them — the hazard is the third caller. A debug screen or an `--info`-style report asking "do we have tz data?" would silently call `pa.set_timezone_db_path()` process-wide.

A name that says so (`locate_tzdata`, `point_arrow_at_tzdata`) costs nothing now. The docstring could also note that the override *persists* when the function returns `False`, since that's what makes `download_tzdata`'s closing `set_timezone_db_path` consistent:

```python
"""Whether Arrow can read a timezone database, from the user's Downloads
folder or from Harlequin's own location.

Leaves Arrow's search path pointed at Harlequin's location when Downloads
has none, which is where a download would land.
"""
```

The two-function split itself is the right call — the app needs to notify between the two phases, which a single `check_and_install` couldn't express without a callback.

---

tests/functional_tests/test_windows_timezone.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def blocking_download() -> None:
+        download_started.set()
+        assert finish_download.wait(timeout=10.0), "the fetch never released it"
+
+    monkeypatch.setattr("harlequin.app.find_tzdata", lambda: False)
+    monkeypatch.setattr("harlequin.app.download_tzdata", blocking_download)
+    app.wants_tzdata = True
+    app._tzdata_ready.clear()
+
```
tconbeer: Nit: these two lines fake Windows by reconstructing, from the outside, a decision `__init__` already makes — so any change to how the app represents that state (including collapsing the two attributes into one, as suggested on app.py:327) breaks both tests for reasons unrelated to what they assert.

The repo already has a seam: `tests/unit_tests/test_external.py:79` does `monkeypatch.setattr(sys, "platform", "win32")`, and `__init__` reads `sys.platform`. That needs building the app inside the test body from the `duckdb_adapter` fixture rather than taking the `app` fixture (whose construction runs before the test body, which is presumably why the pokes are here).

Two notes in the tests' favour, since I checked: the fixed-pause loop below is a **wide** race, not a flaky one — with the gate removed, `ResultsFetched` arrived after 1-3 `pilot.pause()` calls across five runs, and still 1 under 16 spinning processes on 4 cores, so `range(20)` has ~7-20x margin. And it has teeth: replacing `self._tzdata_ready.wait()` with `pass` makes it fail. The vacuous-pass path is closed too, since the `[results_fetched] = [...]` unpack raises on an empty list. The residual weakness is only that it pins "not yet, after 20 turns" rather than "blocked on the event".

---

src/harlequin/app.py
```diff
@@ -1572,6 +1628,10 @@ def _fetch_data(
         submitted_at: float,
         limit: RowLimit,
     ) -> None:
+        # Arrow cannot build a timestamptz column without a timezone
+        # database, and on Windows one may still be downloading. Already set
+        # everywhere else, so nothing waits here.
```
tconbeer: Nit, wording: "Already set everywhere else, so nothing waits here" reads on first pass as "no caller waits at this line" — the opposite of what the next line does. The subject is really the event, and "everywhere else" means every other *platform*. Given that the wait turns out to be broader than the surrounding prose implies, this is the comment most worth making unambiguous:

```python
# Arrow cannot build a timestamptz column without a timezone database,
# and on Windows one may still be downloading. The event is already set
# on every other platform, so only Windows ever blocks here.
```

---

tconbeer (commented):
## Review: scope, verdict, and what I did not check

**Scope.** Head `29507f7`, base `590e16a`, 8 files, +271/−57. Four narrow passes with explicitly disjoint lenses: correctness/concurrency, reuse/simplification, security, and comments/user-facing text. CI is **32/32 green**; locally `mypy` (154 files), `ruff`, `ruff format --check`, `lint-imports` (5/5 contracts) all clean, and both committed generator artifacts regenerate byte-identical.

**Verdict: do not merge as-is** (posted as a comment review rather than Request Changes, since GitHub won't let me request changes on a PR under this account). The design is right — the check genuinely does not belong in front of the first paint, and splitting `check_and_install_tzdata` into a cheap probe plus a download is the correct shape, since the app needs to notify between the two phases. But `self._tzdata_ready.wait()` at app.py:1634 is an untimed, unconditional block on a shared pool thread, and it produces three distinct user-visible failures. Green CI sees none of them, because none of them happen on Linux.

The blocking items:

1. **app.py:1634** — the wait is unconditional (every query blocks, not just `timestamptz`), unbounded (`urlopen`'s timeout is per-socket-operation, so it does not bound `read()`; measured 12.02s against a 2.0s timeout on a trickling peer), and it parks a thread in the shared default executor that `_cancel_query` also lives in — so a stalled download plus a few Ctrl-Enters wedges the app *and* the escape hatch.
2. **app.py:1597-1616** — quitting during the download hangs until it finishes (measured: `exit()` at 0.5s, `run()` returned at 15.03s), then hangs a second time in `concurrent.futures`' atexit join. A regression: the old pre-`tui.run()` download died on one Ctrl-C.
3. **app.py:649-652** — the toast promises that only a `timestamptz` result waits, which is false and is the worst possible place for it to be false.
4. **tests/unit_tests/test_windows_timezone.py:12** — `test_every_request_has_a_timeout` covers one request; removing the second `urlopen`'s timeout leaves it passing.
5. **windows_timezone.py:56** — `extractall()` without `filter=`. Pre-existing and normally out of scope, raised here because the fix is one keyword in a function this PR is already splitting, and because moving to a worker makes the extraction silent.

Items 1, 2 and 3 are one root cause, and a bounded wait plus a real transfer deadline resolves all three.

Non-blocking: a stale `--no-download-tzdata` help string (cli.py:544-547, now contradicted by this PR's own two new strings), the changelog heading, a comment documenting code that moved, and five nits — all inline.

**Confirmed clean,** so you don't re-check them: the `worker_name == "_ensure_tzdata"` branch really does match (`@work` sets `name=method.__name__`; `description=` feeds a different field); `exclusive`/`group="tzdata"` has one member, dispatches once, is submitted before `_connect` so it gets the first slot, and its `finally` covers every raise path; the exit-2 removal breaks nothing (the nine `exit_code == 2` assertions are all config errors and still pass); no dead code, and no reference to the old function name survives anywhere (I grepped every file type — the CI commit was the last one); `pa.set_timezone_db_path` races nothing in-tree; TLS verification is intact; the new notify path loses no redaction, because `pretty_print_error` did not redact either; the 3.14 `tarfile` default does not break extraction; the `TarFile` is collected before `TemporaryDirectory.__exit__`, so no Windows delete failure; and the option checklist (`TUI_ONLY_KEYS`, packaged schema, `hsql` exclusion) was already correct.

## Not checked

- **I did not test on Windows.** Every Windows-specific claim above is reasoned from source plus Linux probes of the same Textual/CPython mechanisms. The platform this code runs on is the one I could not exercise, so that is where the next bug is.
- I did not verify the real IANA tarball extracts under `filter="data"` — my sandbox blocks `iana.org`.
- I did not run the full suite or regenerate snapshots; I ran the touched test files plus `test_cli.py` (67 passed, 1 skipped) and the static gates. CI's 32 jobs cover the rest across Linux/macOS/Windows × 3.10-3.14.
- I did not verify the premise the whole wait rests on — that a `timestamptz` result actually crashes today without tzdata. I took that from the issue and the existing error text.
- `docs/plans/m1-hsql-technical-plan.md:118` is now stale about `windows_timezone.py` being "on the CLI path". I left it alone as a historical design doc rather than retro-editing it, but flagging that I made that call.
- I did not review the harlequin-web docs change, which the PR body notes is still owed — and the `--no-download-tzdata` wording fix should land there too.

Worth noting the second commit here caught a missed caller of the renamed function in `.github/workflows/test.yml` that a `.py`-only grep would miss; I re-grepped across all file types and that was the last one.

---

src/harlequin/app.py
tconbeer: Fixed in 2d5a9d4. The wait is now `wait(timeout=TZDATA_WAIT_SECONDS)` and proceeds when it expires — as you say, no worse off than a fetch on a machine that simply has no database.

On the unbounded half: `_read_bounded()` now reads in 64 KiB chunks and checks a monotonic budget between them, which is the only thing that ends a peer satisfying the socket timeout forever. The per-socket timeout dropped 30s → 10s, renamed `SOCKET_TIMEOUT_SECONDS`, since it was never bounding the transfer and the name implied it was. The size cap is in the same loop.

I did not make the wait *conditional* on `timestamptz`. Column types are only known after the fetch, and `HarlequinCursor.columns()` returns adapter-specific short labels rather than something core can test for — so the honest options were "bound it" or "attempt, catch the Arrow error, wait and retry", and the retry is unsound because the cursor is already drained. So it stays unconditional, and the toast now says so.

Still true after the fix, and worth naming: during a download every query waits up to 30s. That is the cost of the unconditional wait; it is bounded and it no longer strands the thread, but it is not zero.

---

src/harlequin/app.py
tconbeer: Fixed in 2d5a9d4 — took your wording, plus the docs link and `markup=False` so the two toasts are consistent:

```python
"Harlequin is downloading a timezone database, which it needs for "
"timestamptz values. It will only do this once, and query results "
f"wait until it finishes.\n{TZ_DATA_DOCS_URL}",
```

`TZ_DATA_DOCS_URL` is now a module constant in `windows_timezone`, so the toast and the failure message quote one string rather than two copies.

---

src/harlequin/app.py
tconbeer: Fixed in 2d5a9d4, but **bounded rather than instant** — worth being precise, since you measured the old behavior.

`download_tzdata` takes a `stop: threading.Event`, checked between chunks, and `action_quit` sets it as its *first* statement so the transfer unwinds while the caches are written rather than after. `test_quitting_abandons_a_download_in_flight` pins it: the fake download asserts it sees the flag.

What that does **not** buy: if the thread is parked inside `response.read()` when you quit, it still waits out one socket timeout — now ≤10s — before it reaches the next `stop` check. So the worst case went from "as long as the download takes, possibly forever" to ≤10s, not to zero. Making it honest at the keypress needs an interruptible read or a daemon thread, and I did not do either; a daemon thread would let the interpreter exit mid-`extractall`, which risks a half-written tz directory that `locate_tzdata()` would then accept on the next run.

Your point that this is a regression from the pre-`tui.run()` Ctrl-C stands, partially: one Ctrl-C used to kill it immediately. Happy to take the interruptible-read version if ≤10s isn't good enough.

---

src/harlequin/windows_timezone.py
tconbeer: Fixed in 2d5a9d4. Agreed on all three reasons — the silent-on-a-worker one especially, since that is this PR's doing.

On the 3.10.0–3.10.11 floor: `requires-python = ">=3.10"` does admit them, so I guarded rather than assumed:

```python
if hasattr(tarfile, "data_filter"):
    tar.extractall(HARLEQUIN_TZ_DATA_PATH, filter="data")
else:
    tar.extractall(HARLEQUIN_TZ_DATA_PATH)  # noqa: S202
```

That secures everyone from 3.10.12 up and leaves older patch releases exactly where `main` already has them, rather than turning the download into a `TypeError` they'd only see as "could not download".

**I could check the thing your sandbox blocked**, from the other direction — I built the archive rather than fetching it. A flat archive of tzdata-shaped members (`northamerica`, `europe`, `zone.tab`) extracts cleanly under `data`, and a member named `../../escaped.py` is refused:

```
flat archive extracted: ['europe', 'northamerica', 'zone.tab']
traversal refused: OutsideDestinationError '../../escaped.py' would be extracted to '/tmp/escaped.py', which is outside the destination
```

That is the shape of the real payload, not the real payload — if IANA ever ships a symlink, `data` would reject it and the download would fail closed.

Gzip bomb also capped: `MAX_DOWNLOAD_BYTES = 16 MB` bounds the response in `_read_bounded`, and `_extract` sums the members' declared sizes against the same cap before unpacking, so a small archive that claims to unpack to gigabytes is refused before any write.

---

tests/unit_tests/test_windows_timezone.py
tconbeer: Fixed in 2d5a9d4. The fake now returns a real empty `.tar.gz` on the first call and raises on the second, so the assertion is `[SOCKET_TIMEOUT_SECONDS, SOCKET_TIMEOUT_SECONDS]` and it is reached through the extraction step rather than around it. The tarball is built with `tarfile` at call time rather than checked in as a byte blob.

I ran your mutation both ways to confirm the test now has the teeth it was missing:

```
=== with the second timeout removed (must FAIL):
tests/unit_tests/test_windows_timezone.py:52: AssertionError
FAILED tests/unit_tests/test_windows_timezone.py::test_every_request_has_a_timeout
=== restored (must PASS):
1 passed in 0.02s
```

Three more cases came with the fix, each pinning a bound that did not exist before: `test_a_trickling_peer_hits_the_deadline` (the case a per-socket timeout cannot catch), `test_an_oversized_response_is_refused`, and `test_stop_abandons_the_download_without_an_error`.

---

src/harlequin/cli.py
tconbeer: Both fixed in 2d5a9d4.

The comment is gone — you're right that it documented where the code went rather than the line it sat on, and the `_ensure_tzdata` docstring already says it.

The help text now describes the whole flag:

```
(Windows Only) Prevent Harlequin from looking for an IANA timezone
database, or downloading one if it is missing. Harlequin may fail to
load timestamptz values into the Results Viewer.
```

"looking for" rather than only "downloading", which is what `wants_tzdata` actually gated, and the consequence stated instead of "may cause undesired behavior". That leaves one description in the repo instead of three that disagreed.

Worth noting it does not reach the generated schema: `config-v1.json` gives every TUI-only key the same generic `"Read by the harlequin IDE; hsql ignores it. See \`harlequin --help\`."`, so both generators still regenerate byte-identical.

---

CHANGELOG.md
tconbeer: Taken verbatim in 2d5a9d4 — both entries, both headings. You're right that the failure-path change is a bug fix, and that the third clause was the internals half the rule cuts.

---

src/harlequin/app.py
tconbeer: Collapsed in 2d5a9d4, to the Event alone:

```python
self._tzdata_ready = threading.Event()
"""Set once Arrow has a timezone database, or is not getting one."""
self._tzdata_stop = threading.Event()
"""Set on quit, to abandon a download rather than wait it out."""
# only Windows ships without a database Arrow can read, and
# --no-download-tzdata is how a session opts out of looking for one.
if sys.platform != "win32" or no_download_tzdata:
    self._tzdata_ready.set()

# on_mount:
if not self._tzdata_ready.is_set():
    self._ensure_tzdata()
```

The public-name point is the one that decided it — `Harlequin` is what out-of-tree code constructs, and `wants_tzdata` was worth nothing to such a caller.

Your read of the tests was right, and it's load-bearing here: they set both attributes by hand precisely because setting one left the other lying. With the attribute gone they don't set anything — a `windows_app` fixture patches `sys.platform` and constructs the app, per your note on the other thread, so the tests no longer reconstruct the decision from outside at all.

`_tzdata_stop` is a second Event, but it is a genuinely different fact (quit requested, not database ready) and it crosses into `download_tzdata`.

---

tests/functional_tests/test_windows_timezone.py
tconbeer: Taken in 2d5a9d4 — the seam you pointed at, moved into a fixture so all four tests share it:

```python
@pytest.fixture
def windows_app(
    monkeypatch: pytest.MonkeyPatch, duckdb_adapter: type[HarlequinAdapter]
) -> Harlequin:
    monkeypatch.setattr(sys, "platform", "win32")
    return Harlequin(duckdb_adapter([":memory:"], no_init=True), connection_hash="foo")
```

Your diagnosis of why the pokes were there was exactly right: the `app` fixture is constructed before the test body, so the patch had to move into a fixture that builds the app after it. Nothing reaches into app state now, which also means the `wants_tzdata` collapse on the other thread didn't touch these tests.

Thanks for checking the pause loop rather than just flagging it — I'd have tightened it on suspicion otherwise. The new `test_a_fetch_gives_up_on_a_download_that_never_finishes` uses the opposite shape (`range(200)` with an early break, `TZDATA_WAIT_SECONDS` patched to 0.25) because it asserts arrival rather than absence, so it can exit as soon as the message lands.

The residual weakness you name — "not yet, after 20 turns" rather than "blocked on the event" — is still there. Pinning it properly means observing the thread's state, which costs more than it's worth given the mutation check passes.

---

src/harlequin/app.py
tconbeer: Reworded in 2d5a9d4, your version plus what the new bound is doing there:

```python
# Arrow cannot build a timestamptz column without a timezone database,
# and on Windows one may still be downloading. The event is already set
# on every other platform, so only Windows ever blocks here. Bounded,
# because a worker that never returns keeps its thread out of the pool
# `_cancel_query` also runs in; a fetch that goes ahead without the
# database is no worse off than one on a machine that has none.
```

You were right that the old phrasing read as the opposite of what the next line does.

---

src/harlequin/windows_timezone.py
tconbeer: Renamed to `locate_tzdata` in 2d5a9d4, with the docstring saying what persists:

```python
def locate_tzdata() -> bool:
    """Whether Arrow can read a timezone database, from the user's Downloads
    folder or from Harlequin's own location.

    Leaves Arrow's search path pointed at Harlequin's location when Downloads
    has none, which is where a download would put one.
    """
```

The third-caller hazard is the right frame — an `--info`-style probe silently repointing `pa.set_timezone_db_path()` process-wide is exactly the bug that would be hard to find, and `find_*` returning `bool` reads like it could never do that.

---

Issue #991, opened 2026-07-31
Crash: "task_done() called too many times" in database tree loader on catalog refresh


lexo13:
## Summary

Harlequin intermittently crashes with `ValueError: task_done() called too many times`,
raised from the `_database_tree_background_loader` worker. It happens when the data-catalog
tree is reloaded (for example the automatic refresh that runs after executing a query) while
the previous background loader is still processing a node.

## Environment

- harlequin 2.5.2
- harlequin-mysql 1.3.0 (the failing code is the shared `DatabaseTree`, so other adapters
  should be affected too)
- textual 6.4.0
- Python 3.13, Linux

## Traceback (abridged)

```
ValueError: task_done() called too many times
  Worker ERROR name='_database_tree_background_loader' description='_loader()'
  .../textual/worker.py:370 in _run  ->  self._result = await self.run()
```

## Root cause (from reading the source)

In `harlequin/components/data_catalog/database_tree.py`:

- `_loader()` (the `@work(name="_database_tree_background_loader")` worker, ~L308) pulls a
  node off `self._load_queue`, processes it, and in its `finally` block calls
  `self._load_queue.task_done()`.
- `reload()` (~L112) **replaces** the queue with a brand-new `PriorityQueue()` and then
  starts another `self._loader()`.
- The `@work` decorator has no `exclusive=True`, so the previous `_loader` worker is never
  cancelled. If it has already pulled an item from the *old* queue and is mid-processing when
  `reload()` swaps `self._load_queue`, its `finally: self._load_queue.task_done()` then runs
  against the *new* queue, which has no outstanding `get()` -> `task_done() called too many
  times`, and the app crashes.

Because Harlequin auto-refreshes the catalog after each query
(`app.update_schema_data()` -> database-tree reload), any query can trigger the reload, so
the race is easy to hit during normal use.

## Steps to reproduce

1. Open Harlequin against a database (observed with the mysql adapter).
2. Run queries a few times so the post-query catalog refresh fires while the tree is still
   (re)loading.
3. Intermittently the app crashes with the traceback above.

## Suggested fix

Make the loader robust to a queue swap, e.g. one (or both) of:

- decorate `_loader` with `@work(exclusive=True)`, or explicitly cancel the running loader
  inside `reload()`, so only one loader runs at a time;
- have `_loader` capture a local reference to the queue it took the item from and call
  `task_done()` on that same object, instead of on `self._load_queue`.

tconbeer:
[...]

PR #1163, opened 2026-09-16
Stop the stateful tests from waiting on the clock

### LINKED ISSUE #1133: Review the stateful tests for races, so flakes stop riding on re-runs
`tests/functional_tests/test_ssh_recovery.py` flaked twice on [#1132](https://github.com/tconbeer/harlequin/pull/1132) — macOS only, on 3.13 on one commit and 3.14 on another, and green on a re-run both times. That PR's diff touches nothing in the path (it adds the hsql warm-session client), main was green, and the test passed 12/12 locally. So it is the test, not the change, and a re-run is currently the only thing standing between it and a red PR.

Rather than fix that one test, this is a request to review **every test that holds state across an await** — the app tests, the SSH tests, and the fixtures under them — for the same class of race.

## The example

`test_a_refresh_that_reconnects_does_not_ask_for_a_second_one` failed with:

```
ValueError: No Tab with id 'tab-1'
```

from `Tabs.validate_active`, reached through `EditorCollection.action_new_buffer`.

Two things activate the first tab. `action_new_buffer` does it explicitly:

```python
await self.tabs.add_tab(Tab(f"Tab {self.counter}", id=new_buffer_id))
if activate:
    self.tabs.active = new_buffer_id
```

and `Tabs.add_tab` already schedules its own `refresh_active()` on the `from_empty` branch, which sets `active` once the mount resolves. `validate_active` raises if the tab is not in `#tabs-list` when the deferred one fires. Two writers, one of them deferred past an await the test does not control — which is exactly the shape that only shows up on a slow runner.

That specific one is small enough to fix in place (`if activate and self.tab_count > 1`, or drop the explicit set for the first tab), but it is the symptom worth generalizing from.

## What to review

The same shape recurs across the suite:

- **State read after `wait_for_workers`.** The fixture returns when the worker's future resolves, which is *before* the app has handled the message carrying the result. `test_ssh_recovery.py` already carries `_rendered_catalog` and `_first_database_node` — bounded poll loops written to work around exactly this. Anywhere else that reads app state straight after `wait_for_workers` is relying on the pump having drained, and the ones that do it under a bare `await pilot.pause()` are the ones to look at first (273 `pilot.pause()` calls across `tests/functional_tests/`).
- **Widget state written from two places, one deferred.** The tab case above. Worth a sweep of the `AwaitComplete`-returning Textual calls we `await` and then immediately write to.
- **Real clocks and real processes.** `wait_until(..., seconds=10)` in `test_ssh.py` and `test_ssh_recovery.py`, the fake ssh child, the loopback ports from `free_port()` (bind-then-close leaves a window where something else can take the port), and `pilot.pause(0.5)` in `test_app.py`. These are the tests most sensitive to a loaded runner, and the macOS runners are the loaded ones.
- **Teardown ordering.** `dropping_tunnel.stop()` in a `finally` outside `app.run_test()`, so the app is already shutting down while the child is still up.

## Why it matters now

Windows already runs with `--force-flaky --max-runs=3`, which hides this class of failure there. macOS does not, so macOS is where it surfaces — and a re-run is not a fix, it is a coin flip we pay for on every PR. #1129's own review round already fixed one race in this family; the remaining ones are worth finding on purpose rather than one red CI run at a time.

## Done looks like

Each stateful test either waits on a condition it can observe, or is restructured so there is nothing to wait on — and no test depends on a duration. Where a wait is genuinely unavoidable, it is a shared helper with one place to fix, not a local `sleep`.

tconbeer:
Closes #1133

#1151 fixed the crash the `test_ssh_recovery.py` flake surfaced. This is the sweep behind it, over every test that holds state across an await. No source change — the whole diff is tests.

`tests/waiting.py` is now the one place a wait lives:

- `wait_until` polls state another thread or child process sets.
- `wait_for` / `wait_for_value` / `wait_for_messages` poll state the app sets, pumping its messages as they go, and raise naming the condition that never held.
- `settle` is the only duration, for establishing that nothing *else* happened.
- `free_port` / `accepts` / `on_a_free_port` are the loopback helpers, which had three copies between the two ssh modules.

The app-shaped conditions are fixtures over those, in `tests/functional_tests/conftest.py`: `wait_for_editor`, `wait_for_table`, `wait_for_error_modal`, `wait_for_catalog_tree`, `rendered_catalog`, `first_database_node`. A test asks for the editor, the results table, the error modal or the loaded catalog rather than pumping and hoping, and binds what the wait returned instead of reading back through `app.editor`.

Nothing in `tests/functional_tests/` sleeps, pumps a fixed number of times, or loops unbounded any more. The two local poll loops `test_ssh_recovery.py` carried (`_rendered_catalog`, `_first_database_node`) became shared fixtures, and `_until` in `test_hsql_serve.py` became `wait_until`.

Two things from the issue that were not waits:

- `free_port()` drew from the pool the OS hands out for every outbound connection, so a port could be taken between the bind that found it and the child that was given it. It now draws from below that range (20000–30000, under 32768 on Linux and 49152 on macOS and Windows), never names the same port twice in a process, and `on_a_free_port` retries a child that lost the port anyway.
- The dropping tunnel is stopped by its fixture rather than each test's `finally`, so a test that fails before its own teardown leaves no child behind.

A wait on a duration asserts the machine is fast enough, which is why this class of failure shows up on macOS and not locally. Polling a named condition makes a loaded runner slower rather than red, and puts the bound in one place instead of an iteration count copied into every module.

The alternative for the `wait_for_workers` gap — it returns when the worker's future resolves, which is before the app has handled the message carrying the result — was to strengthen that fixture so state is safe to read straight after it. There is no signal to strengthen it with: the handler runs on the pump, and "the pump has drained" is not a thing the fixture can observe without a pilot. So the fixture is unchanged and the waits above are what tests read state through; `AGENTS.md` now says so.

For the port, holding the socket open was not an option (the child has to bind it, and `SO_REUSEADDR` does not allow two live listeners), so the fix is to compete with fewer processes and retry when that is not enough. The range is a judgement call: it is below the ephemeral range every supported platform ships with, and each candidate is still test-bound before it is handed out.

Tradeoff worth naming: the `app.editor` → bound `editor` rewrite is a large, mechanical share of the diff. It follows from dropping the `while app.editor is None` loops, which were what narrowed the `CodeEditor | None` for mypy.

- [x] No.

- [x] Yes.

  Not applicable — no user-facing change. The user-facing half of #1133 was the crash, and #1151 already has its entry (released in 2.14.0).

https://claude.ai/code/session_01B64Sxdqhc4WyrwR2gXDV5L

---
_Generated by [Claude Code](https://claude.ai/code/session_01B64Sxdqhc4WyrwR2gXDV5L)_


tests/functional_tests/test_app.py
```diff
@@ -79,33 +78,28 @@ async def test_queries_do_not_crash_all_adapters(
...
             assert query_submitted_message.queries == [query]
-            await wait_for_workers(app)
-            await pilot.pause()
-            [query_executed_message] = [
-                m for m in messages if isinstance(m, QueriesExecuted)
-            ]
+            [query_executed_message] = await wait_for_messages(
+                pilot, messages, QueriesExecuted
+            )
             assert query_executed_message.cursors
```
tconbeer, line 98: `select 1 where false` no longer waits for the fetch phase at all.

On main the `if query:` branch ended with `await wait_for_workers(app)`, which loops through the handoff — `_execute_query` posts `QueriesExecuted`, whose handler starts `_fetch_data` — so every parametrization waited for the fetch before the test returned. Here the branch ends at `QueriesExecuted`, and `select 1 where false` is then excluded from the only remaining wait, so it returns as soon as that message is dispatched. `_fetch_data` and `push_table` may never run inside the test body.

That parametrization is in the list specifically for the empty-result path — the one AGENTS.md calls out as `create_backend()` with no rows and a header. It's the case most worth watching a crash in, and it's the one case now unwatched.

Either drop the special case and let it `await wait_for_table(...)` too (an empty result still renders a table with a header), or put a trailing `await wait_for_workers(app)` inside `if query:`.

---

tests/waiting.py
```diff
@@ -0,0 +1,191 @@
...
+    await wait_for(
+        pilot,
+        lambda: len(collected()) >= count,
+        description=(
+            f"{count} {message_type.__name__} message"
+            f"{'' if count == 1 else 's'}, saw {len(collected())}"
+        ),
+        seconds=seconds,
+    )
+    return collected()[:count]
```
tconbeer, line 144: Truncating to `count` turns "exactly N" into "at least N", and that was a real assertion.

`[query_submitted_message] = [m for m in messages if isinstance(m, QuerySubmitted)]` raises `ValueError: too many values to unpack` when a second message was posted. `[x] = await wait_for_messages(...)` cannot. Eleven sites in `test_app.py` lose it, including the ones whose names are the claim: `test_app.py:167` still carries the comment `# should only run one query` over a wait that no longer checks "only", and `test_single_query_terminated_with_semicolon` exists to prove a whitespace-only statement produces no extra submission.

`[*_, x] = await wait_for_messages(..., count=3)` at line 254 also changes meaning: main took the last message, this takes the third.

An `exactly: bool = False` that asserts `len(collected()) == count` after the wait would keep the strength where it was load-bearing, and the `[:count]` slice where it isn't.

---

tests/functional_tests/test_ssh_recovery.py
```diff
@@ -182,43 +151,37 @@ async def test_a_reconnect_leaves_a_catalog_that_still_loads(
...
+        # the tree on its own, which is the workaround this replaces
+        schema_updates.clear()
+        app.post_message(QuerySubmitted(queries=["select 1"], limit=None))
+        await wait_for_workers(app)
+        rebuilt_catalog = await rendered_catalog(pilot, app, replacing=first_catalog)
+
+        assert len(schema_updates) == 1
+        assert rebuilt_catalog is not first_catalog
+        assert app.connection is not first_connection
+        db_node = await first_database_node(pilot, app)
```
tconbeer, line 172: This is the one wait in the diff I think can be satisfied by stale state.

`first_database_node` waits for `tree.root.children` to be non-empty — which is already true here, of the *pre-refresh* tree. The gate before it, `rendered_catalog(..., replacing=first_catalog)`, returns the moment the `catalog` var is assigned, and `watch_catalog` is `async`: `await self.reload()` (which is what remounts the nodes) has not run yet. `first_database_node`'s own docstring says exactly this — "Setting the catalog and mounting its nodes are two steps" — but its condition doesn't enforce it.

So `db_node` can be the node from the old catalog, and `assert db_node.data.connection is app.connection` on the next line fails. It fails loudly rather than passing wrongly, which is the good direction, but it's a spurious failure on a loaded runner — the shape this PR is removing everywhere else.

In practice `pilot.pause(None)` drains the reload task often enough that it passes, and that's the part I'd rather not rely on. Gating on the data instead of the count fixes it:

```python
lambda: next(
    (n for n in tree.root.children if getattr(n.data, "connection", None) is app.connection),
    None,
)
```

or have `rendered_catalog` wait for `not tree.loading` after the swap.

---

tests/functional_tests/conftest.py
```diff
@@ -231,14 +236,19 @@ def expand_catalog_node() -> Callable[[Pilot, TreeNode[CatalogItem]], Awaitable[
...
-            if data is None or (
+            return data is None or (
                 getattr(data, "loaded", True)
                 and len(node.children) == len(data.children)
-            ):
-                return
-            await pilot.pause()
+            )
+
+        await wait_for(
```
tconbeer: This condition is satisfiable with the "loading…" placeholder still rendered, for any node whose real child count is 1.

`_load_children` runs on a worker thread and sets `item.children` and `item.loaded = True` **there** (`database_tree.py:495-498`). The node is repopulated later, back on the event loop, in `_loader` → `_populate_node`. In between, the node still holds the single `LOADING_ITEM` that `_show_loading_placeholder` added. So:

- `data.loaded` → `True` (set on the thread)
- `len(node.children)` → 1 (the placeholder)
- `len(data.children)` → 1 (the real child)

…and the wait returns on a tree showing "loading…".

The fixture databases hit this at every level: `tests/data/functional_tests/tiny/schema.sql` is one schema → one table `foo` → one column, and `small`'s `main` schema has one table. Any test that then reads `node.children[0].data` or snapshots the catalog gets the placeholder.

`DatabaseTree` already has the predicate for it:

```python
and not DatabaseTree._has_placeholder(node)
```

The condition is carried over verbatim from the `while True` loop this replaces, so it isn't a regression — but promoting it to the shared fixture that every catalog test now waits through seems like the moment to fix it.

---

tests/waiting.py
```diff
@@ -0,0 +1,191 @@
...
+    def collected() -> list[MessageT]:
+        return [m for m in messages if isinstance(m, message_type)]
+
+    await wait_for(
+        pilot,
+        lambda: len(collected()) >= count,
+        description=(
+            f"{count} {message_type.__name__} message"
+            f"{'' if count == 1 else 's'}, saw {len(collected())}"
+        ),
```
tconbeer: `description=` is an argument expression, so `len(collected())` is evaluated once before the first poll. On timeout it always reports the count from *before* waiting — in practice `saw 0` — which is exactly the number that isn't useful. A wait for 2 `ResultsFetched` where 1 arrived says "saw 0".

Making `description` a `str | Callable[[], str]` and resolving it at the raise fixes it for every future caller; interpolating inside the `raise` in `wait_for` is the smaller change. (It's the only eagerly-interpolating `description` in the diff that isn't fixed for the duration of the wait — `conftest.py:250`, `test_data_catalog.py:129` and `conftest.py:315` all interpolate values that can't change.)

---

tests/waiting.py
```diff
@@ -0,0 +1,191 @@
...
+def free_port() -> int:
+    """A loopback port nothing is listening on, that nothing is about to take.
+
+    Binding port 0 and closing it hands the port back to the pool the OS draws
+    from for every outbound connection, so the next one the machine makes can
+    take it before the child that was given it binds it -- which on a loaded
+    runner is a test that fails for a reason nothing in it explains. Drawing
+    from outside that pool leaves only another test as a competitor, and a
+    caller that hands the port to a child still has `on_a_free_port`.
+    """
```
tconbeer: Three of AGENTS.md's docstring rules in one docstring.

"Binding port 0 and closing it hands the port back to the pool…" describes **the implementation this PR just deleted** from `test_ssh.py`. Per AGENTS.md that reads to the next person as a live distinction; the diff is where it's visible, and only until this merges. "…which on a loaded runner is a test that fails for a reason nothing in it explains" is the PR description, and it's the third copy of that sentence in this diff (here, the module docstring, and `AGENTS.md:69`). And the first clause restates `_PORT_RANGE`'s docstring eight lines up.

```python
def free_port() -> int:
    """A loopback port nothing is listening on, drawn from below the ephemeral range.

    It can only report a port that was free a moment ago; a caller that hands
    one to a child has `on_a_free_port`.
    """
```

Same shape applies to the module docstring's second paragraph (lines 3-4 — the rationale for the whole PR, duplicated in AGENTS.md), `on_a_free_port`'s "and the answer to it is another port" (restates the name), and `wait_for_messages`' "so the list a hook fills is state to wait on like any other" (argues for the helper's existence).

---

tests/waiting.py
```diff
@@ -0,0 +1,191 @@
...
+_PORT_RANGE = (20000, 30000)
+"""Where a test's ports come from: below the ephemeral range every supported
+platform hands out on its own (32768 on Linux, 49152 on macOS and Windows), so
+only something that asked for a port by number competes for one."""
+
+_handed_out: set[int] = set()
+"""Ports this process has already named, which it must not name twice."""
+
+_draw = random.Random(os.getpid())
+"""Seeded per process, so two xdist workers walk different candidates."""
```
tconbeer: Seeding from the pid is strictly worse than not seeding.

An unseeded `random.Random()` already seeds from `os.urandom`, which gives you the stated goal — two xdist workers walking different candidates — unconditionally. The pid seed only *adds* a collision mode: two processes in separate PID namespaces sharing one network namespace (two containers on a host network, a normal CI shape) can hold the same pid, so they get the same seed, the same candidate sequence, and compete for the same ports in the same order. Recycled pids across concurrent jobs on a self-hosted runner do the same.

Dropping the argument is the whole fix.

---

tests/waiting.py
```diff
@@ -0,0 +1,191 @@
...
+    `free_port` can only report a port that was free a moment ago, so a child
+    that binds it races every other process on the machine for it. Losing that
+    race raises `retry_on`, and the answer to it is another port.
+    """
+    for attempt in range(attempts):
+        try:
+            return start(free_port())
+        except retry_on:
+            if attempt == attempts - 1:
+                raise
```
tconbeer, line 206: `retry_on=HarlequinSshError` is much wider than "the port was taken", and the worst case is five minutes of silence.

`SshTunnel.start()` raises `HarlequinSshError` for every failure: the child exiting, the readiness **timeout**, and `Popen` failing outright. `SshTunnel.timeout` defaults to `DEFAULT_SSH_TIMEOUT = 60.0`, and neither `dropping_tunnel` (`test_ssh_recovery.py:62`) nor `child_tunnel(free_port())` at `test_ssh.py:450, 503, 513` passes a `timeout=`. So a fixture that is genuinely broken — not racing for a port — burns 5 × 60s = 300s per test and then surfaces attempt 5's error with nothing saying it was never a port race. That's a worse debugging experience than the flake this replaces. The sites that pass `timeout=0.5` are bounded at 2.5s.

The two cases are distinguishable: under `ExitOnForwardFailure=yes` a child that can't bind exits, so `returncode is not None`, where the timeout path leaves it `None`. Narrowing `retry_on` to that, or just giving `dropping_tunnel` a short explicit `timeout=`, bounds it.

---

tests/waiting.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def wait_until(predicate: Callable[[], bool], *, seconds: float = TIMEOUT) -> bool:
+    """Poll until something another thread or process does becomes true."""
+    deadline = time.monotonic() + seconds
+    while time.monotonic() < deadline:
+        if predicate():
+            return True
+        time.sleep(POLL_INTERVAL)
+    return predicate()
```
tconbeer: Two failure shapes for one idea, in the module whose job is to have one.

`wait_for` takes a `description` and raises with it. `wait_until` returns `bool`, so all nine call sites spell it `assert wait_until(...)` and get a bare `assert False` on timeout — no word about what never happened. That's a small regression where it replaced `_until` in `test_hsql_serve.py`, which did raise with a message.

Give it a `description` and have it raise, and the nine `assert` prefixes go away with it.

---

tests/functional_tests/test_ssh_recovery.py
```diff
@@ -22,96 +20,66 @@
...
-) -> Catalog:
-    """The catalog the tree is holding, once it is not `replacing`.
+async def _wait_for_drop(pilot: Pilot, tunnel: SshTunnel) -> None:
+    """Wait for the watcher thread to notice the child let the forward go.
 
-    `wait_for_workers` returns when `get_catalog()` has answered, which is
-    before the app has handled the message carrying its result, so reading the
-    tree straight after it races the message pump.
+    Pumping rather than sleeping, because the app is running: the notice
+    arrives as a message, and an event loop nobody hands back to cannot take
```
tconbeer: "the notice arrives as a message" isn't what happens here, and the docstring contradicts the argument it's explaining.

The predicate is `tunnel.needs_restart` → `self.dropped and not self.restart_failed`, and `dropped` is set directly on the daemon watcher thread at `ssh.py:685`. No message, no handler. That's why this call passes `interval=POLL_INTERVAL` — and `wait_for`'s own docstring says that interval is "for a condition a thread sets, **which no amount of pumping brings closer**."

"Pumping rather than sleeping" also names the `wait_until` call this replaced, which AGENTS.md rules out.

The real reason to pump is that a blocking sleep would freeze the running app's loop, and `interval=` already says that. One line does it:

```python
    """Wait for the watcher thread to notice the child let the forward go."""
```

---

tests/functional_tests/test_data_catalog.py
```diff
@@ -486,14 +502,25 @@ async def test_reload_while_loader_is_fetching(
...
-        for _ in range(50):
-            await pilot.pause()
+        await wait_for(
+            pilot,
+            BlockingCatalogItem.returned.is_set,
+            description="the blocking fetch to return",
+            interval=POLL_INTERVAL,
+        )
+        # the loader takes the result on the pump, which is where task_done()
+        # used to land on the queue the reload had just put in place
```
tconbeer: "where `task_done()` **used to** land" records a bug that no longer exists in the tree — AGENTS.md rules that out specifically. `# the loader takes the result on the pump` says everything the next reader needs.

Separately, on this block: the negative assertion below (`assert not [w.error for w in loaders ...]`) went from 50 pumps on main to one `pilot.pause()` here, gated on a `returned` event the worker sets **inside** `fetch_children`, before it returns. The error it's watching for is raised by the loader coroutine *after* the worker returns. One `pause()` is `wait_for_idle(0)`, so it's probably enough — but a negative assertion is the shape that passes vacuously when under-waited, and this is the test guarding #991. Worth waiting on the loader's own state (queue drained / no worker in `RUNNING`) rather than a bare pause.

---

tests/functional_tests/conftest.py
```diff
@@ -254,3 +264,115 @@ def fn(app: Harlequin) -> bool:
...
+    The editor collection is mounted lazily, so an app that is running does not
+    yet have an editor to type into.
+    """
+
+    async def editor(pilot: Pilot, app: Harlequin) -> CodeEditor:
+        return await wait_for_value(
+            pilot, lambda: app.editor, description="the Query Editor to be mounted"
+        )
+
+    return editor
```
tconbeer: None of the six new waits needs to be a fixture, and this PR already shows the better shape twice.

They take `(pilot, app)` as arguments, close over nothing, and request no other fixture. `open_export_screen` (`test_export.py:17`) and `open_history` (`test_history_screen.py:46`) — both added by this PR — are plain module-level `async def`s with the identical signature, and they read better at every call site.

What the fixture form costs in the tree as it stands: 191 mentions of `wait_for_editor`, 144 `Awaitable[CodeEditor] | [ResultsTable] | [ErrorModal] | …` annotations across 16 files, six imports per test module that exist only to spell a fixture's type, and 35 `Callable[..., Awaitable[...]]` whose `...` erases argument checking entirely — so mypy validates nothing about `rendered_catalog(pilot, app, replacing=...)` despite strict mode.

Moving all six (plus `expand_catalog_node`) into a `tests/functional_tests/helpers.py`, imported the way `tests.waiting` already is, leaves every test *body* unchanged, deletes a parameter and an annotation from every test *signature*, and restores full argument type-checking. That's most of the diff's remaining bulk.

`rendered_catalog`'s `replacing: object = None` then splits into two honestly-typed functions instead of one sentinel doing double duty (`object` as the annotation is the giveaway) plus a ternary in the description.

---

AGENTS.md
```diff
@@ -66,6 +66,8 @@ uv run --python 3.12 --group test pytest -m 'py12 and not online' --snapshot-upd
 ```
 
 - Async tests need `@pytest.mark.asyncio`. Await `wait_for_workers(app)` (fixture) rather than sleeping — it skips the catalog background loader, which never finishes on its own.
+- **Every wait lives in `tests/waiting.py`, and waits on a condition rather than a duration.** A test that sleeps, or pumps a fixed number of times, asserts that the machine it runs on is fast enough, and fails on a loaded runner for a reason nothing in it explains. `wait_until` polls state another thread or process sets; `wait_for` / `wait_for_value` / `wait_for_messages` pump the app while they poll and name what never happened when they give up; `settle` is the only duration there is, for establishing that nothing *else* happened. `free_port` draws from outside the range the OS hands out on its own, and `on_a_free_port` retries a child that lost the port anyway. The app-shaped conditions are fixtures over those, in `tests/functional_tests/conftest.py`: `wait_for_editor`, `wait_for_table`, `wait_for_error_modal`, `wait_for_catalog_tree`, `rendered_catalog`, `first_database_node`, `expand_catalog_node`.
```
tconbeer: Two things here: the bullet is over budget, and its opening claim isn't true of the tree.

"**Every wait lives in `tests/waiting.py`**" — `wait_for_workers` (`tests/functional_tests/conftest.py:212, 222`) is a fixed-count pump: 8 handoffs × 4 `asyncio.sleep(0)`, and it *returns silently* when it runs out rather than raising, so a slow machine gets no diagnostic at all. `tests/hsql_sessions.py:77` hand-rolls a `deadline`/`time.sleep(0.05)` poll, `test_hsql_serve.py:1556` keeps `_until_status` (the sibling of the `_until` this PR deleted from that same file, and it polls with no sleep between `ask()` calls), and `test_hsql_serve.py:2241` is a bare `time.sleep(3)`.

Most of those are pre-existing, which is the point — the claim is about the tree, and the next agent to read this will trust it. Either fold them in or scope the sentence.

The middle sentence is the third copy of the same rationale in this diff (`waiting.py`'s module docstring and `free_port`'s docstring have it too), and enumerating all seven fixture names will rot on the next rename. Something like:

```markdown
- **Waits wait on a condition, not a duration, and live in `tests/waiting.py`.** `wait_until` polls state another thread or process sets; `wait_for` / `wait_for_value` / `wait_for_messages` pump the app while they poll and name what never happened; `settle` is the only duration. App-shaped waits are fixtures over those in `tests/functional_tests/conftest.py`.
```

---

tests/unit_tests/test_ssh.py
```diff
@@ -691,11 +687,9 @@ def test_watching_twice_does_not_double_the_notice(drop_trigger: Path) -> None:
 def test_a_tunnel_taken_down_on_purpose_is_not_news() -> None:
     """The control is the test above: the same harness does fire, on a death."""
     notices: list[str] = []
-    tunnel = child_tunnel(free_port())
-    tunnel.watch(notices.append)
-    tunnel.start()
+    tunnel = started_tunnel(before_start=lambda t: t.watch(notices.append))
     tunnel.stop()
-    time.sleep(0.5)
+    settle()
```
tconbeer: This one shortened from `time.sleep(0.5)` to `settle()` = 0.3 — a 40% smaller window for "the watcher must *not* fire". The other two conversions (lines 664, 681) were exact 0.3 → 0.3.

Under-waiting a negative assertion can only cause a false pass, so it's the safe direction, but it's an unremarked reduction in what `test_a_tunnel_taken_down_on_purpose_is_not_news` proves. `settle(0.5)` keeps it as it was.

---

tests/functional_tests/test_ssh_recovery.py
```diff
@@ -123,28 +91,27 @@ async def test_a_dropped_tunnel_is_reopened_before_the_next_thing_that_needs_it(
...
+        first_connection = app.connection
+        assert first_connection is not None
+
+        drop_trigger.touch()
+        await _wait_for_drop(pilot, dropping_tunnel)
+        # the child that comes back stays up
+        monkeypatch.delenv("FAKE_SSH_DROP_WHEN")
+
+        app.action_refresh_catalog()
+        await wait_for_workers(app)
```
tconbeer, line 101: Nit: `not in (None, first_connection)` is `==`-based, where main's two asserts were `is not`. No `HarlequinConnection` in this repo defines `__eq__`, so it's identity today — but identity is what the test means, and an adapter that grows an `__eq__` would quietly change what this waits for.

```python
lambda: app.connection is not None and app.connection is not first_connection
```

---

tests/functional_tests/conftest.py
```diff
@@ -254,3 +264,115 @@ def fn(app: Harlequin) -> bool:
...
+@pytest.fixture
+def wait_for_table() -> Callable[[Pilot, Harlequin], Awaitable[ResultsTable]]:
+    """The table the Results Viewer is showing, once a query has filled one in."""
+
+    async def table(pilot: Pilot, app: Harlequin) -> ResultsTable:
+        return await wait_for_value(
+            pilot,
+            app.results_viewer.get_visible_table,
+            description="the Results Viewer to show a table",
+        )
```
tconbeer: Latent trap worth a comment or a rename, not a live bug: `get_visible_table()` is "any table in the active pane", not "the table for the query I just ran".

The clear happens in `@on(QuerySubmitted) execute_query` → `show_loading()` → `clear_all_tables()`, and `clear_all_tables` doesn't await `clear_panes()`'s `AwaitComplete`, so the old pane survives into at least one more turn. A second `wait_for_table` after a re-run would be free to return the *first* query's table.

I checked every call site and none does that today — `test_multiple_queries` correctly gates on `tab_count == 2` instead — so nothing is broken. But "wait for the results table" is what the name promises, and the next multi-query test will reach for it. Taking an expected tab id, or naming it `wait_for_any_table`, would keep it honest.

---

tconbeer (commented):
## Scope

Four reads, each with a stated lens: correctness of the new primitives (can a wait return before the state is ready), fidelity of the rewrite (did a test lose an assertion), prose against AGENTS.md, and reuse. Plus running the thing.

On this branch: **2047 passed**, functional tests green twice back to back, 148 snapshots pass with none unused, ruff clean. (I hit 11 unit-test failures that are my container's fault — I synced `--all-extras`, so 12 adapters are installed and the "every installed adapter" assertions fail. Not the PR.)

The direction is right and two parts are unambiguous improvements: the port fix is correct in its details (bind-then-close leaves nothing in TIME_WAIT because it never listens; omitting `SO_REUSEADDR` is right, since setting it would make the probe *more* permissive than the child), and moving the dropping tunnel's stop into the fixture strictly reduces leaked children. A mechanical check found **zero** snapshot calls, key presses, query strings or parametrize lists changed anywhere in 1371 lines — the sweep really is a sweep.

## What I'd fix before merging

**Two waits can return before the state is ready.** Both were reproduced by widening the window with an injected delay; both pass today only because the pump beats one `pilot.pause()`, which is the assumption this PR exists to remove.

1. `expand_catalog_node` returns with the "loading…" placeholder still rendered, for any node with exactly one real child — `loaded` is set on the worker thread, the repopulation happens later on the loop, and `len(node.children) == len(data.children)` is `1 == 1` in between. The `tiny` fixture DB is a 1-child node at *every* level. Inherited verbatim from the loop it replaces, so not a regression — but it's now the fixture every catalog test waits through. One-line fix, `DatabaseTree._has_placeholder` already exists.
2. `first_database_node` after a catalog refresh can hand back the node from the *previous* catalog. `rendered_catalog` returns when the `catalog` var is assigned, and `watch_catalog` is async — `await self.reload()` hasn't remounted anything yet — and `first_database_node` only asks that `root.children` be non-empty, which is true of the old tree. `test_ssh_recovery.py:179`'s `assert db_node.data.connection is app.connection` is the assertion that then fails, and it's the one guarding #1127.

**One test lost its coverage.** `select 1 where false` in `test_queries_do_not_crash_all_adapters` no longer waits for the fetch phase at all — main's `wait_for_workers` walked the `QueriesExecuted → _fetch_data` handoff, and the new body stops at the message and then excludes that parametrization from `wait_for_table`. It's the empty-result case, which is the one most worth watching a crash in.

**`wait_for_messages` turned "exactly N" into "at least N"** at eleven sites. `test_app.py:167` still says `# should only run one query` over a wait that no longer checks "only".

## Smaller, but worth a pass

- `wait_for_messages`' timeout message reports the count from before the wait — always "saw 0".
- `on_a_free_port`'s `retry_on=HarlequinSshError` catches readiness timeouts too, and `dropping_tunnel` inherits the 60s default: a genuinely broken fixture is 300s of silence, then attempt 5's error.
- `random.Random(os.getpid())` only *adds* a collision mode over the unseeded default.
- `wait_until` returns `bool`, so nine sites get a bare `assert False` where `_until` used to raise with a message.
- Four docstrings/comments narrate removed code or editorialize — `free_port`'s, the module docstring's second paragraph, `# …where task_done() used to land`, and "Stopped here rather than by each test." AGENTS.md is explicit about all four.
- The six new conftest waits close over nothing and take `(pilot, app)`; this PR's own `open_export_screen` and `open_history` show the better shape. 191 mentions of `wait_for_editor`, 144 `Awaitable[…]` annotations, and 35 `Callable[..., Awaitable[…]]` that turn off argument checking.

## One observation, not a finding

`editor = await wait_for_editor(...)` bound once and read across six tab switches in `test_multiple_buffers` is correct — but only because of the "exactly one `CodeEditor`, however many buffers" invariant. That invariant is now load-bearing for ~190 call sites: if `EditorCollection` ever holds two, every one of them silently reads the wrong editor. Worth knowing; arguably worth a line in AGENTS.md next to the invariant itself.

And a question the PR is well placed to answer: `test_multiple_buffers`, `test_word_autocomplete` and `test_member_autocomplete` are still `@pytest.mark.flaky`, same three as on main, and still full of `press(...)` → `pause()` → assert. Those are the tests the repo has already given up on. Can the new primitives retire the markers? That would be the proof the approach works.

## Not checked

- I did not review any source change, because there are none — the diff is tests and AGENTS.md.
- I did not run the suite on macOS or Windows, which is where the flakes this PR targets actually show up. Everything I ran was Linux, and CI on this SHA was still in progress when I looked (nothing failed; several macOS and Windows jobs were queued).
- I did not verify the port range against the CI images' listening ports. 20000-30000 is correctly below every supported platform's ephemeral floor, but a service that binds *after* the probe is unfixable by construction, and I didn't enumerate what the runner images hold.
- My mypy run is unusable (my container can't resolve `textual`/`pytest` stubs — 1215 errors, all cascade). CI's Static Analysis passed, so I'm taking type-correctness from CI, not from my own check.
- The two reproductions behind findings 1 and 2 used injected delays. I did not try to estimate how often a loaded runner actually loses those races.
- I did not re-run the snapshot baselines on 3.10 to confirm the three moved snapshot timings (export, run_query_bar, keys_app) produce identical SVGs — the committed snapshots are unchanged and the suite passes, and `app_snapshot` retries across settle delays, which absorbs most of that class of change.

A green review isn't the merge gate here — the "is the intent preserved?" call is yours. I'd want at least findings 1-3 addressed first; everything below them would be fine as a follow-up.

---

tests/functional_tests/test_app.py
tconbeer: Fixed — dropped the special case, so every non-empty query waits for its table and only the `row_count >= 1` assertion is conditional:

```python
            assert query_executed_message.cursors
            # an empty result is still a table, with a header and no rows
            table = await wait_for_any_table(pilot, app)
            if query != "select 1 where false":
                assert table.row_count >= 1
```

You're right that it was the one parametrization worth watching — an empty result is `create_backend()`'s no-rows path, and it was the only one whose fetch phase the test no longer entered.

---

tests/waiting.py
tconbeer: Fixed, with the `exactly=` you suggested:

```python
    if exactly:
        assert len(collected()) == count, ...
```

All ten unpacking sites in `test_app.py` pass it, which is every site that was `[x] =`, `[_, x] =` or `[*_, x] =` on main. `[*_, x]` with `count=3, exactly=True` is the same message main took, since "the last of exactly three" is the third.

---

tests/functional_tests/test_ssh_recovery.py
tconbeer: Fixed, taking the second option — the catalog waits now gate on `loading`, which `watch_catalog` clears at the end, after the `await self.reload()` that remounts the nodes. `first_database_node` goes through `wait_for_catalog_tree`, so it can't hand back the old tree's node any more.

That also removed the sentinel: `rendered_catalog(pilot, app)` and `replaced_catalog(pilot, app, previous)` are two functions instead of one `replacing: object = None`.

---

tests/functional_tests/conftest.py
tconbeer: Fixed with `DatabaseTree._has_placeholder`, as you suggested:

```python
        return (
            getattr(data, "loaded", True)
            and not DatabaseTree._has_placeholder(node)
            and len(node.children) == len(data.children)
        )
```

Agreed on the reasoning for fixing it here rather than leaving it as inherited: promoting it to the wait every catalog test goes through is what makes it worth more than the `while` loop it came from.

---

tests/waiting.py
tconbeer: Fixed for every caller rather than just this one — `description` is now `str | Callable[[], str]`, resolved at the raise:

```python
Description = Union[str, Callable[[], str]]


def _timed_out(seconds: float, description: Description) -> AssertionError:
    resolved = description() if callable(description) else description
    return AssertionError(f"timed out after {seconds}s waiting for {resolved}")
```

`wait_for_messages` passes a lambda, so it now reports the count it gave up on.

---

tests/waiting.py
tconbeer: All four cut. `free_port` is close to your version:

```python
def free_port() -> int:
    """A loopback port nothing is listening on, drawn from below the ephemeral range.

    It can only report a port that was free a moment ago; a caller that hands
    one to a child has `on_a_free_port`.
    """
```

The module docstring lost the rationale paragraph entirely (AGENTS.md carries it), `# …used to land` became a statement of the live assertion, `wait_for_messages` describes `exactly` instead of arguing for itself, and "Stopped here rather than by each test" is gone from the `dropping_tunnel` fixture.

---

tests/waiting.py
tconbeer: You're right, and the shared-network-namespace case is the one I hadn't thought about. Dropped the argument:

```python
_draw = random.Random()
"""Seeded from the OS, so two xdist workers walk different candidates."""
```

---

tests/waiting.py
tconbeer: Fixed — `retry_on` takes a predicate rather than a type, and the ssh callers pass one that reads the two cases apart the way you described:

```python
def ssh_child_exited(error: BaseException) -> bool:
    """Whether `start()` failed because the child exited, rather than timing out."""
    return isinstance(error, HarlequinSshError) and "exited with code" in str(error)
```

Message-sniffing, but the alternative wasn't available: `start()` calls `stop()` on the way out, which clears `_process`, so `returncode` is gone by the time the caller sees the error. `_child_failed` is the only one of the two that says "exited with code", so a readiness timeout is never retried — which was the 300s case.

It lives in `tests/tunnels.py` because both ssh modules need it.

---

tests/waiting.py
tconbeer: Fixed — `wait_until` takes a `description` and raises, and the nine `assert` prefixes are gone with it. Both primitives raise through the same `_timed_out`, so the message is identical whichever one a test used.

---

tests/functional_tests/test_ssh_recovery.py
tconbeer: You're right on both counts — `dropped` is set on the watcher thread at `ssh.py:685`, no message involved, and I'd written a docstring that contradicted the `interval=` on the next line. Cut to your version:

```python
async def _wait_for_drop(pilot: Pilot, tunnel: SshTunnel) -> None:
    """Wait for the watcher thread to notice the child let the forward go."""
```

---

tests/functional_tests/test_data_catalog.py
tconbeer: Both fixed, and the second one turned into a better test than it was on main. The bare pause is gone; the wait is now the regression itself:

```python
        queue_it_came_from = tree._load_queue
        await tree.reload()
        BlockingCatalogItem.release.set()

        # the loader marks the item done on the queue it took it from, which is
        # the assertion: marking the replacement instead leaves this at 1, and
        # raises "task_done() called too many times" over there
        await wait_for(pilot, item_marked_done, ...)
```

That makes the negative assertion a positive one — it can't pass vacuously, because the thing it waits for is exactly what #991 got wrong. Verified by reverting `queue.task_done()` to `self._load_queue.task_done()` in the source: the test fails with `timed out after 10.0s waiting for the loader to finish with the item it was fetching`. The `returned` event I'd added is gone, since the pause it gated went with it.

---

tests/functional_tests/conftest.py
tconbeer: Done — `tests/functional_tests/helpers.py`, imported the way `tests.waiting` already is. All seven moved, test bodies unchanged, and the signatures lose **259 lines** net.

The `Callable[..., Awaitable[...]]` point was the one that decided it: mypy was checking nothing about those calls in strict mode, and `replaced_catalog(pilot, app, previous)` is now actually type-checked. The sentinel split out as you said — `rendered_catalog(pilot, app)` and `replaced_catalog(pilot, app, previous)`, no `object` annotation, no ternary in the description.

`wait_for_workers` and `transaction_button_visible` stay fixtures: the first is arguably one too, but it's pre-existing and not in this PR's blast radius.

---

AGENTS.md
tconbeer: Fair — the claim was false, and I'd have misled the next agent with it. Scoped the sentence and named what isn't folded in:

> - **A wait waits on a condition, not a duration, and new ones go in `tests/waiting.py`.** `wait_until` polls state another thread or process sets; `wait_for` / `wait_for_value` / `wait_for_messages` pump the app while they poll and raise naming what never happened; `settle` is the only duration. The app-shaped waits are plain functions over those in `tests/functional_tests/helpers.py`. Loopback ports come from `free_port`… Predating all of this and not yet folded in: `wait_for_workers`, `tests/hsql_sessions.py`, and `_until_status` and one `time.sleep(3)` in `test_hsql_serve.py`.

The fixture enumeration is gone, and the next bullet now also says `wait_for_workers` gives up silently after a fixed number of pumps.

---

tests/unit_tests/test_ssh.py
tconbeer: Caught me — that was an unremarked change, not a conversion. `settle(0.5)` restores it.

---

tests/functional_tests/test_ssh_recovery.py
tconbeer: Fixed — back to `is not`, for the reason you give:

```python
            lambda: app.connection is not None
            and app.connection is not first_connection,
```

---

tests/functional_tests/conftest.py
tconbeer: Renamed to `wait_for_any_table`, and the trap is in the docstring rather than left for the next person to find:

```python
async def wait_for_any_table(pilot: Pilot, app: Harlequin) -> ResultsTable:
    """The table the Results Viewer is showing, once a query has filled one in.

    Any table in the active pane: a re-run clears the panes without awaiting
    the removal, so this can still hand back the previous query's table. A test
    that runs a second query gates on `results_viewer.tab_count` instead.
    """
```

Took the rename over the tab-id parameter: every current caller runs one query, so a required id would be ceremony at ~30 sites to describe a case none of them is in. The name now says what it does, and the docstring points at what `test_multiple_queries` does instead.

---

tconbeer:
All 16 threads addressed in `0bff614`, each answered inline. `make check` is green: 2060 passed on 3.10, the py12 slice passed, mypy and lint-imports clean.

## Your question: can the new primitives retire the `flaky` markers?

Yes — all three are gone, and the tests pass 5/5 back to back locally without them.

They shared one shape: press a key, settle, assert. Each had a condition sitting right there.

`test_multiple_buffers` was asserting `active` and `editor.text` after `pause()` + `wait_for_scheduled_animations()`. Activating a tab and swapping its state into the one editor are two steps on the pump, so the text lands after the id does — which is exactly what to wait on:

```python
async def loaded_buffer(pilot: Pilot, app: Harlequin, tab_id: str, text: str) -> None:
    collection = app.editor_collection
    await wait_for(
        pilot,
        lambda: collection.active == tab_id and collection.current_editor.text == text,
        description=f"{tab_id} to be the loaded buffer, holding {text!r}",
    )
```

The two autocomplete tests were snapshotting the completion dropdown after each keypress, and the keypress reaches the list through the editor's own worker — so the snapshot could catch the list a keypress behind. `completions_for(pilot, editor, text)` waits for the buffer to hold `text`, the list to be open, and `option_count > 0`.

One thing that fell out of writing that: the editor auto-closes quotes, so after `quotation_mark` the buffer is `"drivers".""`, not `"drivers"."`. My first attempt asserted the typed text and failed 3/3 — the wait found the bug in my own assumption in about a minute, which is the argument for the whole approach better than anything in the PR description.

This also means Windows' `--force-flaky --max-runs=3` is now the only thing hiding this class of failure anywhere in the suite. Not touching that here.

## Also worth flagging from your review

**The invariant.** Added a line to AGENTS.md next to it, as you suggested — the tests bind the editor once and read it across tab switches, so a second `CodeEditor` would silently give ~190 assertions the wrong widget.

**`wait_for_workers`.** Your AGENTS.md finding was right that it isn't in `tests/waiting.py` and isn't condition-based — it pumps a fixed 8 × 4 and returns silently. I scoped the claim rather than folding it in: it takes no pilot, so "the pump has drained" isn't something it can observe, and giving it one is a bigger change than this PR should carry. It's named in AGENTS.md as not yet folded in, along with `tests/hsql_sessions.py`, `_until_status` and the `time.sleep(3)`.

**Your "not checked" list.** Two I can't close either: I also only ran Linux, and the port range is still unverified against the CI images' listening ports. The four macOS and Windows jobs on this SHA are the real test of both.

---

tconbeer:
Rebased onto `8b99a4c`. `make check` green on the new base: 2071 passed on 3.10, py12 slice passed, mypy and lint-imports clean.

The rebase itself was clean — #1162 touches no file this branch does. But its new `tests/functional_tests/test_windows_timezone.py` is the shape this PR removes, so I swept it (`b9bb321`): a 200-pump loop for the fetch, a 20-pump loop standing in for "and nothing happened", and four `while app.editor is None`. Without that, this PR's AGENTS.md rule would have been false the moment it merged.

That turned up one gap in `tests/waiting.py`: `settle` is a blocking sleep, which stops the loop the thing that must not happen would have to happen on. So there is now an async `settle_app(pilot)` for the negative assertion in a running app. Same idea, one place.

**One thing I left alone in that file**, since it's a design call on your test rather than a mechanical sweep: it blocks the event loop on `threading.Event.wait(timeout=10.0)` in four places (`download_started.wait(...)`, `stop_seen.wait(...)`). That's the same hazard the ssh tests had before this PR — while the app is running, nothing pumps. It works because the download thread sets those events independently of the loop, and each has its own bound. `await wait_for(pilot, download_started.is_set, interval=POLL_INTERVAL, ...)` would be the consistent form; happy to do it if you want, but it changes what the app gets to process during the wait, so it's not a no-op.

---

PR #1166, opened 2026-09-17
Declare each CLI option once, with the flags that say how it may be used

### LINKED ISSUE #1145: Declare an option's group on the option itself, not in a name list beside it
`hsql` sorts every option into one of five groups — `CONNECTION_OPTIONS`, `PER_REQUEST_OPTIONS`, `SERVER_OPTIONS`, `ROLE_OPTIONS`, `CONFIG_OPTIONS` — which decide whether `--serve` accepts it, whether a served request may carry it, and whether it is compared against the session's connection identity. See [`docs/cli-options.md`](https://github.com/tconbeer/harlequin/blob/main/docs/cli-options.md).

Today those groups are frozensets of option *names* at the top of `harlequin/hsql/cli.py`, several definitions away from the `@click.option` they describe. Adding an option means declaring it in one place and classifying it in another, and nothing makes the second step happen — `test_every_option_is_in_exactly_one_group` fails afterwards rather than the declaration being impossible to write without a group.

**Proposed:** a `click.Parameter` subclass carrying the group as a property, so an option declares how and where it can be used in one place:

```python
@click.option(
    "--limit", cls=HsqlOption, group=PER_REQUEST, ...
)
```

The partition test then reads the parameters rather than the name lists, and the two call sites that ask "is this key a connection option" ask the parameter. Same for `CLI_ONLY_SESSION_KEYS`, `CLI_ONLY_SSH_KEYS` and `TUI_ONLY_KEYS`, which are the same shape of list: properties on the declaration rather than membership somewhere else.

Worth checking against `harlequin.options.AbstractOption` while doing it — an adapter's options are all connection options today, and that rule is implicit rather than declared.

Raised in review of #1140, which added the groups doc.

tconbeer:
Closes #1145

`harlequin.core_options` is one ordered list of `CoreOption`s covering both commands' own options. A declaration carries:

| field | says |
| --- | --- |
| `harlequin=On(...)` / `hsql=On(...)` | which command takes it, and how that command spells it — `decls` and `kwargs` merged over the shared ones, which is where the two commands' help differs |
| `group` | when `hsql` reads the value, and so what `--serve` and a served request do with it |
| `cli_only` | whether a config file is refused it |
| `first_pass` | whether the pass that names the adapter reads it off raw argv |

`__post_init__` refuses a declaration `hsql` takes with no group, and one only the IDE takes *with* a group. `attach_core_options()` builds the click options in declaration order, which is `hsql --help`'s order; `Supplied` is the part only a command can answer — a `click.Choice` of what is installed, `_record_source`, the wizard and keys-app callbacks, help naming a run-time default — so nothing here imports either command.

Everything that was a name list is now derived from the declarations: the five group frozensets, `TUI_ONLY_KEYS`, `CLI_ONLY_SESSION_KEYS`, `CLI_ONLY_SSH_KEYS`, `SSH_KEYS`. Same members as before. `first_pass()` builds its probe from `first_pass_options(program)` instead of three hard-coded options plus hsql's duplicate `extra_options` list, and the probe now names `--config` `config_mode`, as the command does.

Both commands carry exactly the parameters they did before. Checked against a worktree of `main` in the same environment: `hsql`'s parameters, `--help` and `--spec` are byte-identical, and `docs/generated/hsql-reference.md` regenerates unchanged. The IDE's parameters are identical in every attribute and differ only in declaration order, which its group-driven `--help` does not read. The only artifact change is the order of the seven IDE-only keys inside the generated schema.

The narrow fix in the issue was a `click.Parameter` subclass taking `group=`. Per the follow-up comment this goes the other way: one module declaring both commands' options, with flags on each. That also picks up `first_pass`, which was the last list of option names kept beside the code that read them.

Tradeoffs:

- **Per-command overrides.** The two commands genuinely differ — `--limit` has a default in `hsql` and none in the IDE, `--config` is a different option in each, most shared options word their help differently. `On(decls=..., kwargs=...)` keeps those differences visible in one place rather than hiding them behind a lowest common denominator. Nothing was made uniform that wasn't already.
- **`Supplied` instead of imports.** The alternative was importing `harlequin.hsql.output`, `harlequin.hsql.modes` and `harlequin.colors` to build choices and help. The last is forbidden to `hsql` outright, so the dependency has to point the other way.
- **`harlequin.config` still does not import `click`.** It is on the execution core's path — 141 modules against `core_options`' 156 — so `take_ssh_keys()` imports the two key lists inside the function.
- **Declaration order is `hsql --help`'s**, since that order is user-visible output and `--spec` is a documented artifact. The IDE's options sit last; its help renders from the rich-click option groups, so their order is free.
- **No `group` on `AbstractOption`** (the issue's "worth checking" note): every adapter option is a connection option by construction, so a field with one possible value would be ceremony. The rule stays where it is applied, in `connection_option_names()`.

- [x] No.

- [x] Yes.

`tests/unit_tests/test_core_options.py` pins that a declaration cannot omit a group, that each command carries exactly what it declares, and that the first pass spells an option the way the command does. `test_every_option_is_in_exactly_one_group` still reads the built command against the derived sets, which is what would catch an option added as a bare `@click.option`.

Full suite passes on 3.10 (2123 passed, 7 skipped, 149 snapshots) and the `py12` subset on 3.12; mypy strict, ruff and all five import-linter contracts are clean. `docs/cli-options.md` is rewritten for the new checklist.

https://claude.ai/code/session_015DYVDb6ZvMBBcejVCMSxuy

---
_Generated by [Claude Code](https://claude.ai/code/session_015DYVDb6ZvMBBcejVCMSxuy)_


src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+    """Read to decide which process runs the invocation."""
+
+    CONFIG = "config"
+    """Read to pick where the other values come from."""
+
+
+@dataclass(frozen=True)
+class Supplied:
+    """A click keyword only the command can fill in: a choice of what is
+    installed, a callback, help naming a run-time value."""
```
tconbeer: let's call this a RuntimeValue or DynamicValue. The docstring is confusing... it should state that this is a method of deferring execution or interpolation until the command is being built.

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+    key: str
+
+
+@dataclass(frozen=True)
+class On:
+    """One command's declaration: empty where it spells the option as the
+    shared declaration does, `decls` replacing it and `kwargs` merging over."""
+
+    decls: tuple[str, ...] = ()
+    kwargs: Mapping[str, Any] = field(default_factory=dict)
```
tconbeer: I don't love the name... I see that in the CoreOption it reads cleverly (`harlequin=On()`), but it especially makes no sense when overrides are passed in. At the very least, make the docstring more clear: "Acts as a sentinel to enable this option for a specific command; optionally provides overrides for command-specific decls and kwargs". Maybe EnabledWithConfig?

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+
+    def __post_init__(self) -> None:
+        if self.harlequin is None and self.hsql is None:
+            raise ValueError(f"{self.name} is declared by neither command.")
+        if (self.hsql is None) is not (self.group is None):
+            raise ValueError(
+                f"{self.name} must declare a group if and only if hsql takes it."
+            )
+
+    def on(self, command: str) -> On | None:
```
tconbeer: this name is especially bad. I might be ok with the dataclass being called On, but this method needs a better name. Maybe `is_enabled_for_command` or `command_config` or `config_for_command` or something explicit like that.

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+        supplied: Mapping[str, Any],
+        *,
+        option_cls: type[click.Option] | None = None,
+        argument_cls: type[click.Argument] | None = None,
+    ) -> Callable[[click.Command], click.Command] | None:
+        """The decorator that puts this option on `command`, or None.
+
+        Click's `@option` appends to `Command.params`, which is what orders it.
+        """
+        spelling = self.on(command)
```
tconbeer: "spelling" suggests it's all the decls, but it's not. This should be called command_config

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+    """Whether a config file is refused it: a value deciding which process
+    runs, or one a discovered file may not weaken."""
+
+    argument: bool = False
+    """Whether it is positional. `CONN_STR` alone is."""
+
+    first_pass: bool = False
+    """Whether the pass that names the adapter reads it off raw argv."""
+
+    supplied_param: str | None = None
```
tconbeer: this is a confusing name, because it doesn't use Supplied()... It also seems to duplicate the functionality of Supplied... I don't really understand why --version can't reuse that mechanism

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+
+    key: str
+
+
+@dataclass(frozen=True)
+class On:
+    """One command's declaration: empty where it spells the option as the
+    shared declaration does, `decls` replacing it and `kwargs` merging over."""
+
+    decls: tuple[str, ...] = ()
```
tconbeer, line 100: it should be more clear that decls REPLACES the shared definition, instead of merging with it.

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+    """The declared keywords, with the command's part filled in."""
+    return {
+        key: _from_the_command(supplied, value.key, name=name)
+        if isinstance(value, Supplied)
+        else value
+        for key, value in kwargs.items()
+    }
+
+
+def _from_the_command(supplied: Mapping[str, Any], key: str, *, name: str) -> Any:
```
tconbeer: also a terrible name... more like "_unwrap_runtime_value" or "_lookup_dynamic_value"

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+        if self.argument:
+            if argument_cls is not None:
+                kwargs["cls"] = argument_cls
+            return click.argument(*decls, **kwargs)
+        if option_cls is not None:
+            kwargs["cls"] = option_cls
+        return click.option(*decls, **kwargs)
+
+
+def _fill_in(
```
tconbeer: how about "_build_click_kwargs" and the docstring should make it clear that this just unwraps/interpolates dynamic/runtime values?

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+class CoreOption:
+    """One option of the `harlequin` or `hsql` command, and how it may be
+    used. `name` is the click parameter, and the profile key."""
+
+    name: str
+    decls: tuple[str, ...] = ()
+    group: Group | None = None
+    """When hsql reads it; None of what hsql does not declare."""
+
+    kwargs: Mapping[str, Any] = field(default_factory=dict)
```
tconbeer: can this be a TypedDict or DataClass instead? (shared with On.kwargs). There are really only a handful of keys that we use.

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+        group=Group.PER_REQUEST,
+        harlequin=On(
+            kwargs={
+                "type": click.Path(file_okay=True, dir_okay=True, path_type=Path),
+                "help": "The default directory or file path for the Data Exporter.",
+            }
+        ),
+        hsql=On(
+            kwargs={
+                "metavar": "PATH",
```
tconbeer: shouldn't this also have "type": click.Path?

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+            }
+        ),
+        hsql=On(
+            kwargs={
+                "help": (
+                    "Load a profile from an available config file. Options passed "
+                    "here take precedence over the profile's. Use the profile "
+                    "named None for Harlequin's defaults instead of the config "
+                    "file's default profile."
+                )
```
tconbeer: No reason to have different help text for the two commands here. Take Harlequin's as the option's.

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+                    "to use to connect to the database at CONN_STR."
+                )
+            },
+        ),
+        hsql=On(
+            decls=("-a", "--adapter"),
+            kwargs={
+                "metavar": "NAME",
+                "help": "The installed adapter plug-in to connect with.",
+            },
```
tconbeer: No reason to have different help text for the two commands here. Take Harlequin's as the option's. Also make the metavar the same for both.

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+            decls=("--read-only", "-r", "read_only"),
+            kwargs={
+                "help": (
+                    "Connect read-only, and refuse to start at all if the adapter "
+                    "cannot. To check an adapter's capabilities, use `hsql --info`."
+                )
+            },
+        ),
+        hsql=On(
+            decls=("-r", "--read-only", "read_only"),
```
tconbeer: these decls should be the same; reorder to make the same.

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+        group=Group.CONNECTION,
+        kwargs={
+            "help": (
+                "Open an SSH tunnel to this destination first, and connect through "
+                "it. A Host alias, host, user@host or ssh://user@host:port, passed "
+                "to ssh verbatim."
+            )
+        },
+        harlequin=On(),
+        hsql=On(kwargs={"metavar": "TEXT"}),
```
tconbeer: should this metavar just be the same for both commands?

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+                )
+            }
+        ),
+        hsql=On(
+            kwargs={
+                "metavar": "TEXT",
+                "help": (
+                    "A local forward, spelled as ssh -L takes one: "
+                    "LOCAL:HOST:REMOTE. Repeatable. Omit it when your ssh config "
+                    "has the LocalForward."
```
tconbeer: make this help text the same for both commands.

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+                    "Fail rather than prompt for a passphrase, a password or a "
+                    "host key. ssh's own BatchMode."
+                )
+            }
+        ),
+        hsql=On(
+            kwargs={
+                "help": (
+                    "Fail rather than prompt for a passphrase, a password or a "
+                    "host key. ssh's own BatchMode; set it in scripts, CI and cron."
```
tconbeer: take harlequin's help text as the command's

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+                    f"{DEFAULT_SSH_TIMEOUT:g}"
+                )
+            }
+        ),
+        hsql=On(
+            kwargs={
+                "metavar": "SECONDS",
+                "help": (
+                    "Seconds to wait for the tunnel's forwards. "
+                    f"[default: {DEFAULT_SSH_TIMEOUT:g}]"
```
tconbeer: unify these help texts

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+                    "the current directory and the home directory (~) and merges "
+                    "them. Use this option to specify the full path to a config "
+                    "file at a different location."
+                ),
+            }
+        ),
+        hsql=On(
+            kwargs={
+                "type": click.Path(dir_okay=False, resolve_path=True, path_type=Path),
+                "metavar": "PATH",
```
tconbeer: use the same metavar for both commands.

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+                "help": "Use this config file instead of the ones hsql discovers.",
```
tconbeer: use harlequin's help text as the command's

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+
+def first_pass_options(command: str) -> list[click.Option]:
+    """The options the first pass reads, as the copies it probes argv with.
+
+    Spellings, whether a value follows, an envvar: no more, because the probe
+    must survive an argv the command would refuse.
+    """
+    probe: list[click.Option] = []
+    for option in CORE_OPTIONS:
+        spelling = option.on(command)
```
tconbeer: "command_config", not "spelling"

src/harlequin/core_options.py
```diff
@@ -0,0 +1,1121 @@
...
+        if not option.first_pass or spelling is None:
+            continue
+        kwargs = {**option.kwargs, **spelling.kwargs}
```
tconbeer: why re-implement this? Isn't this option.to_click? Also, don't you need to load Supplied() values?

tconbeer (commented):
two sets of changes that should land as two commits:

1. Names are confusing/bad, and there are some small issues with implementation/design
2. Help text (in particular) is unnecessarily command-specific. I would like to harmonize help text (and metavars etc) across the commands where that makes sense. This should be a 2nd commit because it will require some test/artifact changes.

src/harlequin/core_options.py
tconbeer: Renamed to `RuntimeValue`, and the docstring now says what it is for: "A stand-in for a value that only exists once a command is being built. It defers a choice of what is installed, a callback, or help naming a computed default: whatever holds it is resolved against the mapping the command passes to `attach_core_options()`, under `key`."

---

src/harlequin/core_options.py
tconbeer: Kept `On` and rewrote the docstring along the lines you gave: "A sentinel enabling an option for one command, and its overrides. `decls` *replaces* the shared spellings rather than adding to them; `kwargs` merge over the shared keywords, key by key. Empty where the command takes the option exactly as declared."

Happy to rename it to `EnabledWithConfig` if you'd still rather — say the word and it's one commit.

---

src/harlequin/core_options.py
tconbeer: Renamed to `config_for_command()`.

---

src/harlequin/core_options.py
tconbeer: Renamed to `command_config`, here and in `first_pass_options()`.

---

src/harlequin/core_options.py
tconbeer: It reuses the mechanism now: the field is `decorator: RuntimeValue | None`, and `--version` declares `decorator=RuntimeValue("version_option")`.

What it can't share is the *shape*. Every other runtime value fills one click keyword; `--version` is a whole decorator, because `click.version_option()` writes the callback that prints the message, and that message is the command's own (`harlequin` prints its adapters, `hsql` prints `prog_name`). Declaring it as kwargs would mean reimplementing that callback and click's lazy version lookup here. The docstring says so.

---

src/harlequin/core_options.py
tconbeer: Done — the docstring now says `decls` *replaces* the shared spellings rather than adding to them, against `kwargs`, which merge key by key.

---

src/harlequin/core_options.py
tconbeer: Renamed to `_unwrap_runtime_value()`, and it now takes the `RuntimeValue` itself rather than its key.

---

src/harlequin/core_options.py
tconbeer: Renamed to `_build_click_kwargs()`; the docstring is now "The declared keywords, with every `RuntimeValue` in them unwrapped."

---

src/harlequin/core_options.py
tconbeer: Added `ClickKwargs`, a `TypedDict(total=False)` shared by `CoreOption.kwargs` and `On.kwargs`, with the twelve keys the declarations use: `help`, `type`, `default`, `show_default`, `metavar`, `is_flag`, `multiple`, `nargs`, `expose_value`, `envvar`, `show_envvar`, `callback`. Each is typed, with `| RuntimeValue` where one can stand in.

It earns its keep: `"is_flagg": True` is now `error: Extra key "is_flagg" for TypedDict "ClickKwargs"` rather than a silently ignored keyword.

---

src/harlequin/core_options.py
tconbeer: It should — moved `type` and `metavar` into the shared kwargs, so `hsql -o` is a `click.Path` like the IDE's.

One behavior note: click now hands the callback a `Path`, so a trailing slash is gone by the time `names_a_directory()` sees it. That check falls back to `path.exists() and is_dir()` and then to "no suffix", so the only case that reads differently is a directory that doesn't exist yet *and* has a dotted name (`-o out.d/`) — which had no directory to write into either way. Checked `-o DIR/` with two result sets end to end.

---

src/harlequin/core_options.py
tconbeer: Done — Harlequin's help is now the option's, and both `On()`s are empty.

---

src/harlequin/core_options.py
tconbeer: Done: Harlequin's help, `metavar="NAME"` for both, and the decls unified to `("-a", "--adapter")` — so this declaration has no per-command half left at all.

---

src/harlequin/core_options.py
tconbeer: Done — `("-r", "--read-only", "read_only")` is now shared. The help still differs, since the IDE refuses to *start* and `hsql` refuses to *run*, and each points at its own `--info`.

---

src/harlequin/core_options.py
tconbeer: Yes — `metavar="TEXT"` is shared now. It was already what the IDE rendered from the type, so its help is unchanged.

---

src/harlequin/core_options.py
tconbeer: Unified on Harlequin's, with `metavar="TEXT"` shared too. `hsql --help` now says "Repeat this option for more than one" in place of "Repeatable."

---

src/harlequin/core_options.py
tconbeer: Done. `hsql` loses "; set it in scripts, CI and cron" from its help.

---

src/harlequin/core_options.py
tconbeer: Unified on Harlequin's — "Seconds to wait for the tunnel's forwards. Default is 60." (I added the full stop it was missing) — with `metavar="SECONDS"` shared. `hsql --help` no longer writes this one as `[default: 60]`, which is the convention its other unset-default options use.

---

src/harlequin/core_options.py
tconbeer: Done — `metavar="PATH"` is shared, which is what the IDE already rendered from `click.Path`.

---

src/harlequin/core_options.py
tconbeer: Done. The `type` stays per-command — the IDE's requires the file to exist, and `hsql --config init` writes one that doesn't yet — but that is now the only thing this declaration splits.

---

src/harlequin/core_options.py
tconbeer: Renamed here too.

---

src/harlequin/core_options.py
tconbeer: It can't be `to_click()`, and the answer to the second question is why: there are no runtime values to load. This pass runs *before* the command exists — the `click.Choice` of installed adapters, `_record_source`, the wizard callback all live on the other side of it, and the whole point is to name the adapter without paying for one.

The probe also has to be weaker than the real option on purpose. Click parses it with `resilient_parsing`, inside `contextlib.suppress`, so anything that raises during conversion loses the *whole* probe — and with it the `-P` the invocation typed. A `Choice` would raise on an adapter name click hasn't vetted yet; the IDE's `--config-path` would raise on a file that isn't there, which is exactly the case `hsql --config init` exists for. So it takes spellings, whether a value follows, and the envvar, and nothing else. A `click.Path` stays a path only so `load_profile()` gets one.

I've said that in the docstring rather than leaving it to be rediscovered, and renamed the local to `command_config`. The new `test_the_first_pass_spells_an_option_the_way_the_command_does` is what keeps the two from drifting apart in spelling, which is the real risk in not sharing the builder.

---

